PR #2, opened 2019-04-12
Add support to systray & fix icon in BrowserWindows

Mutilador:
I'm using AppImage 0.6.5 with Ubuntu 18.04 and the problem is don't have icon in wallet's window. I fixed

Eu estou usando AppImage 0.6.5 com Ubuntu 18.04 e o problema é que a versão de vocês não aparece o ICON na janela, eu corrigi isso.


pedroferreira1:
Hey, thanks for the pull request.

I've tried in Windows, Ubuntu (18.04) and macOS and only on macOS the tray icon is really big (see image). I've tried a solution but didn't work, I still need to investigate more but will probably do this later. 
Screenshot 2019-04-15 at 19 40 32

Besides that, I think we should give the option for the user to turn on/off the tray (the default should be off). Maybe you could add a checkbox option in the menu to change it.

Mutilador:
Hi Pedro,

I fixed and add submenu with enable/disable Sistray.

public/electron.js
```diff
@@ -78,6 +101,14 @@ function createWindow () {
     // when you should delete the corresponding element.
     mainWindow = null
   })
+
+  mainWindow.on("close", (e) => {
```
msbrogli: Maybe we should show a message to the user saying that the application is still running on the background (with a checkbox to stop showing this message).

public/electron.js
```diff
@@ -94,6 +125,32 @@ if (process.platform === 'darwin') {
 // Some APIs can only be used after this event occurs.
 app.on('ready', createWindow)
 
+function startSystray() {
+  tray = new Tray(path.join(__dirname, trayIcon));
+  const contextMenu = Menu.buildFromTemplate([
+    {
+      label: 'Show/Hide',
```
msbrogli: Why don't we adjust the label according to the current state? It would be either `Show` or `Hide`.

msbrogli: I prefer `Hide Wallet` and `Show Wallet`.

public/electron.js
```diff
@@ -94,6 +125,32 @@ if (process.platform === 'darwin') {
...
+      click() {  
+        if (mainWindow.isVisible()) {
+          mainWindow.hide();
+        } else {
+          mainWindow.show();
+        }
+      }
+    },
+    { 
+      label: 'Exit',
```
msbrogli: It seems weird to only show the Exit menu when the systray is enabled. Wouldn't it be better to always show it?

Mutilador: > It seems weird to only show the Exit menu when the systray is enabled. Wouldn't it be better to always show it?

This menu is shown when the user clicks on the systray.

public/electron.js
Mutilador: Okay, you're right.

public/electron.js
Mutilador: done!

public/electron.js
```diff
@@ -17,6 +17,18 @@ Sentry.init({
   release: process.env.npm_package_version
 })
 
+let tray = null
+let iconOS = null;
+let trayIcon = null;
+let msgCheck = false;
```
msbrogli: We should save this variable in local storage. So, we would load it after user restarts the wallet.

public/electron.js
```diff
@@ -94,6 +142,32 @@ if (process.platform === 'darwin') {
 // Some APIs can only be used after this event occurs.
 app.on('ready', createWindow)
 
+function startSystray() {
+  tray = new Tray(path.join(__dirname, trayIcon));
+  const contextMenu = Menu.buildFromTemplate([
+    {
+      label: 'Hide wallet/Show wallet',
```
pedroferreira1: We should set this label as 'Hide Wallet' when `mainWindow.isVisible()` is `true` and 'Show Wallet', otherwise

msbrogli:
@pedroferreira1 It seems it is good to merge. Can you just double check running the code, please?

pedroferreira1:
I've tested here and everything is working fine! I will merge it

PR #5, opened 2019-04-15
Adding notification when a new transaction gets to the wallet

pedroferreira1:
Every time a new transaction for this wallet arrives in the ws, the user will receive a notification.

The user can turn on/off the notifications on the Settings screen.


msbrogli:
Can you add some screenshots of both the notification and the settings screen, please?

pedroferreira1:
Screenshot 2019-04-16 at 00 21 26
Screenshot 2019-04-16 at 00 22 59
Screenshot 2019-04-16 at 00 23 27

src/App.js
```diff
@@ -67,6 +67,17 @@ class Root extends React.Component {
       // We are still receiving lot of ws messages that are destined to the admin-frontend and not this wallet
       // TODO separate those messages
       if (wsData.type === 'wallet:address_history') {
+        // If is a new transaction, we send a notification to the user, in case it's turned on
+        if (!wallet.txExists(wsData.history)) {
+          const message = 'New transaction! Click to open'
```
msbrogli: Maybe "You've received a new transaction."

msbrogli:
We could allow users to configure the message using a few variables, such as:
`unread_count`: Number of new transactions since the last access to the wallet
`amount_received`: A string with the sum of the received values, such as `"1.00 HTR"`, `"1.50 HTR and 2.00 TAO"`, `"1.50 HTR, 2.00 TAO, and 0.50 OCC"`, and `"1.50 HTR, 2.00 TAO, and others"`.

They could also customize the title and the message (but not the icon).

What do you think? Maybe we should do it later.

src/utils/wallet.js
```diff
@@ -1498,6 +1499,67 @@ const wallet = {
...
+  },
+
+  /**
+   * Turns notification on
+   *
+   * @memberof Wallet
+   * @inner
+   */
+  turnNotificationOn() {
+    localStorage.setItem('wallet:notification', true);
```
msbrogli: Leave some comment explaining that we do not need to request permission on native apps.

pedroferreira1:
> We could allow users to configure the message using a few variables, such as:
> `unread_count`: Number of new transactions since the last access to the wallet
> `amount_received`: A string with the sum of the received values, such as `"1.00 HTR"`, `"1.50 HTR and 2.00 TAO"`, `"1.50 HTR, 2.00 TAO, and 0.50 OCC"`, and `"1.50 HTR, 2.00 TAO, and others"`.
> 
> They could also customize the title and the message (but not the icon).
> 
> What do you think? Maybe we should do it later.

O like this but think we should improve later. We can see how this goes and with this first use what kind of messages we miss, then we change it later. I will write down in Trello to be done in the future.

src/utils/wallet.js
pedroferreira1: I've added this comment in the `sendNotification` method, which is the one that we should be asking for permission if we had to

src/App.js
pedroferreira1: Changed

src/App.js
```diff
@@ -67,6 +67,17 @@ class Root extends React.Component {
       // We are still receiving lot of ws messages that are destined to the admin-frontend and not this wallet
       // TODO separate those messages
       if (wsData.type === 'wallet:address_history') {
+        // If is a new transaction, we send a notification to the user, in case it's turned on
+        if (!wallet.txExists(wsData.history)) {
+          const message = 'You\'ve received a new transaction! Click to open it.'
+          const notification = wallet.sendNotification(message);
+          // Set the notification click, in case we have sent one
+          if (notification !== undefined) {
+            notification.onclick = () => {
```
obiyankenobi: Can't we have this part inside `wallet.sendNotification`? We'd just have to also pass the `tx_id` as a parameter. I think it makes more sense to handle everything about notifications there.

pedroferreira1: The idea is that the `wallet.sendNotification` would be generic to send a notification for all the system, that's why I send the `message` already formatted for it.

I could send also a callback to be used onclick but I preferred to do this specific for each type of notification. I don't think the `sendNotification` method should receive a `tx_id`, since other notifications won't need it.

obiyankenobi: Sending the callback is a good idea. But we can do it later, when/if we starting using `wallet.sendNotification ` in more places.

pedroferreira1: Ok

PR #6, opened 2019-04-15
Improving README

pedroferreira1:
- Add a section for code architecture;
- Change screenshots for a better quality image


msbrogli:
> I prefer the old pictures, if you just don't scale it down when fetching from google drive (remove w500-h375)

I agree with @obiyankenobi. The previous pictures were better but in low res. Maybe just include them in a high res.

pedroferreira1:
I've updated the pictures, see if it's ok now @msbrogli @obiyankenobi

PR #8, opened 2019-04-18
Fixing performance issue when loading wallet with lots of transactions and addresses

pedroferreira1:
- Fixed performance issue
- Change wallet history pagination to be based and not page based
- Added loading addresses screen, so user can't execute any operation while the wallet is loading data from the full node
Screenshot 2019-04-18 at 16 53 30
Screenshot 2019-04-18 at 16 53 40
Screenshot 2019-04-18 at 16 54 31


src/components/TokenHistory.js
```diff
@@ -165,8 +273,22 @@ class TokenHistory extends React.Component {
       });
     }
 
+    const renderPage = () => {
+      const page = this.getPageNumber();
+      let span = null;
+      if (page === 1) {
+        span = <span>You are seeing transactions in real time</span>;
```
msbrogli: Maybe use `receiving` instead of `seeing` and include a period at the end of the sentence.

msbrogli: If you agree, then you need to update line 282 as well [1].

[1] https://github.com/HathorNetwork/hathor-wallet/pull/8/files#diff-ff7ed0cb94a737efdbdd9453182467c3R282

src/utils/wallet.js
```diff
@@ -1321,27 +1344,33 @@ const wallet = {
...
 
-    return {historyTransactions, allTokens, newSharedAddress, newSharedIndex};
+    return {historyTransactions, allTokens, newSharedAddress, newSharedIndex, addressesFound: lastGeneratedIndex + 1};
   },
 
   /**
    * Check if address is from the loaded wallet
    *
    * @param {string} address Address to check
+   * @param {Object} data Wallet data in localStorage already loaded
```
msbrogli: I think you should explain that it is optional and should be used for performance improvements. If nothing is passed, the data will be loaded again.
We should also make clear that it is expected to receive the output of the `wallet.getWalletData()`.

src/App.js
```diff
@@ -127,9 +132,17 @@ const returnLoadedWalletComponent = (Component, props, rest) => {
   } else {
     // If was closed and is loaded we need to redirect to locked screen
     if (wallet.wasClosed()) {
-      return <Redirect to={{pathname: '/locked/'}} />;
+      return <Redirect to={{ pathname: '/locked/' }} />;
     } else {
-      return returnDefaultComponent(Component, props);
+      if (rest.loadingAddresses) {
+        // If wallet is still loading addresses we redirect to the loading screen
```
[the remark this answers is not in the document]
pedroferreira1: I still don't know how to do this check and show the loading screen without the need to do this `if` in each of the screens. The possibilities I had this was the best one I thought, how would I do it only to show when needed?

Remembering that there are some screens that even if it's loading the addresses, they shouldn't show the loading screen (like the '/server/' screen)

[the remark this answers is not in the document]
msbrogli: I haven't tested this snippet, but I think something like this may work:

```javascript
const WalletLoadedRoutes = ({ match }) => {
    if (this.props.is_loading) {
        return <LoadingScreen />
    }
    return (
        <Switch>
            <Route exact path="/create_token" component={CreateToken} />
            // All other routes can be added here.
        </Switch>
    );
}
```

I did it based on this example https://reacttraining.com/react-router/web/example/basic.

src/components/TokenHistory.js
pedroferreira1: I agree it's better, changed it!

src/utils/wallet.js
pedroferreira1: Added!

src/App.js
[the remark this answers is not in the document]
pedroferreira1: I've used a different approach to refactor this part of the code, check the last commit please.

What do you think? @msbrogli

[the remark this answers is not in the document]
msbrogli: I think it is much better. Thanks!

PR #14, opened 2019-04-24
Fix minor bugs in the wallet

pedroferreira1:
- Adding timeout to requests;
- Aligning in the center the loading addresses screen;
- If an error happens while loading addresses, the user now can change the server without waiting finish loading (that would never finish).

### DIFF
--- a/src/api/axiosInstance.js
+++ b/src/api/axiosInstance.js
@@ -17,10 +17,13 @@ import $ from 'jquery';
  *
  * @module Axios
  */
-const createRequestInstance = (resolve) => {
+const createRequestInstance = (resolve, timeout) => {
+  if (timeout === undefined) {
+    timeout = 10000;
+  }
   const defaultOptions = {
     baseURL: helpers.getServerURL(),
-    timeout: 10000,
+    timeout: timeout,
     headers: {
       'Content-Type': 'application/json',
     },
--- a/src/api/axiosInstance.js
+++ b/src/api/axiosInstance.js
@@ -23,7 +23,7 @@ import $ from 'jquery';
  * Create an axios instance to be used when sending requests
  *
  * @param {callback} resolve Callback to be stored and used in case of a retry after a fail
- * @param {number} timeout Timeout for the request
+ * @param {number} timeout Timeout in milliseconds for the request
  */
 const createRequestInstance = (resolve, timeout) => {
   if (timeout === undefined) {
--- a/src/api/axiosInstance.js
+++ b/src/api/axiosInstance.js
@@ -20,6 +20,7 @@ import $ from 'jquery';
 const createRequestInstance = (resolve) => {
   const defaultOptions = {
     baseURL: helpers.getServerURL(),
+    timeout: 10000,
     headers: {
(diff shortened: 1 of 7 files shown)


src/api/axiosInstance.js
```diff
@@ -17,9 +17,13 @@ import $ from 'jquery';
  *
  * @module Axios
  */
-const createRequestInstance = (resolve) => {
+const createRequestInstance = (resolve, timeout) => {
```
msbrogli: Update the docs. Is it seconds or milliseconds?

src/api/wallet.js
```diff
@@ -45,7 +45,7 @@ const walletApi = {
    */
   sendTokens(txHex, resolve) {
     const postData = {tx_hex: txHex};
-    return createRequestInstance(resolve).post('thin_wallet/send_tokens', postData).then((res) => {
+    return createRequestInstance(resolve, 300000).post('thin_wallet/send_tokens', postData).then((res) => {
```
msbrogli: Maybe it should be a parameter in the config file.

src/api/axiosInstance.js
```diff
@@ -17,9 +17,13 @@ import $ from 'jquery';
...
+  if (timeout === undefined) {
+    timeout = 10000;
```
msbrogli: Maybe it should be a parameter in the config file.

src/components/RequestError.js
```diff
@@ -48,6 +56,9 @@ class RequestErrorModal extends React.Component {
    */
   handleChangeServer = () => {
     $('#requestErrorModal').modal('hide');
+    // When user decides to change the server when an error happened, we must reset the loading information
+    this.props.dataLoaded({addressesFound: 0, transactionsFound: 0});
```
msbrogli: Shouldn't it be done by the loading screen itself? Maybe in the error callback of a promise?

pedroferreira1: Hummm, I think you are right. I should do in the reject of the promise that requests the addressesHistory, what do you think? 

Maybe in this line: https://github.com/HathorNetwork/hathor-wallet/blob/master/src/utils/wallet.js#L164

What do you think?

src/api/axiosInstance.js
pedroferreira1: Done

src/api/axiosInstance.js
pedroferreira1: Done

src/api/wallet.js
pedroferreira1: Done

src/components/RequestError.js
msbrogli: I think the best place would be where the transaction is generated. I think that the error screen must show up, and the download of the address history will be restarted.

Maybe in this line: https://github.com/HathorNetwork/hathor-wallet/blob/master/src/utils/wallet.js#L251

What do you think?

pedroferreira1: You are 100% correct! I was thinking wrongly about the promise chain. Will do this

PR #18, opened 2019-04-26
feat: adapting the error message when request fails because of a 503

pedroferreira1:
Screenshot 2019-04-26 at 04 09 47

### DIFF
--- a/src/api/axiosInstance.js
+++ b/src/api/axiosInstance.js
@@ -42,7 +42,9 @@ const createRequestInstance = (resolve, timeout) => {
     return response;
   }, (error) => {
     // Update error message in redux depending on the status code
-    const statusCode = error.response.status;
+    // Adding conditional because if the server forgets to send back the CORS
+    // headers, error.response will be undefined
+    const statusCode = error.response ? error.response.status : -1;
     if (statusCode === 503) {
       store.dispatch(updateRequestErrorMessage('Our server has reached its requests limit. You should wait a few seconds and try again.'));
     } else {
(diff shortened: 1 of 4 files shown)


src/api/axiosInstance.js
```diff
@@ -41,6 +41,13 @@ const createRequestInstance = (resolve, timeout) => {
   instance.interceptors.response.use((response) => {
     return response;
   }, (error) => {
+    // Update error message in redux depending on the status code
+    const statusCode = error.response.status;
+    if (statusCode === 503) {
+      store.dispatch(updateRequestErrorMessage('Our server has reached its requests limit. You should wait a few seconds and try again.'));
```
msbrogli: I prefer: "Rate limit exceeded. Sorry about that. You should wait a few seconds and try again."
What do you think?

src/components/RequestError.js
```diff
@@ -85,7 +88,7 @@ class RequestErrorModal extends React.Component {
               </button>
             </div>
             <div className="modal-body">
-              <p>Your request failed to reach the server. What do you want to do?</p>
+              <p>{this.props.requestErrorMessage} What do you want to do?</p>
```
msbrogli: Even though "What do you want to do?" appears in both cases, I think we should move it to the `this.props.requestErrorMessage` (and add them to the end of the messages).

src/api/axiosInstance.js
pedroferreira1: I think it's better too, changed it!

src/components/RequestError.js
pedroferreira1: Makes sense, so in the future we can have messages without this part

PR #22, opened 2019-04-30
fix: special character on password pattern was too restrictive

jansegre:


jansegre:
Something is wrong, it should be a single commit. I'll open a new PR.

PR #24, opened 2019-05-09
refactor: migrate utils code for wallet lib

pedroferreira1:


src/App.js
```diff
@@ -59,25 +62,31 @@ const mapStateToProps = (state) => {
...
+    hathorLib.WebSocketHandler.on('addresses_loaded', this.addressesLoadedUpdate);
+    hathorLib.WebSocketHandler.on('is_online', this.isOnlineUpdate);
+    hathorLib.WebSocketHandler.on('reload_data', this.reloadData);
   }
 
   componentWillUnmount() {
-    WebSocketHandler.removeListener('wallet', this.handleWebsocket);
-    WebSocketHandler.removeListener('storage', this.handleWebsocketStorage);
+    hathorLib.WebSocketHandler.removeListener('wallet', this.handleWebsocket);
+    hathorLib.WebSocketHandler.removeListener('storage', this.handleWebsocketStorage);
```
msbrogli: Shouldn't we remove listener for `address_loaded`, `is_online`, and `reload_data` as well?

src/App.js
```diff
@@ -90,21 +99,58 @@ class Root extends React.Component {
             }
           }
         }
-        this.props.historyUpdate({'data': [wsData.history]});
+        wallet.newAddressHistory(wsData.history);
```
msbrogli: Is `hathorLib.` prefix missing?

src/App.js
```diff
@@ -90,21 +99,58 @@ class Root extends React.Component {
...
+  }
+
+  /**
+   * Method called when need to reload data when websocket reconnects
+   */
+  reloadData = (data) => {
+    // Error modal might be opened, then we need to close it when reloading data
+    // Sometimes when the error modal is openend, it leaves only the backdrop after the reload
+    // That's why we need to close it manually
+    $('#requestErrorModal').modal('hide');
```
msbrogli: It is part of another PR, right? Maybe we should remove it from this one.

src/components/ModalAddManyTokens.js
```diff
@@ -8,6 +8,7 @@
 import React from 'react';
 import $ from 'jquery';
 import tokens from '../utils/tokens';
```
msbrogli: Missing import?

src/components/ModalAddToken.js
```diff
@@ -8,6 +8,7 @@
 import React from 'react';
 import $ from 'jquery';
 import tokens from '../utils/tokens';
```
msbrogli: Missing import?

src/components/ModalBackupWords.js
```diff
@@ -7,12 +7,11 @@
 
 import React from 'react';
 import $ from 'jquery';
-import wallet from '../utils/wallet';
-import helpers from '../utils/helpers';
 import _ from 'lodash';
-import { WORDS_VALIDATION } from '../constants';
 import { updateWords } from '../actions/index';
 import { connect } from "react-redux";
+import hathorLib from '@hathor/wallet-lib';
```
msbrogli: You could have used `import { wallet, helpers } from '@hathor/wallet-lib'`, couldn't you?

src/components/TokenBar.js
```diff
@@ -141,9 +140,9 @@ class TokenBar extends React.Component {
 
     const renderUnknownTokens = () => {
       return (
-        <div title={`${unknownTokens} unknown ${helpers.plural(unknownTokens, 'token', 'tokens')}`} className={`d-flex align-items-center icon-wrapper ${this.state.opened ? 'justify-content-start' : 'justify-content-center'}`} onClick={this.unknownClicked}>
+        <div title={`${unknownTokens} unknown ${hathorLib.helpers.plural(unknownTokens, 'token', 'tokens')}`} className={`d-flex align-items-center icon-wrapper ${this.state.opened ? 'justify-content-start' : 'justify-content-center'}`} onClick={this.unknownClicked}>
```
msbrogli: I don't feel that `plural` should be in hathorLib.

src/components/TxRow.js
```diff
@@ -30,7 +30,7 @@ class TxRow extends React.Component {
     return (
       <tr onClick={(e) => this.handleClickTr(this.props.tx.tx_id)}>
         <td className="pr-3">{this.props.tx.tx_id}</td>
-        <td className="pr-3">{dateFormatter.parseTimestamp(this.props.tx.timestamp)}</td>
+        <td className="pr-3">{hathorLib.dateFormatter.parseTimestamp(this.props.tx.timestamp)}</td>
```
msbrogli: I don't feel that `dateFormatter` should be in hathorLib.

src/App.js
pedroferreira1: No, actually I have the utils in the `hathorLib` and utils specific from the wallet. This one is the wallet utils. Do you think we should name it differently?

src/App.js
pedroferreira1: It was in a different file in the other PR but now I have done a different solution, so this code here was removed.

src/components/ModalAddManyTokens.js
pedroferreira1: Actually not, I have some utils methods that come from the `hathorLib` and others that come from the wallet utils itself. I have a `tokens.js` file in the wallet utils. Like the other comment, do you think we should rename it?

src/components/ModalAddToken.js
pedroferreira1: Same as the other above

src/components/ModalBackupWords.js
pedroferreira1: Yes, I could..I decided to always use `hathorLib` to make clear which utils I am using because of the wallet utils.

src/components/TokenBar.js
pedroferreira1: Humm, I agree that shouldn't be in the lib, I just thought that we use it in both desktop and mobile wallet, then we would need to recreate it there.

src/components/TxRow.js
pedroferreira1: The same as the comment above, I agree it's not a lib method but we use in both wallets, so I really don't know what's best.

src/components/ModalBackupWords.js
msbrogli: Ok.

src/components/TokenBar.js
msbrogli: Ok. Make sense.

src/components/TxRow.js
msbrogli: Ok. Make sense.

src/App.js
msbrogli: Ok!

src/components/ModalAddToken.js
msbrogli: I was a little confused, but it makes sense. I think it is ok.

src/App.js
msbrogli: It is funny that I misunderstood it several times. But I think we are good.

src/App.js
```diff
@@ -90,21 +102,54 @@ class Root extends React.Component {
...
 
+  /**
+   * Method called when WebSocket receives a message after loading address history
+   * We just check and save the version that was loaded and update redux data
+   *
+   * @param {Object} data Object with {'historyTransactions', 'addressesFound'}
+   */
+  addressesLoadedUpdate = (data) => {
+    // Update the version of the wallet that the data was loaded
+    localStorage.setItem('wallet:version', VERSION);
```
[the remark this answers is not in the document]
msbrogli: I think we just write it to the `localStorage` to check whether the database is old when a new version is released.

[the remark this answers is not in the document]
pedroferreira1: I need to set after the `App.js` `componentDidMount` method because we need to verify the current wallet version with the last one to see if they are still compatible. That's why every request to load addresses I set it, to guarantee the storage has this version always updated

src/App.js
pedroferreira1: I think you were used with the old code, if we stay confused with this we can change in the future

PR #25, opened 2019-05-09
Fix: download qrcode is getting the correct address now

pedroferreira1:


src/components/ModalAddressQRCode.js
```diff
@@ -90,14 +85,14 @@ class ModalAddressQRCode extends React.Component {
...
-              <QRCode onClick={this.openQrCode} size={200} value={`hathor:${this.props.lastSharedAddress}`} />
+              <QRCode size={200} value={`hathor:${this.props.lastSharedAddress}`} />
               <span ref="address" className="mt-3">
                 {this.props.lastSharedAddress}
                 <CopyToClipboard text={this.props.lastSharedAddress} onCopy={this.copied}>
                   <i className="fa fa-clone pointer ml-1" title="Copy to clipboard"></i>
                 </CopyToClipboard>
               </span> 
-              <a href={this.getHref()} download={`Hathor address - ${this.props.lastSharedAddress}`} className="hidden" ref="downloadLink">Download</a>
+              <a href="true" download={`Hathor address - ${this.props.lastSharedAddress}`} className="hidden" ref="downloadLink">Download</a>
```
msbrogli: Why `"true"`?

pedroferreira1: The linter requires always a value for `href` when using `<a>`, that's why I use `true` for all `<a>` because it has no side effect

PR #27, opened 2019-05-24
fix: modal backdrop now hides when reloading data in ws

pedroferreira1:


src/WebSocketHandler.js
```diff
@@ -72,6 +73,10 @@ class WS extends EventEmitter {
    */
   onOpen = () => {
     if (this.connected === false) {
+      // Modal might be opened, then we need to close it when reloading data
+      // Sometimes when the error modal is openend, it leaves only the backdrop after the reload
+      // That's why we need to close it manually
+      $('#requestErrorModal').modal('hide');
```
msbrogli: It seems like a workaround. I think we may go with this solution, but we should create a card in Trello to look into this and find out why the overlay is left behind. What do you think?

pedroferreira1: I agree and I had created this card in Trello to revisit it, it was only a fast solution to prevent from happening with other users.

However today I thought about another thing and I think this new solution could be the final one. Can you check please and see what you think?

msbrogli: It seems good. Thanks.

PR #28, opened 2019-05-31
refactor: storage

obiyankenobi:
hathor-wallet-lib now requires to be given a storage.

Also, the lib used to open the websocket connection when it was imported. This was changed and we need to manually start websocket connection.

### DIFF
--- a/src/storage.js
+++ b/src/storage.js
@@ -1,0 +1,21 @@
+export const storageFactory = {
+  getItem(key) {
+    return JSON.parse(localStorage.getItem(key));
+  },
+
+  setItem(key, value) {
+    localStorage.setItem(key, JSON.stringify(value));
+  },
+
+  removeItem(key) {
+    localStorage.removeItem(key);
+  },
+
+  clear() {
+    localStorage.clear();
+  },
+
+  key(n) {
+    return localStorage.key(n);
+  },
+}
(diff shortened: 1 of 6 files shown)


src/storage.js
```diff
@@ -0,0 +1,21 @@
+export const storageFactory = {
```
msbrogli: I don't think it is a factory. It seems a regular implementation of the storage. If so, we should rename it to `LocalStorageStore`.

msbrogli: We should also rename the file to `src/store.js`.

src/storage.js
```diff
@@ -0,0 +1,21 @@
...
+  getItem(key) {
+    return JSON.parse(localStorage.getItem(key));
```
msbrogli: What if the item does not exist?

pedroferreira1: `localStorage.getItem` returns `null` and `JSON.parse(null) = null`

src/storage.js
obiyankenobi: Done. Renamed to `MobileStore`.

obiyankenobi: Renamed and transformed into a class, but kept the name `storage.js` so it is not confused with what's inside the folder `store`, which is related to redux.

msbrogli: Ok.

PR #31, opened 2019-06-07
feat: Token management (mint, melt, delegate authority, destroy authority)

pedroferreira1:
image
image
image
image
image


src/screens/TokenDetail.js
```diff
@@ -0,0 +1,685 @@
...
+    if (this.props.historyTransactions !== prevProps.historyTransactions) {
+      this.updateTokenData();
+    }
+  }
+
+  /**
+   * Update token state after didmount or props update
+   */
+  updateTokenData = () => {
+    const filteredHistoryTransactions = hathorLib.wallet.filterHistoryTransactions(this.props.historyTransactions, this.state.token.uid, true);
```
msbrogli: I don't think we need to go through all transactions. We can follow only the ones with at least one authority output, can't we?

src/screens/TokenDetail.js
```diff
@@ -0,0 +1,685 @@
...
+  componentDidUpdate = (prevProps) => {
+    if (this.props.historyTransactions !== prevProps.historyTransactions) {
+      this.updateTokenData();
+    }
+  }
+
+  /**
+   * Update token state after didmount or props update
+   */
+  updateTokenData = () => {
```
msbrogli: This approach seems to fail when another wallet spends your authority utxo because the list will be outdated. Am I missing something?

src/screens/TokenDetail.js
```diff
@@ -0,0 +1,685 @@
...
+  };
+};
+
+
+/**
+ * Screen to manage a token. Mint, melt, edit, unregister, configuration string.
+ *
+ * @memberof Screens
+ */
+class TokenDetail extends React.Component {
```
msbrogli: It seems this class is doing too much. Maybe we should refactor and move the operations to specific components. What do you think?

src/screens/TokenDetail.js
```diff
@@ -0,0 +1,685 @@
...
+
+  /**
+   * Execute destroy method
+   *
+   * @param {Object} array Array of outputs that we are destroying
+   * @param {string} label Label to show on the success message
+   */
+  executeDestroy = (array, label) => {
+    const data = [];
+    // Get the number of outputs the user requested to destroy in the expected format
```
msbrogli: Isn't it better to let the user choose which one he would like to destroy? A possible scenario would be to have delegated the authority to another address in a sub-level of the HD Wallet. Then, he would like to revoke the authority of that wallet for any reason.

src/screens/TokenDetail.js
```diff
@@ -0,0 +1,685 @@
...
+    this.openPinModal();
+  }
+
+  /**
+   * Execute melt method after form validation
+   */
+  executeMelt = () => {
+    const amountValue = this.amount.current.value*(10**hathorLib.constants.DECIMAL_PLACES);
+    const output = this.state.meltOutputs[0];
+    const promise = hathorLib.tokens.meltTokens(output.tx_id, output.index, output.decoded.address, this.state.token.uid, amountValue, this.state.pin, this.createAnother.current.checked);
```
msbrogli: Does it generate and propagate the transaction? Maybe we should split into two methods, one to generate, and another one to generate (calling the first one) and then to propagate. What do you think?

src/screens/TokenDetail.js
```diff
@@ -0,0 +1,685 @@
...
+  }
+
+  /**
+   * Validate form being displayed, then execute the corresponding method, depending on the action selected
+   */
+  validateForm = () => {
+    const isValid = this.form.current.checkValidity();
+    this.setState({ formValidated: true, errorMessage: '' });
+    if (isValid) {
+      switch (this.state.action) {
```
msbrogli: We would avoid it splitting into multiple individual components.

src/screens/UnknownTokens.js
```diff
@@ -63,7 +63,7 @@ class UnknownTokens extends React.Component {
     for (const token of this.props.allTokens) {
       // If has balance but does not have token saved yet
       if (this.props.registeredTokens.find((x) => x.uid === token) === undefined) {
-        const filteredHistoryTransactions = hathorLib.wallet.filterHistoryTransactions(this.props.historyTransactions, token);
+        const filteredHistoryTransactions = hathorLib.wallet.filterHistoryTransactions(this.props.historyTransactions, token, false);
```
msbrogli: The argument list is becoming too big. Maybe we can refactor to send an object instead of individual arguments. Something like:

```
filterHistoryTransactions = ({ arg1, arg2, arg3}) => {
}
```

Then, we would be able to call this method using any order, and we would be forced to name each argument. What do you think?

src/screens/TokenDetail.js
pedroferreira1: How would another wallet spend your authority output? You mean yourself in another device? Like the mobile app?

And even if this happens, it would update the `historyTransactions` in redux, that would call `componentDidUpdate` with new `props` and then `updateTokenData ` will be called to update the state in the screen.

Is this the problem? I am not sure I fully understand your concern

src/screens/TokenDetail.js
pedroferreira1: Yes, that's right. I've updated the code.

src/screens/TokenDetail.js
pedroferreira1: I think that destroying a mint/melt output is an advanced feature and choosing which one to destroy is really advanced. To be honest I would go with the simple approach right now, so we keep the UI/UX simple and write this task to remember to be done in the future. I feel this is not a case that will be common now.

What do you think?

src/screens/UnknownTokens.js
pedroferreira1: I don't really like sending an object for all arguments, it opens the method for a lot of bad uses, I think. It's not difficult to change, but I don't think it's good. Why do you think 4 arguments is too much?

src/screens/TokenDetail.js
pedroferreira1: It's better if in the future we just want the tx before propagating. Done

src/screens/TokenDetail.js
msbrogli: I meant that multiple devices may have access to the same wallet (like when we are sharing the words in our testnet). But your flow seems correct. So, the screen will be updated.

src/screens/TokenDetail.js
msbrogli: I agree. We may go forward as it is. But I think this task may be common for token administrators. We can definitely do it in the future. Let's just create a card for it, please.

src/screens/UnknownTokens.js
msbrogli: You're right. I don't think it it too much. But it is not intuitive what the last argument `false` means. I also agree that it is too flexible, and we can skip my suggestion.

src/screens/TokenDetail.js
pedroferreira1: Done

src/screens/TokenDetail.js
pedroferreira1: @msbrogli I've made a big refactor that creates a component for each of the actions. Can you see if it's ok now?

src/components/TokenDetailAction.js
```diff
@@ -0,0 +1,438 @@
...
+      );
+    }
+
+    const title = `Destroy ${this.props.action === 'destroy-mint' ? 'Mint' : 'Melt'}`;
+
+    return <TokenAction renderForm={renderForm} validateForm={this.destroy} title={title} buttonName='Destroy' onPinSuccess={this.executeDestroy} pinBodyTop={getDestroyBody()} {...this.props} />
+  }
+}
+
+class TokenAction extends React.Component {
```
msbrogli: Add description of what the component does. In this case, I haven't gotten what it does, lol. :)

src/screens/TokenDetail.js
msbrogli: Great. Much easier to review and to understand. I think you should move each of them to a different file. You can put them in `components/tokens/FILE.js`, e.g. `mint.js`, `melt.js`, `destroy.js`, `delegate.js`, etc.

src/components/TokenDetailAction.js
```diff
@@ -0,0 +1,438 @@
...
+    }
+
+    const title = `Delegate ${this.props.action === 'destroy-mint' ? 'Mint' : 'Melt'}`;
+
+    return <TokenAction renderForm={renderForm} title={title} buttonName='Delegate' onPinSuccess={this.executeDelegate} {...this.props} />
+  }
+}
+
+
+class TokenDestroy extends React.Component {
```
msbrogli: Add description of what the component does. In this case, it is used to destroy mint/melt authorities, right?

src/components/TokenDetailAction.js
```diff
@@ -0,0 +1,438 @@
...
+        </div>
+      )
+    }
+
+    return <TokenAction renderForm={renderForm} title='Melt tokens' buttonName='Go' validateForm={this.melt} onPinSuccess={this.executeMelt} {...this.props} />
+  }
+}
+
+
+class TokenDelegate extends React.Component {
```
msbrogli: Add description of what the component does. In this case, it is used to delegate mint/melt authorities to another address, right?

src/components/TokenDetailAction.js
```diff
@@ -0,0 +1,438 @@
...
+          </div>
+        </div>
+      )
+    }
+
+    return <TokenAction renderForm={renderForm} title='Mint tokens' buttonName='Go' validateForm={this.mint} onPinSuccess={this.executeMint} {...this.props} />
+  }
+}
+
+class TokenMelt extends React.Component {
```
msbrogli: Add description of what the component does. In this case, it is used to melt tokens, right?

src/components/TokenDetailAction.js
```diff
@@ -0,0 +1,438 @@
...
+ */
+
+import React from 'react';
+import $ from 'jquery';
+import ModalPin from '../components/ModalPin';
+import ReactLoading from 'react-loading';
+import hathorLib from '@hathor/wallet-lib';
+
+
+class TokenMint extends React.Component {
```
msbrogli: Add description of what the component does. In this case, it is used to mint new tokens, right?

msbrogli: Add description of what the component does. In this case, it is used to mint new tokens, right?

src/components/TokenDetailAction.js
msbrogli: Add description of what the component does. In this case, it is used to melt tokens, right?

src/components/TokenDetailAction.js
msbrogli: Add description of what the component does. In this case, it is used to destroy mint/melt authorities, right?

src/components/TokenDetailAction.js
pedroferreira1: Done

src/screens/TokenDetail.js
pedroferreira1: Done

src/components/TokenDetailAction.js
pedroferreira1: Done

src/components/TokenDetailAction.js
pedroferreira1: Done

src/components/TokenDetailAction.js
pedroferreira1: Done

src/components/TokenDetailAction.js
pedroferreira1: Done

src/components/TokenDetailAction.js
pedroferreira1: Done

src/components/TokenDetailAction.js
pedroferreira1: Done

src/components/TokenDetailAction.js
pedroferreira1: Done

PR #36, opened 2019-07-12
fix: sometimes we show the ServerStatus component and havent received…

pedroferreira1:
… the network data yet

### DIFF
--- a/src/components/ServerStatus.js
+++ b/src/components/ServerStatus.js
@@ -23,10 +23,13 @@ const mapStateToProps = (state) => {
  * @memberof Components
  */
 const ServerStatus = (props) => {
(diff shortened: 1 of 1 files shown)


src/components/ServerStatus.js
```diff
@@ -26,7 +26,7 @@ const ServerStatus = (props) => {
   return (
     props.isOnline !== undefined &&
```
obiyankenobi: Maybe 
`props.isOnline !== undefined &&  props.network && ...`

pedroferreira1: I've updated to something similar but in a way that I think the code is more clear. Can you see if it's good?

obiyankenobi: Yep

PR #37, opened 2019-07-25
chore: connect to new testnet and update hathor wallet lib

obiyankenobi:


obiyankenobi:
> This won't work when resetting the wallet.
> 
> The `resetAllData` method cleans the storage data for the server, then it will start getting the default server from the lib and not from the wallet because the code in `App.js` won't run again.
> 
> That's why I think it's better to create a method to override the default servers from the lib and not only update the storage

I don't like the idea of the lib being responsible for setting this, nor that we should update the lib when we release a new testnet.

What I can do is set the default on the storage again in `utils/wallet.js: resetAllData()`.

@pedroferreira1 @msbrogli

src/App.js
```diff
@@ -42,12 +42,17 @@ import { dataLoaded, isOnlineUpdate } from "./actions/index";
 import store from './store/index';
 import createRequestInstance from './api/axiosInstance';
 import hathorLib from '@hathor/wallet-lib';
-import { VERSION } from './constants';
+import { DEFAULT_SERVER, VERSION } from './constants';
 import LocalStorageStore  from './storage.js';
```
obiyankenobi: Answering the comment here so it's easier to keep all replies in a thread.

> This won't work when resetting the wallet.
> 
> The `resetAllData` method cleans the storage data for the server, then it will start getting the default server from the lib and not from the wallet because the code in `App.js` won't run again.
> 
> That's why I think it's better to create a method to override the default servers from the lib and not only update the storage

I don't like the idea of the lib being responsible for setting this, nor that we should update the lib when we release a new testnet.

What I can do is set the default on the storage again in `utils/wallet.js: resetAllData()`.

@pedroferreira1 @msbrogli

pedroferreira1: I don't like having to manually set the default server on the storage every time. I think the app should have the list of servers (but the lib could also have the default ones) and the lib would get the default server from a method. This method would be implemented in the app, if you want to change the default server.

Similar to what we have in the `axiosInstance`. The lib has the default axios (which is used by the mobile without any changes) and the wallet desktop needs more, so it overrides the `axiosInstance`.

So the lib would have a method `getDefaultServer` (which would be `return DEFAULT_SERVERS[0]`) and the wallet desktop could override this to return whatever it wants.

obiyankenobi: There's a big difference with `axiosInstance` that we don't store or remove it when we reset the wallet.

I think this debate is more about the design of the app/lib. You believe the lib should provide default servers. I think the app should always set the server(s) it will use so there's no question where it's connecting.

Right now we have to set the default server every time because the lib is responsible for storing this, but I think this should not be the case (and we should update the lib to change this in the future, along with other information currently stored by the lib).

pedroferreira1: I just think that lot of users will want to try our lib creating a PoC, or a fast app, just to test maybe and it's weird that they have to decide the URL they want to connect, since they don't know a lot about it.

I just think having some default servers in the lib would help them.

If we decide to remove the default servers from the lib, we should do it now, not in the future. And even so, the solution right now is not working in the case o reloadData

pedroferreira1: Actually I don't think we need to have default servers (a list), only one default server, so we can always connect and use it.

msbrogli: I have mixed feelings about it. On one hand, I think we should have a list of default server because it is easier for the users (maybe we should only explicit require the users to set the servers, such as, `setTestnetServers('bravo')`). On the other hand, we won't provide a free & public full-node forever, so users should choose where they would like to connect.

@pedroferreira1 I think we need to have a list of servers, so the library can automatically try to connect to another one in case of failure, and even include methods to connect to the best one (e.g. the one with lowest RTT).

src/App.js
```diff
@@ -42,12 +42,17 @@ import { dataLoaded, isOnlineUpdate } from "./actions/index";
...
 
 
 hathorLib.storage.setStore(new LocalStorageStore());
 
+// set default server if needed
+if (!wallet.getServerURL()) {
+  hathorLib.wallet.changeServer(DEFAULT_SERVER);
```
msbrogli: I would pass a list of hosts (instead of a single host).

src/App.js
pedroferreira1: Yes, that's true, we would need a list. And I agree that we won't have a public testnet forever but while we have it, we want to make as easy as possible for users to use it, that's why I like the lib being able to connect to a default server.

obiyankenobi: I still feel we should force it to be set, without a default. Let's remember that the user, in this case, is the developer, not the end (app) user.

As a parallel, when using the AWS cli, you always have to set the region you want to connect.

msbrogli: @obiyankenobi Why don't we force user to set the servers, but include the default servers in the `hathor-wallet-lib`. So, users can either use `hathor.useDefaultServers()` or use `hathor.setServers([...])`.

obiyankenobi: I don't think it makes much sense, as as the goal of having a default is using it 'by default', without user intervention.

I'll implement support in the lib to set the server to connect. One other question is if we should also erase the selected server and switch back to the initial one when resetting the wallet.

@pedroferreira1 @msbrogli @jansegre

pedroferreira1: I agree with @obiyankenobi. I think we should have the default in the lib to be used without any additional code and an options to update it.

About switch back the initial one, it's implemented like that because resetting the wallet is also a way of "escaping" an error (when opens the error modal). So if the error is with the server, it's better if we also clean it.

But I do think both cases are useful. Sometimes you want to "go back to factory settings", then resetting everything, including the server. But sometimes you just want to change the wallet words, in the case we could keep the selected server.

Since we don't have both options right now, I think we should keep cleaning the selected server.

src/App.js
obiyankenobi: It only makes sense to do this if the wallet would automatically select the hosts, which we're not doing for now.

src/App.js
obiyankenobi: Opened PR#17 on the lib to enable this functionality.

obiyankenobi: Changes are complete. Can you guys review again?

src/utils/wallet.js
```diff
@@ -457,6 +457,18 @@ const wallet = {
...
+
+  /**
+   * Get the server URL that the wallet will connect. Null if no URL is set
+   *
+   * @return {string} Server URL
+   *
+   * @memberof Wallet
+   * @inner
+   */
+  getServerURL() {
```
pedroferreira1: Are you using this?

obiyankenobi: You're right, this isn't being used. In fact, it's not even doing what it should be.

Removed it.

src/storage.js
```diff
@@ -1,6 +1,20 @@
...
+    let item;
+    try {
+      item = localStorage.getItem(key);
+      return JSON.parse(item);
+    } catch (e) {
+      // old versions of the wallet would save strings without converting
+      // to JSON, so we catch this exception here and return the string directly
+      if (e instanceof SyntaxError) {
+        // first save in JSON format
+        this.setItem(key, item);
```
msbrogli: Even though this solution fixes the issues you've seen, it has not been tested in different scenarios and may fail.

For now, I think we must include a `// FIXME` comment saying that it is a temporary solution with a link to an issue to version the localStorage and include migration tests between different versions. Then, we can go forward with this PR.

pedroferreira1: I agree

obiyankenobi: Yes, the idea is not catching every possible error, just the one we know.

https://github.com/HathorNetwork/hathor-wallet-lib/issues/19

PR #41, opened 2019-08-07
feat: token deposit

obiyankenobi:
Added a message about token deposit in the create and mint tokens screens.

image

image


src/components/tokens/TokenAction.js
```diff
@@ -134,6 +134,7 @@ class TokenAction extends React.Component {
     return (
       <div key={this.props.action}>
         <h2>{this.props.title}</h2>
+        <p>{this.props.subtitle}</p>
```
pedroferreira1: This will add a blank height when there is no `subtitle`. `<p>` tags have `margin-bottom`.

pedroferreira1:
I think that we also should show the amount value that will be deposited in both screens (mint and create token), not only that will be 1%.

When the user changes the amount, it should show the HTR amount required.

src/screens/CreateToken.js
```diff
@@ -249,6 +196,7 @@ class CreateToken extends React.Component {
         <p className="mt-5">Here you will create a new customized token. After the creation, you will be able to send this new token to other addresses.</p>
         <p>Custom tokens share the address space with all other tokens, including HTR. This means that you can send and receive tokens using any valid address.</p>
         <p>Remember to make a backup of your new token's configuration string. You will need to send it to other people to allow them to use your new token.</p>
+        <p>When creating and minting tokens, a <strong>deposit of {hathorLib.tokens.depositPercentage * 100}%</strong> in HTR is required. This will be returned when these tokens are later melted.</p>
```
msbrogli: I would say: "... These will be returned if these tokens are melted later. For further information, see [link]".
I'm not sure about what link to use. Should we point to the RFC? Maybe we should have a document in our website.

src/components/tokens/TokenAction.js
msbrogli: I think we should calculate the deposit in real-time and show to the user.
For example:
It the user typed 120 we would write "A deposit of 1.20 HTR will be made".

obiyankenobi: This is not a problem due to margin collapsing. `<p> margin-bottom` is 'ignored'.

`<p>`
image
`<form>`
image

src/screens/CreateToken.js
obiyankenobi: "When creating and minting tokens, a deposit of 1% in HTR is required. If these tokens are later melted, this HTR deposit will be returned. Read more about it here."

Linked to the RFC as we have no other document yet.

src/components/tokens/TokenAction.js
obiyankenobi: Did that and updated the pictures. Can you review again?

msbrogli: Great. Thanks!

src/screens/CreateToken.js
```diff
@@ -151,23 +107,22 @@ class CreateToken extends React.Component {
...
-        reject(message);
-      });
-    });
 
-    promise.then((token) => {
+    const retPromise = hathorLib.tokens.createToken(
+      address,
+      this.refs.shortName.value,
+      this.refs.symbol.value,
+      parseInt(this.state.amount*(10**hathorLib.constants.DECIMAL_PLACES), 10),
```
msbrogli: We should move it to an utils method `valueFloatToInt(this.state.amount, hathorLib.constants.DECIMAL_PLACES)`.

src/utils/tokens.js
```diff
@@ -75,6 +75,24 @@ const tokens = {
...
+   * Returns the deposit amount in 'pretty' format
+   *
+   * @param {number} mintAmount Amount of tokens to mint
+   *
+   * @memberof Tokens
+   * @inner
+   */
+  getDepositAmount(mintAmount) {
+    if (mintAmount) {
+      const amountValue = mintAmount*(10**hathorLib.constants.DECIMAL_PLACES);
```
msbrogli: Use the suggested util method `valueFloatToInt(...)`.

src/screens/CreateToken.js
obiyankenobi: Done

src/utils/tokens.js
obiyankenobi: Done

obiyankenobi:
Waiting for new lib version so we can update package.json

Issue #43, opened 2019-09-11
Add support for new token structure


msbrogli:
The new token structure is describe in Hathor RFC 4 [1].

The following items should be covered:

1. When registering a new token, validate the token's information from the configuration string with the information on-chain.
1. Do not allow the registration of two tokens with the same symbol.
1. Do not allow the registration of two tokens with the same name.
1. In the tx details screen, we can stop using the UNK symbol for unknown tokens.
1. In the Unknown Tokens screen, we can show token's details.

We should always let the user know whether the token is a registered one or not. If it is not registered and have the same symbol or name as a registered symbol, we should show a fraud alert.

[1] RFC 4 - Token: https://gitlab.com/HathorNetwork/rfcs/blob/token-creation/text/0004-tokens.md

PR #44, opened 2019-09-20
feat: ui improvements with token info

pedroferreira1:
This pull request solves the issue #43.

- Add extra validation when registering a token;
  - Verifying if token name and symbol match with the one saved in the DAG;
  - Check if has already added another token with same name or symbol.
- Showing in wallet history list when a transaction is creating a new token (for the Hathor list shows 'Token deposit')
- Transaction detail:
  - Type can be 'Token creation transaction';
  - No more unknown tokens, now we get the information from the full node when not registered;
  - Warning icon for unregistered tokens;
  - Link to access token detail.
- Token detail:
  - Showing global token information (total supply, if can still mint or melt);
  - If token is not registered in the wallet, show warning icon and icon to register it;
  - Can't edit token name and symbol anymore.

image
image
image
image
image
image


src/screens/TokenDetail.js
```diff
@@ -52,29 +56,67 @@ class TokenDetail extends React.Component {
     walletAmount: 0,
     action: '',
     successMessage: '',
+    totalSupply: 0,
+    canMint: null,
+    canMelt: null,
+    paramUID: null,
+    unknown: null,
```
obiyankenobi: Naming a variable `unknown` seems like we're trolling future developers who'll have to maintain this code 🤣 

I think `isUnknown` would be a better name.

src/screens/TokenDetail.js
```diff
@@ -286,6 +336,47 @@ class TokenDetail extends React.Component {
...
+      }
+    }
+
+    const renderWalletInfo = () => {
+      if (this.state.unknown) {
+        return null;
+      } else {
+        return (
+          <div className="token-detail-wallet-info">
+            <h4 className="mt-3 mb-2"><strong>Wallet data</strong></h4>
```
obiyankenobi: With this new info on the token details screen, we might want to add a link to a page where we explain what each of them mean (which we don't have yet).

We also probably need to add more text to explain that the top part is about the token in general and the bottom (wallet data) is what you have on your wallet.

src/components/TxData.js
```diff
@@ -35,6 +35,12 @@ class TxData extends React.Component {
    */
   state = { raw: false, children: false, tokens: [] };
 
+  // Array of token uid that was already found to show the symbol
+  tokensFound = [];
```
obiyankenobi: I'll add a comment on the backend MR for the tx detail API. I think we can have all this information returned in the API, so there's no need to have any of this here.

pedroferreira1: Actually I don't see how the API could help here, this information depends on the tokens registered on the wallet (information saved locally), so it would be impossible for the API to know.

src/screens/TokenDetail.js
pedroferreira1: I don't really think it's trolling because we have the description of each state, so people can understand it. But I agree that the name could be better, so I think I will change to `tokenUnknown`. What do you think?

I feel like `unknown` and `isUnknown` is pretty much the same

src/screens/TokenDetail.js
pedroferreira1: I don't know if I like this page with the explanation (but maybe we should do it). Maybe just adding more text to this page could solve the doubts and (if we do) some short tutorial videos explaining it.

I tried to make this page as clean as possible, we could try to add also a title to the top part (something like 'General token data' don't know) and an (?) icon with explanation of each one of them.

Really don't know what's best

src/screens/TokenDetail.js
obiyankenobi: `tokenUnknown` sounds good.

src/components/TxData.js
obiyankenobi: Sorry, my comment was on the wrong place. I'm not sure we have an array with all tokens, known and unknown (maybe `state.tokens`).

The idea is that we shouldn't need to call `hathorLib.walletApi.getTokenInfo`. We can return a token's name and symbol in the tx details api.

src/screens/TokenDetail.js
obiyankenobi: I think adding the amount, mint and melt authorities right after the token name at the top and the other could be named 'About this token', after this first part (so inverting the current order).

But, as there's a lot of information on it, I still think we should link to a page where everything is explained (maybe a video like you mentioned).

src/components/TxData.js
pedroferreira1: The API was updated and the call to `hathorLib.walletApi.getTokenInfo` is not needed anymore

src/components/TxData.js
```diff
@@ -396,10 +409,17 @@ class TxData extends React.Component {
...
+          return token.uid;
+        } else {
+          return <Link to={`/token_detail/${token.uid}`}>{token.uid}</Link>
+        }
+      }
       const tokens = this.state.tokens.map((token) => {
         return (
           <div key={token.uid}>
-            <span>{token.name} <strong>({token.symbol})</strong> | {token.uid}</span>
+            <span>{token.name} <strong>({token.symbol})</strong> {token.unknown && <i title='This token is not registered in your wallet.' className='fa text-warning fa-warning'></i>} | {renderTokenUID(token)}</span>
```
msbrogli: I think we should add the warning symbol in the inputs and outputs as well. Many users may not pay attention to the token list below the inputs and outputs.

src/screens/TokenDetail.js
```diff
@@ -286,6 +353,50 @@ class TokenDetail extends React.Component {
       );
     }
 
+    const renderIcons = () => {
+      if (this.state.tokenUnknown) {
+        return (
+          <div>
+            <i className="fa fa-warning text-warning ml-3" title="Token is not registered"></i>
+            <i className="fa fa-check pointer ml-3" title="Register token" onClick={this.registerClicked}></i>
```
msbrogli: I don't like the "correct" symbol next to the warning symbol. Maybe we should open an explanation popup when the user click on the warning. And there may have a register button. What do you think?

src/screens/TokenDetail.js
pedroferreira1: All screens have changed a lot. Can you check again, please?

src/components/TxData.js
pedroferreira1: All screens have changed a lot. Can you check again, please?

src/screens/TokenDetail.js
pedroferreira1: All screens have changed a lot. Can you check again, please?

src/components/TxData.js
```diff
@@ -198,12 +213,67 @@ class TxData extends React.Component {
...
+  /**
+   * Returns if the token from uid in parameter is not registered in the wallet
+   *
+   * @param {string} uid UID of the token to check
+   *
+   * @return {boolean} If token is unknown (not registered)
+   */
+  isTokenUnknown = (uid) => {
+    const tokenConfig = this.state.tokens.find((token) => token.uid === uid);
+    if (tokenConfig === undefined) return false;
```
obiyankenobi: When can this happen? Shouldn't `this.state.tokens` have all tokens from this screen?

src/components/tokens/TokenMelt.js
```diff
@@ -78,6 +78,7 @@ class TokenMelt extends React.Component {
               <label className="form-check-label" htmlFor="keepMint">
                 Create another melt output for you?
               </label>
+              <p className="subtitle">Let it marked, unless you know what you are doing</p>
```
obiyankenobi: Leave it selected unless...

src/components/tokens/TokenMint.js
```diff
@@ -138,6 +138,7 @@ class TokenMint extends React.Component {
               <label className="form-check-label" htmlFor="keepMint">
                 Create another mint output for you?
               </label>
+              <p className="subtitle">Let it marked, unless you know what you are doing</p>
```
obiyankenobi: Leave it selected unless...

src/screens/Wallet.js
```diff
@@ -76,7 +85,55 @@ class Wallet extends React.Component {
...
+        if (!hathorLib.wallet.isAddressMine(output.decoded.address, walletData)) {
+          continue;
+        }
+
+        // This token is not the one of this screen
+        if (output.token !== this.props.selectedToken) {
+          continue;
+        }
+
+        if (hathorLib.wallet.isMintOutput(output) || hathorLib.wallet.isMeltOutput(output)) {
```
obiyankenobi: `walletLib.wallet.isAuthorityOutput`

src/components/TxData.js
```diff
@@ -32,8 +40,17 @@ class TxData extends React.Component {
...
-  state = { raw: false, children: false, tokens: [] };
+  state = {
+    raw: false,
+    children: false,
+    tokens: [],
+    tokenClicked: null,
+  };
+
+  // Array of token uid that was already found to show the symbol
+  tokensFound = [];
```
obiyankenobi: Same comment as https://github.com/HathorNetwork/hathor-explorer/pull/4#discussion_r334092472

src/components/tokens/TokenMelt.js
msbrogli: I think they say "checked" instead of "selected".

src/components/tokens/TokenMint.js
msbrogli: I think they say "checked" instead of "selected".

src/screens/Wallet.js
```diff
@@ -76,7 +85,55 @@ class Wallet extends React.Component {
...
+  unregisterClicked = () => {
+    $('#unregisterModal').modal('show');
+  }
+
+  /**
+   * When user confirms the unregister of the token, hide the modal and execute it
+   */
+  unregisterConfirmed = () => {
+    $('#unregisterModal').modal('hide');
+    tokens.unregisterToken(this.props.selectedToken);
```
msbrogli: Is it fast? Should we add an exception handler?

src/screens/Wallet.js
```diff
@@ -109,10 +166,81 @@ class Wallet extends React.Component {
...
+      return (
+        <div>
+          <p>Are you sure you want to unregister the token <strong>{token.name} ({token.symbol})</strong>?</p>
+          <p>You won't lose your tokens, you just won't see this token on the side bar anymore.</p>
+        </div>
+      )
+    }
+
     const renderUnlockedWallet = () => {
+      const isHathorToken = this.props.selectedToken === hathorLib.constants.HATHOR_TOKEN_CONFIG.uid;
```
msbrogli: It appears in so many places that we should consider creating an util method for it.

src/screens/Wallet.js
```diff
@@ -121,10 +249,18 @@ class Wallet extends React.Component {
       <div>
         {!this.state.backupDone && renderBackupAlert()}
         <div className="content-wrapper">
+         {/* This back button is not 100% perfect because when the user has just unlocked the wallet, it would go back to it when clicked
+           * There is no easy way to get the previous path
+           * I could use a lib (https://github.com/hinok/react-router-last-location)
+           * Or handle it in our code, saving the last accessed screen
+           * XXX Is it worth it to do anything about it just to prevent this case?
+           */}
+          <BackButton {...this.props} />
```
msbrogli: I didn't understand the problem. Can you give me an example, please?

pedroferreira1: 1. Wallet is locked (user is in the locked screen)
2. User unlocks the wallet and goes to the main screen (token history)
3. User clicks 'Back' and goes to locked screen again.

To prevent this case we have to install a lib, so we can have access to the last path.

src/components/tokens/TokenMint.js
pedroferreira1: Updated to 'Leave it checked, unless'

src/components/tokens/TokenMelt.js
pedroferreira1: Updated to 'Leave it checked unless'

src/components/TokenGeneralInfo.js
```diff
@@ -0,0 +1,182 @@
...
+    transactionsCount: 0,
+  };
+
+  componentDidMount() {
+    this.updateTokenInfo();
+  }
+
+
+  componentDidUpdate = (prevProps) => {
+    if (this.props.historyTransactions !== prevProps.historyTransactions || this.props.token.uid !== prevProps.token.uid) {
```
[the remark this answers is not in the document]
obiyankenobi: I think we can move on and improve in the future, if we feel it's needed.

src/components/ModalUnregisteredTokenInfo.js
```diff
@@ -0,0 +1,134 @@
...
+      if (!this.state.token) {
+        return null;
+      }
+      return <TokenGeneralInfo token={this.state.token} showConfigString={false} />;
+    }
+
+    const renderHeader = () => {
+      return (
+        <div className="d-flex flex-row">
+          <h5 className="modal-title" id="exampleModalLabel">{this.state.token.name} ({this.state.token.symbol})</h5>
```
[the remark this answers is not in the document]
pedroferreira1: No, removed it, thanks!

src/components/ModalUnregisteredTokenInfo.js
```diff
@@ -0,0 +1,134 @@
...
+          <span className='ml-2 unregistered-token-badge'> Unregistered token </span>
+        </div>
+      )
+    }
+
+    return (
+      <div className="modal fade" id="unregisteredTokenInfoModal" tabIndex="-1" role="dialog" aria-labelledby="unregisteredTokenInfoModal" aria-hidden="true">
+        <div className="modal-dialog modal-lg" role="document">
+          <div className="modal-content">
+            <div className="modal-header">
```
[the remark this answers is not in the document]
pedroferreira1: It can, it was just preventing any problems in case we don't have a token is state. But this case is 'impossible'

src/components/ModalUnregisteredTokenInfo.js
```diff
@@ -0,0 +1,134 @@
...
+              {this.state.token && renderHeader()}
```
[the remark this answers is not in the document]
pedroferreira1: I shouldn't do this for the whole render method, since I feel the modal should open while the state is being updated. But I will do it only once to render the modal content.

src/components/ModalUnregisteredTokenInfo.js
```diff
@@ -0,0 +1,134 @@
...
+              <button type="button" className="close" data-dismiss="modal" aria-label="Close">
+                <span aria-hidden="true">&times;</span>
+              </button>
+            </div>
+            <div className="modal-body">
+              {renderModalBody()}
```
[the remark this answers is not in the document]
pedroferreira1: Makes sense. Updated method name

src/screens/Wallet.js
pedroferreira1: This is true only right now, in the last version of our full node we had another authority output and in the next one we might have. Do you think it's worth it to change it just to prevent doing this `||` and we need to remember to change this in case we create another authority output?

In my opinion we should keep like this.

src/components/TokenAdministrative.js
```diff
@@ -269,62 +225,47 @@ class TokenDetail extends React.Component {
...
   }
 }
 
-export default connect(mapStateToProps)(TokenDetail);
+
+/*
+ * token: Token to show administrative tools {name, symbol, uid}
+ */
+TokenAdministrative.propTypes = {
+  token: PropTypes.object.isRequired
```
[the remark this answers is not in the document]
pedroferreira1: Done!

src/components/TokenGeneralInfo.js
```diff
@@ -0,0 +1,182 @@
...
+    )
+  }
+}
+
+/*
+ * token: Token to show the information {name, symbol, uid}
+ * showConfigString: If should show the configuration string of the token with the qrcode
+ */
+TokenGeneralInfo.propTypes = {
+  token: PropTypes.object.isRequired,
```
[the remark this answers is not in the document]
pedroferreira1: Done

src/components/TxData.js
```diff
@@ -198,12 +211,67 @@ class TxData extends React.Component {
...
+   * Returns if the token from uid in parameter is not registered in the wallet
+   *
+   * @param {string} uid UID of the token to check
+   *
+   * @return {boolean} If token is unknown (not registered)
+   */
+  isTokenUnknown = (uid) => {
+    const tokenConfig = this.state.tokens.find((token) => token.uid === uid);
+    if (tokenConfig === undefined) return false;
+    return tokenConfig.unknown;
```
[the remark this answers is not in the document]
pedroferreira1: We set it here: https://github.com/HathorNetwork/hathor-wallet/blob/8f5eda5a28a9f82d3c6111c7908e3a651449a52a/src/components/TxData.js#L121

But this object is completed with unknown key here: https://github.com/HathorNetwork/hathor-wallet/blob/8f5eda5a28a9f82d3c6111c7908e3a651449a52a/src/components/TxData.js#L144

And we use it here: https://github.com/HathorNetwork/hathor-wallet/blob/8f5eda5a28a9f82d3c6111c7908e3a651449a52a/src/components/TxData.js#L495

src/components/TxData.js
```diff
@@ -396,10 +480,19 @@ class TxData extends React.Component {
     }
 
     const renderTokenList = () => {
+      const renderTokenUID = (token) => {
+        if (token.uid === hathorLib.constants.HATHOR_TOKEN_CONFIG.uid) {
+          return token.uid;
```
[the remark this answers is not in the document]
pedroferreira1: Makes sense, to keep it consistent. Updated it

src/screens/Wallet.js
pedroferreira1: Lib code:
```
  unregisterToken(uid) {
    const tokens = this.getTokens();
    const filteredTokens = tokens.filter((token) => token.uid !== uid);
    this.saveToStorage(filteredTokens);
    return filteredTokens;
  },  
```
Wallet code after lib code: (only updating with new array of tokens)
```
store.dispatch(newTokens({tokens, uid: hathorLib.constants.HATHOR_TOKEN_CONFIG.uid}));
```

In my opinion it's really fast and only local (no need for a loading) and no error possible. What do you think?

src/screens/Wallet.js
obiyankenobi: Sure, let's keep like this.

src/components/TxData.js
pedroferreira1: Yes. I was just doing extra validation in case something really weird happens. Because if by any chance `tokenConfig` it would raise js exception and the wallet would stop.

I feel like it's extra care but doesn't hurt having it. What do you think?

src/components/TxData.js
pedroferreira1: They are both different cases because in the wallet we need to know if the token is unknown, so:

1. I was using it before because we weren't rendering authority inputs/outputs, so no need to add to the array a token that had only authorities (but it has changed now);
2. Here in the wallet I need to validate if each token is unknown and I use this `tokensFound` array not to execute this check if I have more than one input/output for the same token. 

So I think here in the wallet is necessary (I could have done this differently - e.g., using a set but this also works) but the explorer might have a refactor. What do you think?

obiyankenobi: Ok, no problem.

src/screens/Wallet.js
msbrogli: The `saveToStorage` may be slow or fail for multiple reasons. We may return a promise or let the general exception handler report any problem.

src/components/TxData.js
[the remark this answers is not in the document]
msbrogli: Thanks!

src/components/TokenGeneralInfo.js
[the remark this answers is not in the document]
msbrogli: I agree. Seems good for now.

src/screens/Wallet.js
msbrogli: I see. Maybe you can include this example in the comment to make it easier for future readers.

About the problem itself, can't the lock screen just check that the wallet is unlocked and "go back again", instead of adding a new dependency?

pedroferreira1: I can try it, but feel like the page will 'blink' and I really don't like it. But maybe not. Will try it.

src/components/TxData.js
obiyankenobi: No problem.

src/screens/Wallet.js
msbrogli: Make sense. Blinking is an undesirable side effect. Your solution seems better.

src/components/TokenGeneralInfo.js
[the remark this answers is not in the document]
pedroferreira1: @msbrogli I need one check here. You've suggested to use the exact object on this prop types, however I use this component in two different places.

1. About token: it's perfect using the exact;
2. TransactionDetail (when opening the unregistered token modal): in this page I already use a token object that has {'name', 'uid', 'symbol', 'unknown'} and reuse it to send to this component (`TokenGeneralInfo`). However this is not the exact object we are expecting (we have the 'unknown' as a plus and it throws a warning.

I thought about two solutions:

1. Send to the component an updated version of the object, where I remove the 'unknown' key;
2. Update the prop types, to use shape instead of exact (https://reactjs.org/docs/typechecking-with-proptypes.html), so I can have a 'superobject' also in this props, with all needed keys and some optional more.

I feel the best option is the first, so we define what we want and all components must adjust but I don't know if the code will look like a workaround removing a key to send an object as props.

What do you think?

src/screens/Wallet.js
pedroferreira1: I've added the error handling when unregistering the token. Can you see if it's good?

Wallet commit: https://github.com/HathorNetwork/hathor-wallet/pull/44/commits/a7194bf589a6f2c935c3d16e863cd620f024472b
Lib commit: https://github.com/HathorNetwork/hathor-wallet-lib/pull/25/commits/2751ad630fa9382ad02bc09742aed7e21c6299bd

I will create a PR for the mobile app also now to add this handling.

msbrogli: Great! Thanks!

src/components/TokenGeneralInfo.js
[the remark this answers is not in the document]
msbrogli: I think it is ok to use the `shape` instead of `exact`. I don't see any problem to get "superobjects", and I think it is simpler and less error prone than creating "partial objects".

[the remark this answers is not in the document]
pedroferreira1: I've updated to use `shape`.

We have now:

```
TokenGeneralInfo.propTypes = { 
  token: PropTypes.shape({
    name: PropTypes.string.isRequired,
    symbol: PropTypes.string.isRequired,
    uid: PropTypes.string.isRequired,
  }), 
  showConfigString: PropTypes.bool.isRequired,
}
```

pedroferreira1:
Closes https://github.com/HathorNetwork/hathor-wallet/issues/43

PR #47, opened 2019-10-22
chore: first QA draft for desktop wallet

pedroferreira1:


pedroferreira1:
In this QA there aren't many tests to validate errors, like trying to register the same token twice, use a wrong configuration string, or spend the same input twice.

Should we add it?

QA.md
```diff
@@ -0,0 +1,130 @@
+# Test Sequence
+
+### Those tests must be done in an Electron application. Some features might work on the web browser but show different behavior when using electron.
```
obiyankenobi: These tests

QA.md
```diff
@@ -0,0 +1,130 @@
...
+
+- **New token error**
+1. [] Click on 'Custom tokens', then 'Create a new token'.
+1. [] Type 'Test token' in the token name, 'Test' in the symbol and 100 in the amount.
+1. [] Validates that the deposit required is 1.00 HTR.
+1. [] Try to create the token and receive an error that you don't have enough HTR.
+
+- **Wallet screen**
+1. [] It shows a red label with the testnet name in the top right corner of the screen.
+1. [] Copy the address and send HTR to it.
```
obiyankenobi: Before this step, the user needs some balance. It's important that we make it clear, as in the last section we expected him to have none (when creating a token).

QA.md
```diff
@@ -0,0 +1,130 @@
...
+1. [] Click on the transaction that has the token deposit (this is the transaction responsible for creating the new token).
+1. [] Validate it has an input of HTR, an output of the total amount created for the new token, a mint and melt authority output and it might have a change HTR output (depending on the input spent).
+1. [] Validate it has the new created token in the token list.
+1. [] Click on the token uid and should redirect to the wallet main screen with TST selected.
+
+- **Send tokens**
+1. [] Copy the address.
+1. [] Click on 'Send tokens'.
+1. [] Send an amount of HTR and TST to the copied address.
+1. [] The first transaction on the list must have amount 0.00.
```
obiyankenobi: The first transaction on the list must have amount 0.00 (for both HTR and TST).

obiyankenobi:
> In this QA there aren't many tests to validate errors, like trying to register the same token twice, use a wrong configuration string, or spend the same input twice.
> 
> Should we add it?

I think so, specially trying to register *different* tokens with _same_ name and symbol.

msbrogli:
> In this QA there aren't many tests to validate errors, like trying to register the same token twice, use a wrong configuration string, or spend the same input twice.
> 
> Should we add it?

I think we should. Maybe include sections of tests?

msbrogli:
It seems you need to replace `[]` by `[ ]` to let Markdown render it as a checkbox.

pedroferreira1:
> It seems you need to replace `[]` by `[ ]` to let Markdown render it as a checkbox.

That's true, thanks!

Another thing, in the other QA PR (https://github.com/HathorNetwork/hathor-wallet-mobile/pull/8) even the section title is numbered. I think we should follow the same model but I prefer not having a number on the section title.

What you think? @msbrogli @obiyankenobi @jansegre

QA.md
pedroferreira1: Actually this is the step where it receives the first transaction. I don't see why the user needs balance before it.

> Copy the address and send HTR to it.

pedroferreira1: Ahhh, I think I understand now the confusion. I need to make clear you have to send this HTR here from another wallet, isn't it? Will add this

obiyankenobi: Yes. I thought the tester would be sending tokens to himself.

pedroferreira1: Done, thanks

pedroferreira1:
@msbrogli @obiyankenobi added more tests.

- Try to register token with invalid configuration string;
- Register different token with same name;
- Try to send token selecting an input that was already used.

Do you see anything else to add?

obiyankenobi:
> @msbrogli @obiyankenobi added more tests.
> 
> * Try to register token with invalid configuration string;
> * Register different token with same name;
> * Try to send token selecting an input that was already used.
> 
> Do you see anything else to add?

Sounds good for now.

QA.md
```diff
@@ -0,0 +1,148 @@
+# Test Sequence
+
+### These tests must be done in an Electron application. Some features might work on the web browser but show different behavior when using electron.
+
+1. **Initialization**
+    1. Try to click 'Get started' without selecting checkbox (must show error).
+    1. Start new wallet.
+    1. Try to click 'Create my words' without selecting checkbox (must show error).
+    1. Select to do backup now. (Really save those words, we will load this wallet later with them).
+    1. Click 'Validate' before selecting all the words.
```
msbrogli: We need to say what to expect. In this case, an error message should appear.

QA.md
```diff
@@ -0,0 +1,148 @@
...
+    1. Click 'Validate' with all words selected but in the wrong order.
```
msbrogli: We need to say what to expect. In this case, an error message should appear.

QA.md
```diff
@@ -0,0 +1,148 @@
...
+    1. Click on 'Wallet' in the navigation bar, select the HTR token and copy the address (and save it).
+    1. Click on the first transaction (the one with amount 0.00).
+    1. Copy the first input from HTR and from your address (copy the transaction ID and the index).
+    1. Click on 'Send tokens'.
+    1. Paste the address and 1.00 as the amount. Deselect 'Choose inputs automatically' and paste the transaction ID and index copied before.
+    1. Must see an error that the output is already spent.
+
+1. **Late backup**
+    1. Reset wallet and start a new one without doing backup. It must show a yellow warning saying a backup must be done.
+    1. Do the backup (following procedures in the 'Initialization' tests). The backup have to disappear.
```
msbrogli: Typo: ... The backup **message** have to disappear.

PR #55, opened 2019-12-04
style: updating new logo and new color

pedroferreira1:
image
image
image
image


src/index.scss
```diff
@@ -378,7 +385,7 @@ svg.svg-wrapper {
 }
 
 .outside-content-wrapper {
-  background-color: $darkBlue;
+  background-color: $purpleHathor;
```
obiyankenobi: Did you try the Welcome Screen with a black background?

pedroferreira1: I've just uploaded an image with this option. I think it's better as a background

obiyankenobi: I also preferred it.

PR #58, opened 2020-01-02
feat: wallet now will be connected to the mainnet by default

pedroferreira1:
image


src/screens/Welcome.js
```diff
@@ -43,14 +43,14 @@ class Welcome extends React.Component {
               <img className="hathor-logo" src={logo} alt="" />
               <p className="mt-4 mb-4">Welome to Hathor Wallet!</p>
             </div>
-            <p className="mb-4">This wallet is connected to a <strong>testnet</strong>.</p>
-            <p>Your HTR and other tokens may be reset at any time.</p>
-            <p>If one offers to sell some tokens to you, one is a scammer.</p>
+            <p className="mb-4">This wallet is connected to the <strong>mainnet</strong>.</p>
```
msbrogli: Isn't it better to have two Welcome Screens? One for each version of the wallet? Eventually we will publish a Wallet for testnet, right?

src/screens/Welcome.js
```diff
@@ -43,14 +43,14 @@ class Welcome extends React.Component {
...
+            <p>In this wallet you can send tokens and see the history of your transactions. Besides that, you can also create and manage new tokens.</p>
```
msbrogli: Suggestion: Using this wallet you can (i) check your balance and history, (ii) send & receive HTR and other tokens running on Hathor, and (iii) create and manage your own tokens.

src/screens/Welcome.js
```diff
@@ -43,14 +43,14 @@ class Welcome extends React.Component {
...
+            <p>You are responsible for your tokens and the safety of them.</p>
```
msbrogli: Suggestion: You should never share your seed with other people. Anyone with access to your seed have full control of your balance.

package.json
```diff
@@ -21,10 +21,10 @@
   "productName": "Hathor Wallet",
   "description": "Light wallet for Hathor Network",
   "author": "Hathor Labs <contact@hathor.network> (https://hathor.network/)",
-  "version": "0.10.0",
+  "version": "0.11.0",
   "private": true,
   "dependencies": {
-    "@hathor/wallet-lib": "^0.7.0",
+    "@hathor/wallet-lib": "^0.8.0",
     "@sentry/electron": "^0.17.0",
```
obiyankenobi: Should we remove Sentry?

src/screens/Welcome.js
```diff
@@ -43,14 +43,14 @@ class Welcome extends React.Component {
...
             <p>For further information, check our website (https://hathor.network/).</p>
           <form ref="agreeForm" className={`w-100 mb-4 ${this.state.formValidated ? 'was-validated' : ''}`}>
             <div className="form-check">
               <input required type="checkbox" className="form-check-input" id="confirmAgree" />
-              <label className="form-check-label" htmlFor="confirmAgree"> I agree to participate in the testnet of Hathor, and I acknowledge that the tokens are not for real.</label>
+              <label className="form-check-label" htmlFor="confirmAgree"> I understand that I am responsible for my wallet and the tokens.</label>
```
msbrogli: Suggestion: I understand that I am fully responsible for keeping my seed secret and for the custody of my tokens.

src/screens/Welcome.js
obiyankenobi: I'd use the same wallet and do as Trezor. If someone wants to use testnet, he should change the server address.

src/screens/Welcome.js
obiyankenobi: At this point we didn't reach the seed part. Maybe talking about it here would be a bit confusing.

src/screens/Welcome.js
obiyankenobi: Same as above about seed.

src/screens/Welcome.js
pedroferreira1: I agree, that's one of the reasons I haven't mentioned the words here. Maybe even talk about the "safety of tokens" is confusing but I feel it's good so we can have the link to the website if someone wants to read more.

package.json
pedroferreira1: I think it's good to remove I just don't know if we should do it now, or in the next release

obiyankenobi: My opinion is that we should remove for the mainnet wallet.

pedroferreira1: I think I feel the same. I was just thinking what could go wrong removing it now but I guess it's not complicated. I could remove it.

@msbrogli what do you think?

msbrogli: I don't we have to disable it by default and skip the initial screen requesting to enable it. But we should keep the choice in the settings, so, if any user has a problem, they can enable it to let us know the details.

src/screens/Welcome.js
msbrogli: Hum, seems good. The downside is that naive users may change the configuration to the testnet and be scammed.

Maybe changing the server to the testnet must have a huge alert about the risks and scammers.

src/screens/Welcome.js
msbrogli: You're right. Maybe: "You should never share nothing besides your addresses to other people. You're fully responsible for keeping your tokens safe."

What do you think?

src/screens/Welcome.js
msbrogli: Yeah. Maybe: "I understand that I am fully responsible for keeping my tokens safe and that it is not possible to revert transactions after they are executed."

src/screens/Welcome.js
obiyankenobi: I'd just tweat: "You should never share **_any information_** besides"

src/screens/Welcome.js
obiyankenobi: Looks good.

package.json
obiyankenobi: Agreed.

src/screens/Welcome.js
msbrogli: Great! Thanks!

src/screens/Welcome.js
pedroferreira1: I agree. I will continue with this PR for now and create and issue to handle this

package.json
pedroferreira1: Done

PR #64, opened 2020-01-14
feat: prevent selecting lock block rewards as inputs of a tx

pedroferreira1:


src/screens/SendTokens.js
```diff
@@ -91,27 +91,33 @@ class SendTokens extends React.Component {
...
-      }, (message) => {
-        this.setState({ errorMessage: message, loading: false });
-      });
-    } catch(e) {
-      if (e instanceof hathorLib.errors.AddressError || e instanceof hathorLib.errors.OutputValueError) {
-        this.setState({ errorMessage: e.message, loading: false });
-      } else {
-        // Unhandled error
-        throw e;
+    const promiseData = this.getData();
```
obiyankenobi: Mixing `async/await` methods with promises. We should use only one standard. Should open an issue for this?

PR #65, opened 2020-01-15
feat: updating redux data with height and using it when calculating b…

pedroferreira1:
…alance

### DIFF
--- a/src/reducers/index.js
+++ b/src/reducers/index.js
@@ -47,6 +47,8 @@ const initialState = {
   addressesFound: 0,
   // Quantity of transactions already loaded to give a feedback to the user
   transactionsFound: 0,
+  // Height of the network arrived from ws data
+  height: 0,
 };
 
 const rootReducer = (state = initialState, action) => {
@@ -91,6 +93,8 @@ const rootReducer = (state = initialState, action) => {
       return Object.assign({}, state, {addressesFound: action.payload.addressesFound, transactionsFound: action.payload.transactionsFound});
     case 'update_request_error_status_code':
       return Object.assign({}, state, {requestErrorStatusCode: action.payload});
+    case 'update_height':
+      return Object.assign({}, state, {height: action.payload.height});
     default:
       return state;
   }
--- a/src/reducers/index.js
+++ b/src/reducers/index.js
@@ -47,7 +47,7 @@ const initialState = {
   addressesFound: 0,
   // Quantity of transactions already loaded to give a feedback to the user
   transactionsFound: 0,
-  // Height of the network arrived from ws data
+  // Height of the best chain of the network arrived from ws data
   height: 0,
 };
 
--- a/src/screens/CreateToken.js
+++ b/src/screens/CreateToken.js
@@ -19,11 +19,13 @@ import hathorLib from '@hathor/wallet-lib';
 const mapStateToProps = (state) => {
   const balance = hathorLib.wallet.calculateBalance(
     Object.values(state.historyTransactions),
-    hathorLib.constants.HATHOR_TOKEN_CONFIG.uid
+    hathorLib.constants.HATHOR_TOKEN_CONFIG.uid,
+    state.height
   );
   return {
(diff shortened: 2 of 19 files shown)


src/screens/Wallet.js
```diff
@@ -32,12 +32,13 @@ const mapDispatchToProps = dispatch => {
...
 const mapStateToProps = (state) => {
   const filteredHistoryTransactions = hathorLib.wallet.filterHistoryTransactions(state.historyTransactions, state.selectedToken, true);
-  const balance = hathorLib.wallet.calculateBalance(filteredHistoryTransactions, state.selectedToken);
+  const balance = hathorLib.wallet.calculateBalance(filteredHistoryTransactions, state.selectedToken, state.height);
   return {
     balance: balance,
     historyTransactions: filteredHistoryTransactions,
     selectedToken: state.selectedToken,
     tokens: state.tokens,
+    height: state.height,
```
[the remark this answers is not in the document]
pedroferreira1: This code has changed and does not have it anymore. Can you see again please?

src/screens/CreateToken.js
```diff
@@ -19,11 +19,13 @@ import hathorLib from '@hathor/wallet-lib';
...
   const balance = hathorLib.wallet.calculateBalance(
     Object.values(state.historyTransactions),
-    hathorLib.constants.HATHOR_TOKEN_CONFIG.uid
+    hathorLib.constants.HATHOR_TOKEN_CONFIG.uid,
+    state.height
   );
   return {
     htrBalance: balance.available,
     historyTransactions: state.historyTransactions,
+    height: state.height,
```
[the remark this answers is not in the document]
pedroferreira1: The code has changed a lot, can you see again please?

src/reducers/index.js
```diff
@@ -47,6 +47,8 @@ const initialState = {
   addressesFound: 0,
   // Quantity of transactions already loaded to give a feedback to the user
   transactionsFound: 0,
+  // Height of the network arrived from ws data
```
msbrogli: I think we should explain better what height of the network means because there are at least two different definitions.

src/reducers/index.js
```diff
@@ -47,6 +47,8 @@ const initialState = {
...
+  height: 0,
```
msbrogli: It seems we are not using it anywhere in the PR.

src/reducers/index.js
```diff
@@ -91,6 +93,8 @@ const rootReducer = (state = initialState, action) => {
       return Object.assign({}, state, {addressesFound: action.payload.addressesFound, transactionsFound: action.payload.transactionsFound});
     case 'update_request_error_status_code':
       return Object.assign({}, state, {requestErrorStatusCode: action.payload});
+    case 'update_height':
+      return Object.assign({}, state, {height: action.payload.height});
```
obiyankenobi: From what I've seen, this won't update the locked/unlocked balance automatically, right?

pedroferreira1: For what we have here: https://react-redux.js.org/using-react-redux/connect-mapstate

> It is called every time the store state changes.

So every time the wallet receives an update on the height it will update the redux and call all `mapStateToProps` again, which updates the balance of each screen.

I will double check now only that all `calculateBalance` are called from `mapStateToProps`

src/reducers/index.js
pedroferreira1: We need it on redux so it will call `mapStateToProps` when it changes, so I can calculate the balance again.

src/reducers/index.js
obiyankenobi: Ah ok, it checks again all state variables. Thanks

src/reducers/index.js
pedroferreira1: Improved the comment explaining that we are getting the height of the best chain

src/reducers/index.js
msbrogli: Ok. Thanks!

PR #86, opened 2020-02-11
fix: crash while loading

### LINKED ISSUE #85: Wallet crashes when loading thousands of transactions
Discord miner reported this to @jansegre . He's got around 15k blocks.

```
Error Message: Cannot use 'in' operator to search for '0000000000001c6b53869ca44f6e0b376712148afb5d177df4f78f5a10920a65' in undefined
Stack trace: TypeError: Cannot use 'in' operator to search for '0000000000001c6b53869ca44f6e0b376712148afb5d177df4f78f5a10920a65' in undefined
    at Object.txExists (file:///C:/Users/1/AppData/Local/Programs/hathor-wallet/resources/app.asar/build/static/js/2.0069d731.chunk.js:2:458406)
    at t.handleWebsocket (file:///C:/Users/1/AppData/Local/Programs/hathor-wallet/resources/app.asar/build/static/js/main.31c35dba.chunk.js:1:189246)
    at t.f.emit (file:///C:/Users/1/AppData/Local/Programs/hathor-wallet/resources/app.asar/build/static/js/2.0069d731.chunk.js:2:479484)
    at t.value (file:///C:/Users/1/AppData/Local/Programs/hathor-wallet/resources/app.asar/build/static/js/2.0069d731.chunk.js:2:2628528)
    at WebSocket.ws.onmessage (file:///C:/Users/1/AppData/Local/Programs/hathor-wallet/resources/app.asar/build/static/js/2.0069d731.chunk.js:2:2628098)
```

lib PR: https://github.com/HathorNetwork/hathor-wallet-lib/pull/52
wallet PR: https://github.com/HathorNetwork/hathor-wallet/pull/86

obiyankenobi:
Closes #85

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -42,12 +42,12 @@ import store from './store/index';
 import createRequestInstance from './api/axiosInstance';
 import hathorLib from '@hathor/wallet-lib';
 import { DEFAULT_SERVER, VERSION } from './constants';
-import LocalStorageStore  from './storage.js';
+import { HybridStore } from './storage.js';
 
 hathorLib.network.setNetwork('mainnet');
-hathorLib.storage.setStore(new LocalStorageStore());
+hathorLib.storage.setStore(new HybridStore());
 
-// set default server to bravo testnet
+// set default server
 hathorLib.wallet.setDefaultServer(DEFAULT_SERVER);
 
 const mapDispatchToProps = dispatch => {
--- a/src/App.js
+++ b/src/App.js
@@ -145,9 +145,6 @@ class Root extends React.Component {
     // Update the version of the wallet that the data was loaded
     hathorLib.storage.setItem('wallet:version', VERSION);
 
-    // Check api version everytime we load address history
-    version.checkApiVersion();
-
     // Update redux with loaded data
     store.dispatch(dataLoaded({ addressesFound: data.addressesFound, transactionsFound: Object.keys(data.historyTransactions).length }));
   }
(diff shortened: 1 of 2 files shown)


src/App.js
```diff
@@ -158,6 +155,10 @@ class Root extends React.Component {
    * @param {boolean} data Boolean if websocket is online
    */
   isOnlineUpdate = (data) => {
+    if (data) {
+      // Check api version everytime we connect to the server
+      version.checkApiVersion();
```
msbrogli: On Hathor Mobile Wallet, we check the API version before marking the connection as online. Should we do the same here?

src/storage.js
```diff
@@ -32,4 +41,46 @@ class LocalStorageStore {
...
+ * there are thousands of txs on the wallet history (we got this error with 15k txs) and the localStorage
+ * does not store data after a certain limit (it fails silently).
+ *
+ * In theory, there shouldn't be a limit for localStorage with Electron, as it's been patched here: https://github.com/electron/electron/pull/15596.
+ * However, this other comment (https://github.com/electron/electron/issues/13465#issuecomment-494983533) suggests
+ * there was still a problem.
+ *
+ * It's not a problem if 'wallet:data' is not persisted, as this data is always loaded when we connect to the server.
+ */
+class HybridStore {
```
msbrogli: Nice workaround!

src/App.js
obiyankenobi: The reason I remove this from `addressesLoaddedUpdate` is that we don't need to check the api version every time we get a new tx. In my tests, I had 400 addresses with 1 tx each. During wallet load, we'd call `checkApiVersion` 400 times in a short period.

We could change it to mark as online only after this check, but I didn't want to change the current behavior. I'm not sure if there'd be any other side effects of doing this. To be honest, this is not needed to fix the bug. It just seemed something simple that could improve load, without incurring any other changes.

msbrogli: I agree. Thanks!

PR #87, opened 2020-02-11
feat: remove explorer menu dropdown and add link to public explorer

### LINKED ISSUE #73: Remove 'Explorer' from wallet and add link to Public Explorer
This would prevent an amount of duplicated code we currently have.

pedroferreira1:
Closes #73


src/components/Navigation.js
```diff
@@ -11,14 +11,38 @@ import { t } from 'ttag';
...
  * @memberof Components
  */
 class Navigation extends React.Component {
+
+  /**
+   * Method called when user clicked on Esxplorer menu
+   *
+   * @param {Object} e Event for the click
+   */
+  goToExplorer = (e) => {
```
obiyankenobi: Could we have an util method for opening web pages? It seems it's needed in some different places.

obiyankenobi: Maybe just an issue to do it later.

obiyankenobi: Looks good now!

PR #89, opened 2020-02-14
fix: endless loading after wallet update (0.12.0 -> 0.13.0)

obiyankenobi:


package.json
```diff
@@ -70,7 +69,7 @@
     "@sentry/browser": "^5.0.5",
     "@sentry/cli": "^1.40.0",
     "electron": "^5.0.3",
-    "electron-builder": "^22.1.0",
+    "electron-builder": "^20.43.0",
```
obiyankenobi: Downgraded this as newer versions have an error running Wine.

src/utils/wallet.js
```diff
@@ -304,6 +304,18 @@ const wallet = {
       tokens: dataToken,
     }));
 
+    // before cleaning data, check if we need to transfer xpubkey to wallet:accessData
+    const accessData = hathorLib.wallet.getWalletAccessData();
+    if (accessData.xpubkey === undefined) {
+      // XXX from v0.12.0 to v0.13.0, xpubkey changes from wallet:data to access:data.
+      // That's not a problem if wallet is being initialized. However, if it's already
+      // initialized, we need to set the xpubkey in the correct place.
+      const oldData = JSON.parse(localStorage.getItem('wallet:data'));
```
msbrogli: What if `wallet:data` is undefined? Have you tested initializing a brand new wallet?

package.json
msbrogli: This change came from another commit. But I think it is worth discussing.

pedroferreira1: I think you upgraded it because the new macOS version requires it but in this version there is an error to generate Windows package (that's what I remember but I might be wrong). So in macOS version of @obiyankenobi notebook it's not needed.

obiyankenobi: @pedroferreira1 is correct. This is the version that works for generating Windows builds. I remember you used to have this error @msbrogli . That's why I'm the only one that can generate packages nowadays.

What would happen is that I had to manually change it on `package.json` before generating the packages. Sometimes I'd forget and hit the error in the Windows build, which takes a while. So i thought it made more sense to downgrade ot.

src/utils/wallet.js
obiyankenobi: This situation cannot happen. Only chance `wallet:data` is undefined in `localStorage` is if the user already has the newest version. In that case, he wouldn't enter this `if`.

Yes, I tested initializing a new wallet.

msbrogli: Ok. Thanks!

package.json
msbrogli: Wine stopped running when only 64 bit apps were allowed by macOS. Ok, seems safe to downgrade it.

obiyankenobi: I also generate the `.pkg` file if we want to publish.

PR #93, opened 2020-02-21
fix: menu mobile is not transparent anymore

### LINKED ISSUE #92: fix: menu mobile is transparent
One of the users had a problem because didn't see the menu opened, so the items were above the screen, which made impossible to use the wallet.

image

pedroferreira1:
image

Close #92

### DIFF
--- a/src/index.scss
+++ b/src/index.scss
@@ -299,6 +299,18 @@ svg.svg-wrapper {
     padding: 0.5rem 1rem;
   }
 
+  .main-nav nav.navbar .navbar-brand {
+    padding-left: 1rem;
+  }
+
+  .main-nav nav.navbar {
+    background-color: #000;
+  }
+
+  .navigation-search .version-wrapper:first-child {
+    padding: 0;
+  }
+
   .wallet-wrapper {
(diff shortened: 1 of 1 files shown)


src/index.scss
```diff
@@ -299,6 +299,18 @@ svg.svg-wrapper {
...
+  }
+
+  .main-nav nav.navbar {
+    background-color: #000;
+  }
+
+  .navigation-search .version-wrapper:first-child {
+    padding: 0;
+  }
+
```
obiyankenobi: Actually, I vote for removing mobile support from Desktop wallet.

msbrogli: I agree. This code came from hathor-explorer and is not important here. I don't think we should remove support, but we don't have to worry about it.

obiyankenobi: I'm confused.

> I agree

> I don't think we should remove support

Do you agree to remove the responsive code from the wallet? Given the bug it caused (see linked issue), I don't think we can just ignore it.

obiyankenobi: Ok, let's keep mobile support

PR #100, opened 2020-03-31
feat: List of addresses from the wallet

pedroferreira1:
- [x] Remove debug logs;
- [x] Add comments in methods/states/propTypes;
- [x] Add link to explorer address search screen on address row click.

Unresolved questions:

1. Should we have a new item on the menu? At first I tried to add some link/button close to the address box in the main wallet screen but none of my attempts were good, so I just added this new menu.
2. I created a new pagination because we already have all the elements on the list and I always think it's better to have all the pages there (if it's possible). Should we keep the same pagination as in the other screens with 'Previous'/'Next'?
3. Should we show all addresses including the ones from the `GAP_LIMIT`. I think we should because maybe the user has generated new addresses, sent to someone but still haven't received the txs, in that case none of those addresses would appear on the list. However, showing all addresses might be confusing to the user.

image
image

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -42,6 +42,7 @@ import { DEFAULT_SERVER, IPC_RENDERER, VERSION } from './constants';
 import { HybridStore } from './storage.js';
 import ModalAlert from './components/ModalAlert';
 import SoftwareWalletWarningMessage from './components/SoftwareWalletWarningMessage';
+import AddressList from './screens/AddressList';
 
 hathorLib.network.setNetwork('mainnet');
 hathorLib.storage.setStore(new HybridStore());
@@ -96,7 +97,9 @@ class Root extends React.Component {
     hathorLib.WebSocketHandler.removeListener('is_online', this.isOnlineUpdate);
     hathorLib.WebSocketHandler.removeListener('reload_data', this.reloadData);
 
-    IPC_RENDERER.removeAllListeners("ledger:closed");
+    if (IPC_RENDERER) {
+      IPC_RENDERER.removeAllListeners("ledger:closed");
+    }
   }
 
   handleWebsocket = (wsData) => {
@@ -192,6 +195,7 @@ class Root extends React.Component {
         <StartedRoute exact path="/wallet/passphrase" component={ChoosePassphrase} loaded={true} />
         <StartedRoute exact path="/server" component={Server} loaded={true} />
         <StartedRoute exact path="/transaction/:id" component={TransactionDetail} loaded={true} />
+        <StartedRoute exact path="/addresses" component={AddressList} loaded={true} />
         <StartedRoute exact path="/new_wallet" component={NewWallet} loaded={false} />
         <StartedRoute exact path="/load_wallet" component={LoadWallet} loaded={false} />
         <StartedRoute exact path="/wallet_type" component={WalletType} loaded={false} />
--- a/src/components/Navigation.js
+++ b/src/components/Navigation.js
@@ -55,9 +55,6 @@ class Navigation extends React.Component {
                 <NavLink to="/custom_tokens/" exact className="nav-link" activeClassName="active" activeStyle={{ fontWeight: 'bold' }}>{t`Custom tokens`}</NavLink>
               </li>
               <li className="nav-item">
-                <NavLink to="/addresses/" exact className="nav-link" activeClassName="active" activeStyle={{ fontWeight: 'bold' }}>{t`Addresses`}</NavLink>
-              </li>
-              <li className="nav-item">
                 <a className="nav-link" href="true" onClick={this.goToExplorer}>{t`Public Explorer`}</a>
               </li>
             </ul>
--- a/src/components/Navigation.js
+++ b/src/components/Navigation.js
@@ -55,6 +55,9 @@ class Navigation extends React.Component {
                 <NavLink to="/custom_tokens/" exact className="nav-link" activeClassName="active" activeStyle={{ fontWeight: 'bold' }}>{t`Custom tokens`}</NavLink>
               </li>
               <li className="nav-item">
+                <NavLink to="/addresses/" exact className="nav-link" activeClassName="active" activeStyle={{ fontWeight: 'bold' }}>{t`Addresses`}</NavLink>
+              </li>
+              <li className="nav-item">
                 <a className="nav-link" href="true" onClick={this.goToExplorer}>{t`Public Explorer`}</a>
               </li>
             </ul>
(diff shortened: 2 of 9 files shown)


msbrogli:
Instead of used, I'd include four columns: Number of Transactions, Total Received, Total Sent, and Balance. If it is too much, we can keep just Number of Transactions and Balance. What do you think?

msbrogli:
I'm not sure whether it should be a menu. Maybe we can put a "Show all addresses" next to the "Address to receive tokens" box.

msbrogli:
About the GAP limit, we can include an explanation that the last 20 addresses must be empty. We could also add a separator like "------- (show the last 20 addresses) -------", toggling to hide when clicked.

obiyankenobi:
I wonder if it's useful for users to know the index. I think it might be more confusing. 

Regarding where to access this page, I don't think we should place it on the menu or the main page. This looks like something people won't use often. I'd rather have just a link on the settings page, even though it's not a setting option.

msbrogli:
@obiyankenobi The index is useful to generate the address derivation path. Maybe the user might have a way to see the full path.

pedroferreira1:
> Instead of used, I'd include four columns: Number of Transactions, Total Received, Total Sent, and Balance. If it is too much, we can keep just Number of Transactions and Balance. What do you think?

@msbrogli I don't like it for the same reason we made the explorer 'Address Search' page so complex. An address might have many tokens and the balance is different for each other. The best option would be to have an address screen similar to the one from the explorer, but this is lot of work for now. I've changed the 'used' column for the 'number of transactions' column and added a link to the explorer, so the user can see all the details of the address. What do you think? 

> I don't think we should place it on the menu or the main page. This looks like something people won't use often. I'd rather have just a link on the settings page, even though it's not a setting option.

I don't know about the Settings page. I don't like it on the menu either and I think the best place would be close to the address box. I will try some options and paste here so we can discuss. If we don't find a good fit, I think the Settings is a better place than the menu.

> What about a toggle do show/hide unused addresses? We could also have a "?" icon next to it with more information.

The problem is not the unused addresses, the problem to me is the addresses from the gap limit. That most users won't use but some will, so it's gonna be weird for most of them to see 20 unused addresses in the end and some will be okay because they will understand. Maybe we should just add this '?' explaining about the last 20 unused addresses.

pedroferreira1:
Options of places to add a button/link to go to Address List screen.

1. 
image

2. 
image

3. 
image

4. 
image

5. 
image

What do you think? @msbrogli @obiyankenobi

msbrogli:
> @msbrogli I don't like it for the same reason we made the explorer 'Address Search' page so complex. An address might have many tokens and the balance is different for each other. The best option would be to have an address screen similar to the one from the explorer, but this is lot of work for now. I've changed the 'used' column for the 'number of transactions' column and added a link to the explorer, so the user can see all the details of the address. What do you think?

I agree, and I think your solution is good. Thanks!

msbrogli:
Number 3 is the one I like the most, as the screenshot below:

image

src/App.js
```diff
@@ -96,7 +97,9 @@ class Root extends React.Component {
     hathorLib.WebSocketHandler.removeListener('is_online', this.isOnlineUpdate);
     hathorLib.WebSocketHandler.removeListener('reload_data', this.reloadData);
 
-    IPC_RENDERER.removeAllListeners("ledger:closed");
+    if (IPC_RENDERER) {
```
msbrogli, line 100: Is it a bug fix? Maybe we should do it in another PR.

src/components/Navigation.js
```diff
@@ -54,6 +54,9 @@ class Navigation extends React.Component {
               <li className="nav-item">
                 <NavLink to="/custom_tokens/" exact className="nav-link" activeClassName="active" activeStyle={{ fontWeight: 'bold' }}>{t`Custom tokens`}</NavLink>
               </li>
+              <li className="nav-item">
+                <NavLink to="/addresses/" exact className="nav-link" activeClassName="active" activeStyle={{ fontWeight: 'bold' }}>{t`Addresses`}</NavLink>
```
msbrogli, line 58: If we go with number 3, we should remove this change.

src/screens/AddressList.js
```diff
@@ -0,0 +1,213 @@
...
+
+  /**
+   * Return total number of transactions using an address
+   *
+   * @param {String} address Address to get the total number of transactions
+   * @param {Object} historyTransactions History of transactions from the wallet data storage
+   *
+   * @return {Number} Total number of transactions using this address
+   */
+  getNumberOfTransactions = (address, historyTransactions) => {
```
msbrogli, line 75: This algorithm seems really slow for big wallets because we are iterating all transactions multiple times, one for each address. I think it can receive the `addressKeys` and iterate only once storing the counters in a dictionary. Then, it returns the dictionary. Something like this:

```
counter = {};
for (const key in addressKeys) {
  counter[key] = 0;
}
```

Then, instead of `if (el.decoded.address === address)`, it just have to check if the key exists `if (el.decoded.address in counter)`.

obiyankenobi:
I don't think option 3 is bad, but the problem I have with placing "See all addresses" there is that we're exposing something a bit more advanced, that people don't usually need to see. As we're debating about the gap addresses, it might lead to people having more doubts and feeling insecure/unsafe. That's why I think it should be placed on Settings page.

pedroferreira1:
As discussed in the meeting today, we will go with number 3 and won't change anything for the GAP limit unused addresses.

src/App.js
pedroferreira1: It's not exactly a bug fix. This raises an error only when running in the browser (so only for dev). I would prefer not to have a PR only for this but if you think it's important, I can do it.

src/components/Navigation.js
pedroferreira1: Removed

src/App.js
msbrogli: I got it. No, I think it is good.

src/screens/AddressList.js
pedroferreira1: Updated the code. Can you see if it's good?

src/screens/AddressList.js
```diff
@@ -0,0 +1,218 @@
...
+   * @return {Number} Total number of pages of the list
+   */
+  getTotalPages = (array) => {
+    return Math.ceil(array.length / WALLET_HISTORY_COUNT);
+  }
+
+  /**
+   * Update number of transactions for each address
+   *
+   * @param {Object} addressData Transactions by address object {'address': 0}
```
msbrogli, line 75: I think `{'address': 0}` may be misleading. I'd use `{[address]: 0}`.

pedroferreira1: Changed

PR #105, opened 2020-04-14
refactor: small changes to support new lib version

### LINKED ISSUE #39: WebSocket connects when wallet starts
It should only connect after the user starts the wallet.

If the wallet is not started yet, the ws is open during the init screens and will be closed and immediately re-opened when we initialize the wallet.

I'm not sure if it should also be closed when wallet is locked (right now ws is open).

pedroferreira1:
Closes #39

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -66,7 +66,6 @@ const mapStateToProps = (state) => {
 
 class Root extends React.Component {
   componentDidMount() {
-    hathorLib.WebSocketHandler.setup();
     hathorLib.WebSocketHandler.on('wallet', this.handleWebsocket);
     hathorLib.WebSocketHandler.on('storage', this.handleWebsocketStorage);
 
@@ -103,7 +102,7 @@ class Root extends React.Component {
   }
 
   handleWebsocket = (wsData) => {
-    if (hathorLib.wallet.loaded()) {
+    if (hathorLib.wallet.loaded() && !this.props.loadingAddresses) {
       // We are still receiving lot of ws messages that are destined to the admin-frontend and not this wallet
       // TODO separate those messages
       if (wsData.type === 'wallet:address_history') {
@@ -180,7 +179,9 @@ class Root extends React.Component {
    * Method called when need to reload data when websocket reconnects
    */
   reloadData = (data) => {
-    wallet.reloadData();
+    if (hathorLib.wallet.loaded() && !this.props.loadingAddresses) {
+      wallet.reloadData();
+    }
   }
 
   render() {
--- a/src/App.js
+++ b/src/App.js
@@ -102,7 +102,7 @@ class Root extends React.Component {
   }
 
   handleWebsocket = (wsData) => {
-    if (hathorLib.wallet.loaded() && !this.props.loadingAddresses) {
+    if (hathorLib.wallet.loaded()) {
       // We are still receiving lot of ws messages that are destined to the admin-frontend and not this wallet
       // TODO separate those messages
       if (wsData.type === 'wallet:address_history') {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1141,9 +1141,9 @@
       }
     },
     "@hathor/wallet-lib": {
-      "version": "0.11.1",
-      "resolved": "https://registry.npmjs.org/@hathor/wallet-lib/-/wallet-lib-0.11.1.tgz",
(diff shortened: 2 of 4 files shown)


src/App.js
```diff
@@ -66,7 +66,6 @@ const mapStateToProps = (state) => {
 
 class Root extends React.Component {
   componentDidMount() {
-    hathorLib.WebSocketHandler.setup();
```
msbrogli, line 69: Why don't we have to call it anymore?

src/App.js
```diff
@@ -103,7 +102,7 @@ class Root extends React.Component {
   }
 
   handleWebsocket = (wsData) => {
-    if (hathorLib.wallet.loaded()) {
+    if (hathorLib.wallet.loaded() && !this.props.loadingAddresses) {
```
msbrogli, line 105: What changed in the lib that we need this new condition?

src/App.js
```diff
@@ -180,7 +179,9 @@ class Root extends React.Component {
    * Method called when need to reload data when websocket reconnects
    */
   reloadData = (data) => {
-    wallet.reloadData();
+    if (hathorLib.wallet.loaded() && !this.props.loadingAddresses) {
```
msbrogli, line 182: The same question here: what changed in the lib that we need this if statement?

src/App.js
pedroferreira1: We don't need the websocket ready when we mount the app, only when we start the wallet.

We are already calling `setup` in the lib when starting the wallet.

src/App.js
pedroferreira1: This was already a bug, in my opinion.

The reload is called every time we move to the online state and it was not online before. However, it makes no sense to reload the wallet data before loading it, or while we are loading it.

src/App.js
pedroferreira1: This was already a bug, in my opinion.

Just like the if added in the `reloadData` websocket message, it makes no sense to handle a ws message (like a new tx message) if it's still loading data.

src/App.js
obiyankenobi: By the way, we have issue #39 describing this problem.

src/App.js
msbrogli: I thought `hathorLib.wallet.loaded()` was checking exactly that, but it only checks if the wallets was initialized, right? It is not related to the connection and its state.

src/App.js
msbrogli: About the first condition (`hathorLib.wallet.loaded()`), I agree. But the second condition should never happen right? What if the connection is down while the wallet is syncing (i.e., loading the transaction address)? Wouldn't it prevent the wallet from starting it over?

src/App.js
msbrogli: Great. Thanks! We should add to the description of the issue that it solves #39.

src/App.js
msbrogli: I think we should handle messages while we are loading the addresses, otherwise a race condition may occur. For example: Let's say the wallet has only one address, and we just requested it to the server.

1) Wallet requests the transaction list
2) The server receives the request and sends the response
3) A new transaction arrives in the server, and a message is sent in the websocket.
4) The websocket message arrives in the wallet.
5) The response arrives in the wallet.

Notice that, even thought the response was sent before, it arrived later. We cannot guarantee it would arrive first because the HTTP and the Websocket are two different connections. In this case, if we discard the websocket message, we would show the wrong balance to the user.

src/App.js
pedroferreira1: If the connection is down while the wallet is syncing it will show the error modal with the options:

- Retry request
- Reset wallet
- Change server

The automatically reload data shouldn't be possible while loading the data from the server, only after the wallet is already loaded.

src/App.js
pedroferreira1: Yes, the `hathorLib.wallet.loaded()` checks only if the wallet has an `xpub`, i.e. if it was already created.

I agree with you, it's important to handle new messages while loading the addresses.

Removed that condition from the `if`.

msbrogli: Ok. Thanks!

src/App.js
msbrogli: I got it. Just to double check: have you tested connection issues to confirm that the wallet can still connect?

pedroferreira1: What do you mean about 'connection issues'? I've tested becoming offline and then online again and the data is reloaded normally.

msbrogli: As we discussed in Slack, it is ok. Thanks!

PR #112, opened 2020-04-30
Improving UX when sending tokens splitting loading message into two steps

pedroferreira1:


src/screens/SendTokens.js
```diff
@@ -244,7 +252,13 @@ class SendTokens extends React.Component {
       // Must update the shared address, in case we have used one for the change
       wallet.updateSharedAddress();
       $('#alertModal').modal('hide');
```
obiyankenobi, line 254: The message is shown in the modal, right? Shouldn't we also hide it inside `setTimeout`?

pedroferreira1: It's shown in the modal only on ledger. We should improve how to test ledger wallet changes.

But you are right, I've moved the hide execution to inside the `setTimeout`

PR #114, opened 2020-05-08
fix: user shouldnt be able to see wallet type screen for now

pedroferreira1:
Steps to reproduce:

1. Go to some initial screen (not the welcome) before loading the wallet (e.g. backup screen);
2. Close the wallet app;
3. Open the wallet, then it will start on  wallet type screen (to select between software and hardware wallet)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -279,7 +279,9 @@ const returnStartedRoute = (Component, props, rest) => {
       }
     } else {
       if (rest.loaded) {
-        return <Redirect to={{pathname: '/wallet_type/'}} />;
+        // When the wallet is opened, the path that is called is '/', which currenctly redirects to the Wallet component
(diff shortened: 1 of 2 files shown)


src/App.js
```diff
@@ -279,7 +279,9 @@ const returnStartedRoute = (Component, props, rest) => {
       }
     } else {
       if (rest.loaded) {
-        return <Redirect to={{pathname: '/wallet_type/'}} />;
+        // When the wallet is opened, the path that is called is '/', which currenctly redirects to the Wallet component
+        // in that case, if the wallet is not loaded but it's started, it should redirect to the signin/wallet type screen
+        return <Redirect to={{pathname: '/signin/'}} />;
```
obiyankenobi, line 284: Worth adding a test to QA.

pedroferreira1: Done

PR #119, opened 2020-05-27
feat: using tx mining API to send transactions

pedroferreira1:
It's still a draft because there are some things to be done:

- [x] Add comments and docstring;
- [x] Add propTypes in new components;
- [x] Handle some error cases (marked with TODO in the code);
- [x] Test sending transaction with ledger.


src/screens/SendTokens.js
```diff
@@ -8,18 +8,16 @@
...
 import { connect } from "react-redux";
 import BackButton from '../components/BackButton';
 import hathorLib from '@hathor/wallet-lib';
 import wallet from '../utils/wallet';
-import colors from '../index.scss';
 import ModalAlertNotSupported from '../components/ModalAlertNotSupported';
 import ModalAlert from '../components/ModalAlert';
 import ledger, { LedgerError } from '../utils/ledger';
-import { IPC_RENDERER } from '../constants';
+import { MIN_JOB_ESTIMATION, MIN_POLL, IPC_RENDERER } from '../constants';
```
obiyankenobi, line 20: `MIN_JOB_ESTIMATION, MIN_POLL` not used.

src/components/SendTxHandler.js
```diff
@@ -0,0 +1,92 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import { t } from 'ttag';
+import hathorLib from '@hathor/wallet-lib';
```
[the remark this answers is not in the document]
pedroferreira1: I left because I was going to use (now I've added the prop-types and it's needed), sorry.

src/components/ModalSendTx.js
```diff
@@ -0,0 +1,176 @@
...
+    } else {
+      this.refs.formPin.classList.remove('was-validated');
+      const pin = this.refs.pinInput.refs.pin.value;
+      if (hathorLib.wallet.isPinCorrect(pin)) {
+        $('#pinModal').data('bs.modal')._config.backdrop = 'static';
+        $('#pinModal').data('bs.modal')._config.keyboard = false;
+        const sendTransaction = this.props.prepareSendTransaction(pin);
+        if (sendTransaction) {
+          // Show send tx handler component and start sending
+          this.setState({ step: 1, loading: true, sendTransaction });
```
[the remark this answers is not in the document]
pedroferreira1: I agree. Removed from the state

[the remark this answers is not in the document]
obiyankenobi: Still there. Maybe you forgot to push?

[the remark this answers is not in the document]
pedroferreira1: Yes, I forgot to push. Updated now

src/components/SendTxHandler.js
```diff
@@ -0,0 +1,138 @@
...
+    const renderBody = () => {
+      if (this.state.errorMessage) {
+        return (
+          <p className="text-danger mt-3 white-space-pre-wrap">{this.state.errorMessage}</p>
+        )
+      } else {
+        return (
+          <div className="d-flex flex-column">
+            <p>{this.state.loadingMessage}</p>
+            <p>{this.state.miningEstimation && renderMiningEstimation()}</p>
```
msbrogli, line 117: Should we add an elapsed time?

pedroferreira1: We already have the estimation time going decreasing, another time increasing would be confusing for the user in my opinion.

What do you think? @msbrogli @obiyankenobi @jansegre

obiyankenobi: I agree.

jansegre: I think we could decrease it on every second, and re-adjust when the server updates. This could turnout very simple if we use the estimated time returned to calculated a "willCompleteAt" absolute time, and the frontend only has to always displays the seconds until "willCompleteAt" regardless of what is updating it.

msbrogli: As the tx-mining-service always returns a fixed expected time, I agree that we can decrease it on every second. If it reaches zero, we can just write "soon" as suggested by @jansegre .

obiyankenobi: Before we make any changes, we should settle the matter of the mining estimate. If we can't get a reliable number, this is pointless.

jansegre: I agree with @obiyankenobi. We could hide/comment/disable this UI feature and use it when time estimations are at least decent. I don't think we even need to make any calculations, if we launch it without estimates and measure how long it is taking for a few weeks, we can get an idea of the mining time distribution in practice and make a static guess, and improve in the future.

msbrogli: I agree with @obiyankenobi and @jansegre. We can disable this UX feature and release a new version. The log will tell us the expected time vs. the total time, and we can use it to improve the estimator later on. When the estimator is good enough, we enable it again.

pedroferreira1: I've removed the time estimation from the UI and when it's good enough we can add again.

PR #127, opened 2020-06-24
fix: prevent blank screen when full node version check request fails

pedroferreira1:


src/screens/Server.js
```diff
@@ -73,6 +73,9 @@ class Server extends React.Component {
     hathorLib.wallet.changeServer(newServer);
     const promise = version.checkApiVersion();
     promise.then(() => {
+      // The user might need to change the server before unlocking the wallet when the server he was connected to is down
+      // So after change we unlock the wallet because the PIN was already requested and validated
+      hathorLib.wallet.unlock();
```
msbrogli, line 78: I don't understand why we need to unlock again. If the PIN has already been validated, wouldn't it be unlocked?

msbrogli:
I still haven't understood what the root cause is in the first place.

src/index.js
```diff
@@ -19,6 +19,14 @@ import './index.css';
 import store from "./store/index";
 import { Provider } from "react-redux";
 
+import hathorLib from '@hathor/wallet-lib';
+import createRequestInstance from './api/axiosInstance';
+// This code used to be on the componentDidMount of App.js file
+// However there is a checkVersion request that is executed before the screen is mounted for the first time
```
obiyankenobi, line 25: Where is it? The call on `returnLoadedWalletComponent` on `App.src`?

src/App.js
```diff
@@ -232,7 +229,13 @@ const returnLoadedWalletComponent = (Component, props, rest) => {
   } else {
     // If was closed and is loaded we need to redirect to locked screen
     if (hathorLib.wallet.wasClosed()) {
-      return <Redirect to={{ pathname: '/locked/' }} />;
+      if (isServerScreen) {
+        // If there is a server problem before the wallet is unlocked
+        // the server screen must be the priority over the locked screen
+        return returnDefaultComponent(Component, props);
```
obiyankenobi, line 235: Not sure when this would happen. If wallet was closed and the user was on the server screen, t won't be on the server screen next time it reopens.

Also, he should only be able to go to the server screen after unlocking the wallet.

pedroferreira1:
# Problem

When you open the wallet and the server you were connected to does not reply the `/version/` request. This happens because this first request is executed before we set the custom wallet axios (that handles all errors).

image

## Solution (part 1)

Set the custom axios instance before anything happens on the app. I moved the code below from `src/App.js` to `src/index.js`.

```
hathorLib.axios.registerNewCreateRequestInstance(createRequestInstance);
```

This already fixes the bug because it will show the error modal and the user can change the server. However it will be like in the video below.

gif1

1. Open the wallet and the request error happens, so open error modal
2. Clicks to change server
3. Opens Locked screen because is the only screen allowed after you open the wallet
4. Unlock the wallet
5. Call `/version/` again, the same error happens and the modal is opened again;
6. Click to change server
7. Now success

## Solution (part 2)

If the wallet was opened I am allowing it to go to the server screen, in case an error happened. (see code below)

```
      if (isServerScreen) {
        // If there is a server problem before the wallet is unlocked
        // the server screen must be the priority over the locked screen
        return returnDefaultComponent(Component, props);
```

gif2

About @msbrogli comment

> I don't understand why we need to unlock again. If the PIN has already been validated, wouldn't it be unlocked?

`hathorLib.wallet.unlock();` just removes the 'wallet:locked' localStorage key and this is usually one when unlocking the wallet. However in that case we are unlocking it from the Change Server screen, so I am calling it.

obiyankenobi:
Ok, I got the idea. I'm wondering if we should actually be doing the version check before the user enters the pin to unlock the wallet. It seems to me this should only happen when the wallet is unlocked and we actually try to connect to the server.

I tracked it and it happens on `returnLoadedWalletComponent` on `App.js`, when we create the routes. To be honest, I don't remember why we added this call here. We already check the api version every time we connect to the websocket and that should be enough.

This is how I imagine things should work without this version check when creating the routes.

Scenario 1: server is down
1. user enters pin and unlocks the wallet;
2. call to load wallet will fail and error modal is shown;
3. user can change server;

Scenario 2: server is up but versions don't match
1. user enters the pin and unlocks the wallet;
2. websocket will connect and trigger a version check;
3. versions won't match and we'll be displayed the Version screen;

Does it make sense?

PS - is the user presented the Version error screen before entering the pin in case we are connected to a server with an old version?

msbrogli:
> Ok, I got the idea. I'm wondering if we should actually be doing the version check before the user enters the pin to unlock the wallet. It seems to me this should only happen when the wallet is unlocked and we actually try to connect to the server.
> 
> I tracked it and it happens on `returnLoadedWalletComponent` on `App.js`, when we create the routes. To be honest, I don't remember why we added this call here. We already check the api version every time we connect to the websocket and that should be enough.
> 
> This is how I imagine things should work without this version check when creating the routes.
> 
> Scenario 1: server is down
> 
> 1. user enters pin and unlocks the wallet;
> 2. call to load wallet will fail and error modal is shown;
> 3. user can change server;
> 
> Scenario 2: server is up but versions don't match
> 
> 1. user enters the pin and unlocks the wallet;
> 2. websocket will connect and trigger a version check;
> 3. versions won't match and we'll be displayed the Version screen;
> 
> Does it make sense?
> 
> PS - is the user presented the Version error screen before entering the pin in case we are connected to a server with an old version?

I totally agree with @obiyankenobi, and I feel it is worth going this way.

pedroferreira1:
I've done some changes and now the only change is that we don't check the version before unlocking the screen.

src/screens/Server.js
pedroferreira1: This code does not exist anymore. Can you see again, please?

PR #151, opened 2020-11-17
fix: sidebar hiding some elements

thr0wn:
<table>
  <tr>
    <td>Issue</td>
     <td>Fix</td>
  </tr>
  <tr>
    <td></td>
    <td></td>
  </tr>
 </table>

No side effects found on other screens:
<table>
  <tr>
    <td></td>
<td></td>
<td></td>
<td></td>
  </tr>
 </table>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
       "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.8.3.tgz",
       "integrity": "sha512-a9gxpmdXtZEInkCSHUJDLHZVBgb1QS0jhss4cPP93EW7s+uC5bikET2twEF3KV+7rDblJcmNvTR7VJejqd2C2g==",
       "requires": {
-        "@babel/highlight": "7.8.3"
+        "@babel/highlight": "^7.8.3"
       }
     },
     "@babel/compat-data": {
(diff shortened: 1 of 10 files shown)


package-lock.json
```diff
@@ -15,47 +15,47 @@
       "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.8.3.tgz",
       "integrity": "sha512-a9gxpmdXtZEInkCSHUJDLHZVBgb1QS0jhss4cPP93EW7s+uC5bikET2twEF3KV+7rDblJcmNvTR7VJejqd2C2g==",
       "requires": {
-        "@babel/highlight": "7.8.3"
+        "@babel/highlight": "^7.8.3"
```
pedroferreira1, line 18: I don't think we should change anything on the `package-lock.json`. There are many version changes that could impact the wallet.

pedroferreira1 (changes requested):

pedroferreira1:
Check this here. I think it was because of the nav position change from `fixed` to `sticky`. Is there a reason to change?

bug-nav

package-lock.json
thr0wn: You are right. All changes on `package-lock.json` were reverted, thanks.

thr0wn:
The purpose of position `sticky` was to consider the nav `height` while rendering the page, but I reverted as it caused some troubles with electron. Seems to be working fine now.

pedroferreira1 (approved):

obiyankenobi (approved):

PR #153, opened 2020-11-22
chore: use wait-on concurrently with electron

thr0wn:
* `npm run electron-dev` now runs `npm start` concurrently with `wait-on && electron`. It will automatically open electron after 'http://localhost:3000' is available

* `npm run electron-deps` now runs on `postinstall` hook


obiyankenobi (approved):

pedroferreira1 (approved):

msbrogli:
Why are we updating the `package-lock.json`?

msbrogli:
> Why are we updating the `package-lock.json`?

@thr0wn

thr0wn:
> Why are we updating the `package-lock.json`? @thr0wn

My fault, `package-lock.json` is no longer changed here.

PR #156, opened 2020-11-30
feat: New backup validation

thr0wn:
- Validates the backup in five steps using words randomly chosen between the 24 seed words to assure that the backup was correctly done.

- In each step, the user must find the correct word between some wrong words, similar to the mobile wallet backup.

### DIFF
--- a/src/components/ModalBackupWords.js
+++ b/src/components/ModalBackupWords.js
@@ -156,11 +156,13 @@ class ModalBackupWords extends React.Component {
   /**
    * Called when user finishes the backup. We validate the backup and shows a success message, in case of success
    */
-  handleValidationStepOption = (option, validationStep) => {
+  handleValidationStepOption = (option, validationStepIndex) => {
+    const validationSteps = [ ...this.state.validationSteps ];
+    const validationStep = validationSteps[validationStepIndex];
     validationStep.chosenOption = option;
     if (option !== validationStep.correctWord) {
       this.setState({
-        validationSteps: this.state.validationSteps,
+        validationSteps,
         errorMessage: (
           <span>
             <span>{t`Wrong word. Please double check the words you saved.`}</span>
@@ -175,7 +177,7 @@ class ModalBackupWords extends React.Component {
       return;
     }
     validationStep.done = true;
-    this.setState({ validationSteps: this.state.validationSteps });
+    this.setState({ validationSteps });
   };
 
   render() {
@@ -265,8 +267,8 @@ class ModalBackupWords extends React.Component {
             role='list'
             aria-label='Backup Validation Options'
           >
-            {validationStep.options.map((option) => (
-              <div key={option} role='listitem'>
+            {validationStep.options.map((option, index) => (
+              <div key={index} role='listitem'>
                 <button
                   className={`btn btn-dark btn-block validate-step-option ${
                     validationStep.chosenOption === option
@@ -274,7 +276,7 @@ class ModalBackupWords extends React.Component {
                       : ''
                   }`}
                   onClick={() =>
-                    this.handleValidationStepOption(option, validationStep)
+                    this.handleValidationStepOption(option, this.state.validationSteps.indexOf(validationStep))
                   }
                   disabled={this.state.errorMessage}
                 >
--- a/src/components/ModalBackupWords.js
+++ b/src/components/ModalBackupWords.js
@@ -11,7 +11,7 @@ import { t } from 'ttag';
 import $ from 'jquery';
 import _ from 'lodash';
 import { updateWords } from '../actions/index';
-import { connect } from "react-redux";
+import { connect } from 'react-redux';
 import hathorLib from '@hathor/wallet-lib';
 import { WORDS_VALIDATION } from '../constants';
 
@@ -51,33 +51,37 @@ class ModalBackupWords extends React.Component {
       passwordSuccess: false,
       showValidation: false,
       passwordFormValidated: false,
-      wordsValidation: [],
-      chosenWords: [],
-    }
+      validationSteps: [],
+    };
   }
 
   componentDidMount = () => {
     $('#backupWordsModal').on('hide.bs.modal', (e) => {
-      this.setState({ errorMessage: '', passwordSuccess: false, showValidation: false, wordsValidation: [], chosenWords: [] });
+      this.setState({
+        errorMessage: '',
+        passwordSuccess: false,
+        showValidation: false,
+        validationSteps: [],
+      });
       if (this.props.needPassword) {
         this.props.updateWords(null);
       }
       if (this.refs.password) {
         this.refs.password.value = '';
       }
-    })
+    });
 
     $('#backupWordsModal').on('shown.bs.modal', (e) => {
       if (this.refs.password) {
         this.refs.password.focus();
       }
-    })
-  }
+    });
+  };
 
   componentWillUnmount = () => {
     // Removing all event listeners
     $('#backupWordsModal').off();
-  }
+  };
 
   /**
    * Called when the user writes the password, so we can decrypt the saved words and show to him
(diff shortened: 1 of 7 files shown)


thr0wn:
### Selection the wrong word
wrong-way

### Passing all steps
good-way

src/components/ModalBackupWords.js
```diff
@@ -228,87 +257,126 @@ class ModalBackupWords extends React.Component {
...
-          <div className="chosen-words-wrapper w-100 d-flex flex-row flex-wrap align-items-start">
-            {renderChosenWords()}
-          </div>
-          <div className="words-validation d-flex flex-row align-items-start flex-wrap">
-            {renderWordsValidation()}
-          </div>
+        <div className='validation-progress d-flex flex-row justify-content-center'>
+          {this.state.validationSteps.map((validationStep) => (
+            <div
+              key={validationStep.correctWord}
```
obiyankenobi, line 298: This can be a problem if we have repeated words and both are chosen in the validation. Better to use the list index, no?

src/components/ModalBackupWords.js
```diff
@@ -228,87 +257,126 @@ class ModalBackupWords extends React.Component {
...
+            {!this.state.showValidation && (
+              <div className='modal-footer'>{renderButtons()}</div>
+            )}
           </div>
         </div>
       </div>
-    )
+    );
   }
 }
```
obiyankenobi, line 382: I'll add the comment here so we can discuss it in a thread.

Currently, there's one modal that displays the words and another that handles the backup. And when we close the second one the first is displayed. I'd prefer if there's only one modal for everything.

So the words are displayed in a modal and the backup process happens on the same one. When someone makes a mistake and starts over, he could be sent to the words list again, before restarting the backup process.

What do you think? @thr0wn @pedroferreira1

obiyankenobi (changes requested):

src/components/ModalBackupWords.js
```diff
@@ -228,87 +257,126 @@ class ModalBackupWords extends React.Component {
...
+            aria-label='Backup Validation Options'
+          >
+            {validationStep.options.map((option) => (
+              <div key={option} role='listitem'>
+                <button
+                  className={`btn btn-dark btn-block validate-step-option ${
+                    validationStep.chosenOption === option
+                      ? 'chosen-option'
+                      : ''
+                  }`}
```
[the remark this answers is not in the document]
thr0wn: @obiyankenobi, this ternary operator is inside `renderValidationStep`, which is called on every render. `handleValidationStepOption` does trigger a state change [here](https://github.com/HathorNetwork/hathor-wallet/blob/b1af7eedd12c29c3aaf8680c05f58ace8d83c512/src%2Fcomponents%2FModalBackupWords.js#L173).

The purpose of this ternary is highlight the chosen option (using the class `.chosen-option`), in order to change its background to red, like below:
image

src/components/ModalBackupWords.js
thr0wn: You are right. I will use the list index, although it is generally not [recommended](https://reactjs.org/docs/lists-and-keys.html), it is better than `validationStep.correctWord`, thanks!

src/components/ModalBackupWords.js
thr0wn: @obiyankenobi, technically the words and the backup/validation are displayed in the same modal, only its content is changing (`this.state.showValidation` controls which one is displayed). It may seem like two models due to the width change on the backup/validation:

<table>
  <thead>
    <tr>
      <th>Words</td>
      <th>Validation</td>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>
image
      </td>
      <td>
        image
      </td>
    </tr>
  </tbody>
</table>

With no changes on the modal width and no close button on the backup validation, it should behave as you suggested.
What do you think?

obiyankenobi: What gave that impression is that, on the first modal (Selecting the wrong word), when you press close, it displays the words again.

And probably keeping the same width is better. You can even increase the width size on the words step if it's better.

src/components/ModalBackupWords.js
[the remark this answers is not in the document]
obiyankenobi: Got it, thanks.

src/components/ModalBackupWords.js
```diff
@@ -96,58 +100,85 @@ class ModalBackupWords extends React.Component {
...
-      this.setState({ errorMessage: t`Invalid number of words` });
+  handleValidationStepOption = (option, validationStep) => {
+    validationStep.chosenOption = option;
+    if (option !== validationStep.correctWord) {
+      this.setState({
+        validationStep,
+        errorMessage: (
+          <span>
+            {t`Wrong word. Please double check the words you saved.`}
+            <strong onClick={() => this.setState({ showValidation: false, errorMessage: '' })}>Click here to start the process again.</strong>
```
pedroferreira1, line 162: This should have the translation tag as well

src/components/ModalBackupWords.js
```diff
@@ -51,33 +51,37 @@ class ModalBackupWords extends React.Component {
       passwordSuccess: false,
       showValidation: false,
       passwordFormValidated: false,
-      wordsValidation: [],
-      chosenWords: [],
-    }
+      validationSteps: [],
```
pedroferreira1, line 54: I would be good to have a description of this state in the comment before `this.state = `. As far as I understand it's an array of objects, so it should have each key of the object the its meaning.

Besides that, you should remove `wordsValidation` and  `chosenWords` from the comment because they were removed

src/components/ModalBackupWords.js
```diff
@@ -96,58 +100,85 @@ class ModalBackupWords extends React.Component {
...
-      validationArray = validationArray.slice(wordIdx);
+    if (validationStep.last) {
+      this.props.validationSuccess();
+      return;
     }
-    this.setState({ errorMessage: '' });
-    this.props.validationSuccess();
-  }
+    validationStep.done = true;
+    this.setState({ validationStep });
```
pedroferreira1, line 173: Is this a new state? I don't see it in the definition of `this.state`

src/components/ModalBackupWords.js
pedroferreira1: I agree that keeping the same width is better. I don't really like the width of the "new modal", I would decrease it a bit and the other can increase to match.

I will give my impressions on the UI here in this thread.

1. I think the `handleClose` method shouldn't move back to the words modal, doing that we need to click 'close' twice to close the modal. So `handleClose` should change the state `showValidation` back to false and close the modal (you should change the state only after modal hidden event is triggered)
2. I think the validation progress should show the step you are in (we 'paint' the circle purple only after you choose the word, I think the progress should say 'you are in step 1', so 1 purple circle painted). - I am not sure about this suggestion, I would like to hear your thoughts by my instinct says it's better this way.
3. I know it's not from this PR but I believe the success message is being shown really fast. To update it you can go to `src/screens/NewWallet.js` line 123 and update from 1000 to 3000 maybe.

What do you think? @obiyankenobi @thr0wn

pedroferreira1 (changes requested):

src/components/ModalBackupWords.js
thr0wn: Changing `this.state` directly is not a good practice, as it should be treated as immutable (more details here: https://reactjs.org/docs/react-component.html#state). We should use `this.setState` whenever is possible.

obiyankenobi: I think he means explaining in the comments what `validationStep` means.

src/components/ModalBackupWords.js
thr0wn: Compiling all suggestions, we have the following:

1. Success
back-success

2. Error
back-error

3. Close
backup-close

Sounds good?

src/components/ModalBackupWords.js
thr0wn: Sorry. I added a comment about each `validationStep` property that will help to understand this object.

src/components/ModalBackupWords.js
obiyankenobi: Looks good.

src/components/ModalBackupWords.js
```diff
@@ -96,58 +105,88 @@ class ModalBackupWords extends React.Component {
...
    */
-  chosenWordRemoved = (word) => {
-    let chosenWords = this.state.chosenWords;
-    const idx = chosenWords.indexOf(word);
-    chosenWords.splice(idx, 1);
-    this.setState({ chosenWords });
-  }
+  handleClose = () => {
+    if (this.state.showValidation) {
+      this.setState({ showValidation: false, errorMessage: '' });
```
pedroferreira1, line 186: I think you don't need this here because on `componentDidMount` you already set this state on ` $('#backupWordsModal').on('hide.bs.modal', (e) => {` event handler

src/components/ModalBackupWords.js
pedroferreira1: Yeah, I like it

src/components/ModalBackupWords.js
pedroferreira1: Not only that, I don't see where `this.state.validationStep` is used, we have two different states `validationStep` and `validationSteps` and the first one is never used.

thr0wn: Absolutely right. Put `validationStep` into the state did not make sense. Fixed.

src/components/ModalBackupWords.js
thr0wn: Done

src/components/ModalBackupWords.js
thr0wn: Done

obiyankenobi (approved):

src/components/ModalBackupWords.js
thr0wn: Done

thr0wn: Done

src/components/ModalBackupWords.js
thr0wn: `handleClose` removed, thanks!

src/components/ModalBackupWords.js
```diff
@@ -96,59 +105,78 @@ class ModalBackupWords extends React.Component {
...
    * Called when user finishes the backup. We validate the backup and shows a success message, in case of success
    */
-  handleValidateBackup = () => {
-    if (this.state.chosenWords.length !== WORDS_VALIDATION) {
-      this.setState({ errorMessage: t`Invalid number of words` });
+  handleValidationStepOption = (option, validationStep) => {
+    validationStep.chosenOption = option;
+    if (option !== validationStep.correctWord) {
+      this.setState({
+        validationSteps: this.state.validationSteps,
```
pedroferreira1, line 163: What's this `setState` is doing? It's just saving the same data, isn't it?

src/components/ModalBackupWords.js
```diff
@@ -96,59 +105,78 @@ class ModalBackupWords extends React.Component {
...
-  /**
-   * Called when user clicks in a word on the backup validation (just add the word in the array)
-   */
-  validationWordClicked = (word) => {
-    let chosenWords = this.state.chosenWords;
-    chosenWords.push(word);
-    this.setState({ chosenWords });
-  }
+    validationStep.done = true;
+    this.setState({ validationSteps: this.state.validationSteps });
```
pedroferreira1, line 178: The same comment of the last `setState`, it's just setting the same data, isn't it?

src/components/ModalBackupWords.js
thr0wn: At line 160 the statement `validationStep.chosenOption = option;` changed the property `chosenOption` of `validationStep`, which is an element of `validationSteps` array. So, `validationSteps` isn't "deep equal" to its previous value before line 160. As this change will only be visible with a re-render, I included `validationSteps` inside the `setState` call to indicate that it has changed, although it is not really necessary and even a `this.forceUpdate()` could do the re-render job. I can do something like the below to be more clear about the intention to change `validationSteps`.
```js
this.setState({
    validationSteps: [ ...this.state.validationSteps ],
    ...
});
``` 
The point here is the need for a re-render to update any markup affected by `validationSteps`. I will do `validationSteps: [ ...this.state.validationSteps ]` clearly indicate that `validationSteps` has changed, ok?

src/components/ModalBackupWords.js
thr0wn: Same of https://github.com/HathorNetwork/hathor-wallet/pull/156#discussion_r537501578.

thr0wn: Resolved by https://github.com/HathorNetwork/hathor-wallet/pull/156/commits/e3958cb3d1836dbfe98dd12e9cb4c1e60cdfe566.

src/components/ModalBackupWords.js
thr0wn: Resolved by https://github.com/HathorNetwork/hathor-wallet/pull/156/commits/e3958cb3d1836dbfe98dd12e9cb4c1e60cdfe566.

pedroferreira1 (approved):

PR #160, opened 2020-12-18
feat: updating lib network when changing for a testnet/mainnet server

pedroferreira1:
Still need to bump the lib version, so the tx mining service works for the testnet

### DIFF
--- a/public/electron.js
+++ b/public/electron.js
@@ -34,7 +34,7 @@ if (process.platform === 'darwin') {
 }
 
 const appName = 'Hathor Wallet';
-const walletVersion = '0.18.0';
+const walletVersion = '0.18.1';
 
 // Keep a global reference of the window object, if you don't, the window will
 // be closed automatically when the JavaScript object is garbage collected.
--- a/public/electron.js
+++ b/public/electron.js
@@ -34,7 +34,7 @@ if (process.platform === 'darwin') {
 }
 
 const appName = 'Hathor Wallet';
-const walletVersion = '0.18.1';
+const walletVersion = '0.18.0';
 
 // Keep a global reference of the window object, if you don't, the window will
 // be closed automatically when the JavaScript object is garbage collected.
--- a/src/App.js
+++ b/src/App.js
@@ -49,6 +49,11 @@ hathorLib.storage.setStore(new HybridStore());
 // Default if data on storage is null is mainnet
 // because we currently support only mainnet
 const currentNetwork = hathorLib.storage.getItem('wallet:network') || 'mainnet';
+// We always set the network in the version.js utils file when we receive the /version/ from the full node
+// however when we first start the app we also need to set it because this /version/ request is async
+// and we might start loading the transactions before we receive it.
+// It would be better if we load the transactions only after the /version/ response
+// but while we don't refactor this, we have to save the network in the storage to set here
 hathorLib.network.setNetwork(currentNetwork);
 
 // set default server
--- a/src/App.js
+++ b/src/App.js
@@ -45,14 +45,11 @@ import SoftwareWalletWarningMessage from './components/SoftwareWalletWarningMess
 import AddressList from './screens/AddressList';
 
 hathorLib.storage.setStore(new HybridStore());
-const currentNetwork = hathorLib.storage.getItem('wallet:network');
-if (currentNetwork === 'testnet') {
-  hathorLib.network.setNetwork('testnet');
-} else {
-  // Default if data on storage is null is mainnet
-  // because we currently support only mainnet
-  hathorLib.network.setNetwork('mainnet');
-}
+
+// Default if data on storage is null is mainnet
+// because we currently support only mainnet
+const currentNetwork = hathorLib.storage.getItem('wallet:network') || 'mainnet';
+hathorLib.network.setNetwork(currentNetwork);
 
 // set default server
 hathorLib.wallet.setDefaultServer(DEFAULT_SERVER);
--- a/src/App.js
+++ b/src/App.js
@@ -44,8 +44,15 @@ import ModalAlert from './components/ModalAlert';
 import SoftwareWalletWarningMessage from './components/SoftwareWalletWarningMessage';
 import AddressList from './screens/AddressList';
 
-hathorLib.network.setNetwork('mainnet');
 hathorLib.storage.setStore(new HybridStore());
+const currentNetwork = hathorLib.storage.getItem('wallet:network');
+if (currentNetwork === 'testnet') {
+  hathorLib.network.setNetwork('testnet');
+} else {
+  // Default if data on storage is null is mainnet
+  // because we currently support only mainnet
+  hathorLib.network.setNetwork('mainnet');
+}
 
 // set default server
 hathorLib.wallet.setDefaultServer(DEFAULT_SERVER);
--- a/src/App.js
+++ b/src/App.js
@@ -45,17 +45,6 @@ import SoftwareWalletWarningMessage from './components/SoftwareWalletWarningMess
 import AddressList from './screens/AddressList';
 
 hathorLib.storage.setStore(new HybridStore());
-
-// Default if data on storage is null is mainnet
-// because we currently support only mainnet
-const currentNetwork = hathorLib.storage.getItem('wallet:network') || 'mainnet';
-// We always set the network in the version.js utils file when we receive the /version/ from the full node
-// however when we first start the app we also need to set it because this /version/ request is async
-// and we might start loading the transactions before we receive it.
-// It would be better if we load the transactions only after the /version/ response
-// but while we don't refactor this, we have to save the network in the storage to set here
-hathorLib.network.setNetwork(currentNetwork);
-
 // set default server
 hathorLib.wallet.setDefaultServer(DEFAULT_SERVER);
 
@@ -240,7 +229,10 @@ const returnLoadedWalletComponent = (Component, props, rest) => {
     promise.then(() => {
       wallet.localStorageToRedux();
     });
-    return <WaitVersion {...props} />;
+    return <Redirect to={{
+      pathname: '/loading_addresses/',
+      state: {path: props.match.url}
+    }} />;
   } else if (reduxState.isVersionAllowed === false && !isServerScreen) {
     return <VersionError {...props} />;
   } else {
--- a/src/screens/Server.js
+++ b/src/screens/Server.js
@@ -7,6 +7,7 @@
 
 import React from 'react';
 import { t } from 'ttag';
+import SpanFmt from '../components/SpanFmt';
 import $ from 'jquery';
 import version from '../utils/version';
 import wallet from '../utils/wallet';
@@ -169,8 +170,8 @@ class Server extends React.Component {
     const renderConfirmBody = () => {
       return (
         <div>
-          <p>The selected server connects you to a testnet. Beware if someone asked you to do it, the <strong>tokens from testnet have no value</strong>. Only continue if you know what you are doing.</p>
-          <p>To continue with the server change you must type 'testnet' in the box below and click on 'Connect to testnet' button.</p>
+          <p><SpanFmt>{t`The selected server connects you to a testnet. Beware if someone asked you to do it, the **tokens from testnet have no value**. Only continue if you know what you are doing.`}</SpanFmt></p>
+          <p>{t`To continue with the server change you must type 'testnet' in the box below and click on 'Connect to testnet' button.`}</p>
           <div ref="testnetWrapper" className="mt-2 d-flex flex-row align-items-center">
             <input type="text" ref="testnetTest" className="form-control col-4" />
             <span className="text-danger ml-2">{this.state.testnetError}</span>
@@ -181,7 +182,7 @@ class Server extends React.Component {
 
     const renderSecondaryModalButton = () => {
       return (
-        <button onClick={this.confirmTestnetServer} type="button" className="btn btn-secondary">Connect to testnet</button>
+        <button onClick={this.confirmTestnetServer} type="button" className="btn btn-secondary">{t`Connect to testnet`}</button>
       );
     }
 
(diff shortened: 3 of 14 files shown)


src/App.js
```diff
@@ -44,9 +44,13 @@ import ModalAlert from './components/ModalAlert';
 import SoftwareWalletWarningMessage from './components/SoftwareWalletWarningMessage';
 import AddressList from './screens/AddressList';
 
-hathorLib.network.setNetwork('mainnet');
 hathorLib.storage.setStore(new HybridStore());
 
+// Default if data on storage is null is mainnet
+// because we currently support only mainnet
+const currentNetwork = hathorLib.storage.getItem('wallet:network') || 'mainnet';
+hathorLib.network.setNetwork(currentNetwork);
```
obiyankenobi, line 52: Why is this needed? I thought we would always perform `hathorLib.network.setNetwork` when we receive the response from the version API call.

pedroferreira1: Yeah, I am doing it in the `src/utils/version.js` file, however this is needed for the moment I open the app and because of the way we are doing to validate the version API.

We execute async the version API call and the wallet load, so when we start loading we still don't have the version API call response. This was done because otherwise we would be 'waiting' the version API call in a blank screen. I need to do a bigger refactor to fix this and didn't want to do it in this PR

obiyankenobi: Ok, got it. Agree on holding the refactor for later. I'd just say to put this as a comment so other people don't have this same doubt.

obiyankenobi (approved):

thr0wn (approved):

src/App.js
pedroferreira1: Added the comment

pedroferreira1:
As @msbrogli suggested, I've added a confirmation when the user wants to connect to a testnet, to prevent a wrong connection. What do you think @msbrogli?


Uploading Testnet checkbox.mov…

public/electron.js
```diff
@@ -34,7 +34,7 @@ if (process.platform === 'darwin') {
 }
 
 const appName = 'Hathor Wallet';
-const walletVersion = '0.18.0';
+const walletVersion = '0.18.1';
```
msbrogli, line 37: I feel we shouldn't change the version here.

src/App.js
```diff
@@ -44,9 +44,18 @@ import ModalAlert from './components/ModalAlert';
...
 
-hathorLib.network.setNetwork('mainnet');
 hathorLib.storage.setStore(new HybridStore());
 
+// Default if data on storage is null is mainnet
+// because we currently support only mainnet
+const currentNetwork = hathorLib.storage.getItem('wallet:network') || 'mainnet';
+// We always set the network in the version.js utils file when we receive the /version/ from the full node
+// however when we first start the app we also need to set it because this /version/ request is async
+// and we might start loading the transactions before we receive it.
```
msbrogli, line 54: I think we should refactor it before merging the PR. Why wasn't it a problem in the current version of the wallet?

src/screens/Server.js
```diff
@@ -30,12 +31,14 @@ class Server extends React.Component {
      * loading {boolean} If should show spinner while waiting for server response
      * newServer {boolean} If user selected checkbox that he wants to set a new server
      * selectedValue {string} Server selected from the user
+     * testnetServer {boolean} If user selected checkbox that he wants to connect to the testnet
```
msbrogli, line 34: I'd call it mainnet. I feel that the most important thing is whether we are connected to the mainnet or not.

src/screens/Server.js
```diff
@@ -69,10 +72,20 @@ class Server extends React.Component {
...
+    const currentNetwork = hathorLib.network.getNetwork().name;
+    const currentServer = hathorLib.helpers.getServerURL();
     // Update new server in local storage
     hathorLib.wallet.changeServer(newServer);
     const promise = version.checkApiVersion();
-    promise.then(() => {
+    promise.then((data) => {
+      // If the user has not explicitly said that he wants to connect to a testnet but end up connecting to one, we show error and don't continue
+      if (!this.state.testnetServer && data.network !== 'mainnet') {
+        this.setState({ errorMessage: t`You tried to connect to a testnet server and did not select that. Beware if someone asked you to do it, the tokens from testnet have no value.`, loading: false });
```
msbrogli, line 83: I'd say: "You expect to connect to Hathor's mainnet but the server is connected to a different network. Beware that tokens from other networks have no value."

src/utils/version.js
```diff
@@ -31,9 +32,16 @@ const version = {
...
         store.dispatch(isVersionAllowedUpdate({allowed: hathorLib.helpers.isVersionAllowed(data.version, hathorLib.constants.MIN_API_VERSION)}));
-        // Update network in redux
-        store.dispatch(networkUpdate({network: data.network}));
-        resolve();
+        // Set network in lib to use the correct address byte
+        let network;
+        if (data.network === 'mainnet') {
+          network = 'mainnet';
+        } else {
+          // Can we assume it will be testnet? I think it's safe
```
msbrogli, line 40: I think so. We can define that all other networks are testnet except the mainnet.

msbrogli (changes requested):

public/electron.js
pedroferreira1: You mean shouldn't change in this file or shouldn't change in this PR?

src/utils/version.js
pedroferreira1: Agreed

src/screens/Server.js
pedroferreira1: Do you think the checkbox should be "I want to connect to the mainnet"? If that's what you mean by your comment I don't really agree.

The mainnet is the default network, thus if someone wants to deceive a user into connecting in the testnet to pretend that they received real tokens, I think the checkbox and the messages should be about the testnet, like I added "I want to connect to the testnet and understand that the tokens have no value there".

If we require the checkbox only for the mainnet someone might add a testnet server and won't need to read/select the checkbox.

What do you think?

src/screens/Server.js
obiyankenobi, line 83: Isn't it better to open a modal when the user switches to testnet asking him to confirm? I feel that having the checkbox there always looks weird and a modal explicitly states what the action is. We can even make the default action on the modal be to cancel it.

obiyankenobi:
We also need to change the explorer link.

src/screens/Server.js
pedroferreira1: To be honest I feel that the modal or the current checkbox is pretty much the same, I even prefer the checkbox but I see no benefits in changing to add the modal.

obiyankenobi: The modal would only open if the user is switching to a testnet. It looks weird to me having the checkbox if the user hasn't typed a testnet node api.

pedroferreira1:
Now the explorer URL changes depending on the full node network. I still think we should be able to customize it, just like @msbrogli suggested we should do with the tx mining service URL but I thought about keeping like this in this PR and create another one for the customizations.

For this I created another alias on AWS (explorer.testnet.hathor.network) so we won't need to update this if we decide to release a new testnet in the future, we always can access the latest testnet explorer using this URL. What do you think? @msbrogli @obiyankenobi

src/screens/Server.js
pedroferreira1: I think this screen will be used either by experienced users who knows what they are doing, so the checkbox would be ok and by users that are being deceived and in this case I prefer having this protection before changing the server, rather than creating a modal that can't be closed and has some requirements to protect this user who has already done something wrong. But again, it's pretty much the same for me, I don't see much difference, I can change to a modal.

obiyankenobi: I think a modal where the default action is cancelling the change seems safer. Or even requiring on the modal to select a checkbox to confirm. We should get more opinions @msbrogli @jansegre @andreabadesso @thr0wn

src/App.js
pedroferreira1: It was not a problem because we weren't changing the network. I've just done this refactor and now this code is not needed anymore. Can you check again please?

public/electron.js
msbrogli: I think we shouldn't change in this PR. It might be part of other changes. I think it is better to create a separate PR to increase the version, and we can take the opportunity to put the changelog in that PR.

src/screens/Server.js
msbrogli: I meant just the flag. The checkbox should keep the testnet message. If the flag is `mainnetServer` and the checkbox is market, the flag is `false`.

src/screens/Server.js
msbrogli: If we go with a modal, I think we should require the user to write "I want to connect to the testnet" or similar (instead of just clicking in a button). Anyway, I think connecting to the testnet is a major thing and should be "hard to do". We can require the user to enable "Dev mode" to do that, for example. I know it feels late to discuss UI here but I feel that the easiest it is, the more people will be scammed.

src/App.js
msbrogli: Great! Thanks!

msbrogli (approved):

public/electron.js
pedroferreira1: Ok, makes sense. Will rollback those changes

pedroferreira1:
https://user-images.githubusercontent.com/3298774/105863259-78884e80-5fcf-11eb-95cc-31f424247d16.mov

The first solution was this one: https://github.com/HathorNetwork/hathor-wallet/pull/160#issuecomment-757064548

I believe @obiyankenobi is more like the video uploaded in this comment and that @msbrogli would like to have some hybrid solution, is that right?

One other thing I was thinking about is if someone tries to start a single node (using mainnet as network) but isolated from our network. In that case it could ask the user to connects to the node, it would show 'mainnet' but it wouldn't be connected to our mainnet. This is a real problem, right?

@msbrogli @obiyankenobi what are your thoughts? How should we proceed with this UI/UX?

obiyankenobi:
@pedroferreira1 That's exactly what I was looking for. I might just suggest changing the default modal button to Cancel. Or have "Connect to testnet" and "Cancel change", where the latter is the default/highlighted button.

> One other thing I was thinking about is if someone tries to start a single node (using mainnet as network) but isolated from our network. In that case it could ask the user to connects to the node, it would show 'mainnet' but it wouldn't be connected to our mainnet. This is a real problem, right?

Seems like a problem in every crypto. I can always spin up an isolated BTC node. I'm not sure there's much we can do there. Users should only connect to nodes they trust.

src/screens/Server.js
andreabadesso: I would agree with the modal if changing the server was "routine usage", but changing the server is already an "advanced" usage of the wallet, there are multiple steps before being able to do it.

So I think the checkbox should be enough.

src/screens/Server.js
```diff
@@ -69,10 +72,20 @@ class Server extends React.Component {
     }
 
     this.setState({ loading: true, errorMessage: '' });
+    const currentNetwork = hathorLib.network.getNetwork().name;
+    const currentServer = hathorLib.helpers.getServerURL();
     // Update new server in local storage
     hathorLib.wallet.changeServer(newServer);
```
andreabadesso, line 78: Wouldn't it be safer to change server only after the `checkApiVersion` call is resolved?

I justed tested and you are able to bypass the checkbox entirelly if you change the server when not connected to the internet.

Here are the steps to reproduce:

1. Change server to https://node1.foxtrot.testnet.hathor.network/v1a/ while disconnected from the internet
2. Do not click the "I want to connect to a full node. I understand..." checkbox
3. It will give a connection error, but will let you close the modal
4. Connect to the internet
5. You are now connected to the testnet

I don't know if a scammer would be able to convince someone to turn off the internet before changing the server, but you never know

andreabadesso (changes requested):

src/screens/Server.js
andreabadesso: I would agree with the modal if changing the server was "routine usage", but changing the server is already an "advanced" usage of the wallet, there are multiple steps before being able to do it.

So I think the checkbox should be enough.

src/screens/Server.js
pedroferreira1: I agree it would be better and safer to change the server only after the `checkApiVersion`. However this would require a bigger refactor on the lib because right now all API calls are made to the server in the `localStorage`. That's why I change the server before, check the network and, in case it's a testnet, I change back to the old server.

I think our APIs could receive an optional parameter, which would be the URL to call the API but we currently don't have it.

My suggestion is to create an issue to add this optional parameter and in the meantime this problem can be handled just going back to the old server in case of promise rejection. What do you think?

andreabadesso: Sounds good, I agree

src/screens/Server.js
msbrogli: I just saw the testnet-connection.mov video and I think that's the best one. I agree with @obiyankenobi that we should rename the buttons to "Connect to testnet" and "Cancel change" while cancelling is the default action.

We can go further and ask users to write "testnet" in a textbox to force them to read before clicking in the button. I feel it is a serious business to allow users to connect to the testnet from our wallet because it opens a new world of opportunities for scammers.

pedroferreira1:
https://user-images.githubusercontent.com/3298774/106167001-85da4000-616b-11eb-898a-894c002e8f4c.mov

@msbrogli @obiyankenobi @andreabadesso this is the final version of the UI/UX with the latest comment that @msbrogli has sent.

What do you think? Can we move on with this?

src/screens/Server.js
pedroferreira1: I've just added the latest video with the updated UI/UX. We can track the comments in this thread.

@andreabadesso @obiyankenobi @msbrogli

src/screens/Server.js
pedroferreira1: Done

src/screens/Server.js
pedroferreira1: This code is outdated

src/screens/Server.js
pedroferreira1: This code is outdated

src/screens/Server.js
obiyankenobi: This flow seems really good to me.

andreabadesso: Looks good to me

msbrogli: Great. \o/

andreabadesso (approved):

PR #165, opened 2021-02-02
fix: when starting/loading wallet should check API version

pedroferreira1:


andreabadesso (approved):
LGTM

src/components/ChoosePin.js
```diff
@@ -36,13 +37,23 @@ class ChoosePin extends React.Component {
     this.props.updatePin(value);
   }
 
+  /**
+   * Check API version and update server network before loading data
+   */
+  handleSuccess = () => {
```
msbrogli, line 43: Isn't it better to add it to the reconnect event? I feel it is more related to each action instead of the act of setting a PIN.

msbrogli (approved):

msbrogli:
> LGTM

image

src/components/ChoosePin.js
pedroferreira1: Yeah, it's better. Changed it, thanks

PR #169, opened 2021-02-05
fix: we should clean state and variables when we receive a new tx on txdata component

pedroferreira1:


src/components/TxData.js
```diff
@@ -61,6 +63,14 @@ class TxData extends React.Component {
 
   componentDidUpdate = (prevProps) => {
     if (prevProps.transaction !== this.props.transaction) {
+      // When we are changing the transaction but keeping the same component
```
msbrogli, line 66: I agree we must clean the state. But I feel that we should rever reuse a component. We should always create a new one (and it must be done by whoever is using it). A simple way to do it is to set a `<TxData ... key={txId} />` because it will prevent react from matching the same component with another transaction.

src/components/TxData.js
```diff
@@ -61,6 +63,14 @@ class TxData extends React.Component {
...
+      // we must clean the state
+      this.tokensFound = [];
+      if (this.state.raw) {
+        // XXX Show raw transaction should come directly from state, we should refactor this later
+        this.toggleRaw();
```
msbrogli, line 71: Should we call it after reseting the state? I feel we might have a minor bug here.

msbrogli (changes requested):

src/components/TxData.js
pedroferreira1: I agree with you! Just updated the code

src/components/TxData.js
pedroferreira1: This code is outdated

src/components/TxData.js
msbrogli: Great!

msbrogli (approved):

andreabadesso (approved):

PR #180, opened 2021-04-09
fix: we shouldn't show ledger message in software wallet before official support

pedroferreira1:


src/components/Version.js
```diff
@@ -22,6 +22,10 @@ class Version extends React.Component {
    * If it's software wallet show modal warning
    */
   walletTypeClicked = () => {
+    // XXX We currently do not support ledger.
```
msbrogli, line 25: Why don't we create a constant `LEDGER_ENABLED` and use it along the wallet? We can also create a parameter `--enable-ledger-support-beta` so users can enable support when needed.

msbrogli (changes requested):

src/components/Version.js
pedroferreira1: I've done some changes to create the `LEDGER_ENABLED` constant. Can you see what you think?

I didn't create the `enable-ledger-support-beta` parameter as I think is not really helpful

andreabadesso (approved):

src/components/Version.js
msbrogli: I feel it would be useful for anyone using the Ledger in dev mode. Maybe we should request a `--enable-beta-features` as well.

msbrogli: I just created Issue https://github.com/HathorNetwork/hathor-wallet/issues/183 with this feature request.

msbrogli (approved):

PR #181, opened 2021-04-09
feat: show local date and time in settings screen

pedroferreira1:
The main purpose of this feature is to help debug some timestamp issues we've been having.

Question: Should we show in this format? I like showing the full date/time and timezone. I could remove the weekday only but I left because it's not worse with it in my opinion.

image

### DIFF
--- a/src/screens/Settings.js
+++ b/src/screens/Settings.js
@@ -126,9 +126,13 @@ class Settings extends React.Component {
 
   render() {
     const serverURL = hathorLib.helpers.getServerURL();
+    const now = new Date();
     return (
       <div className="content-wrapper settings">
         <BackButton {...this.props} />
+        <div>
+          <p><strong>{t`Date and time:`}</strong> {now.toString()}</p>
+        </div>
         <div>
           <p><SpanFmt>{t`**Server:** You are connected to ${serverURL}`}</SpanFmt></p>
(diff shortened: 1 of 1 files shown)


src/screens/Settings.js
```diff
@@ -126,9 +126,13 @@ class Settings extends React.Component {
 
   render() {
     const serverURL = hathorLib.helpers.getServerURL();
+    const now = new Date();
     return (
       <div className="content-wrapper settings">
         <BackButton {...this.props} />
+        <div>
+          <p><strong>{t`Date and time:`}</strong> {now.toString()}</p>
```
msbrogli, line 134: Should we put the `timestamp` in a "!" badge?

msbrogli (approved):

src/screens/Settings.js
pedroferreira1: Done. Added the badge with the timestamp as title

andreabadesso (approved):

andreabadesso (approved):
I understand that the idea of this PR is to make it easier to debug when a user has issues with his clock.

But a suggestion would be to receive and store somehow the full node's time and warn the user if the difference is higher than a threshold 

This would allow the user to self diagnose the problem, avoiding a call on our helpdesk entirelly

msbrogli (approved):

PR #188, opened 2021-04-23
feat: lib facade integration

pedroferreira1:


msbrogli (changes requested):

src/components/SendTokensOne.js
```diff
@@ -124,31 +124,14 @@ class SendTokensOne extends React.Component {
         const index = input.current.index.current.value;
 
         if (txId && index) {
-          data['inputs'].push({'tx_id': txId, 'index': index, 'token': this.state.selected.uid });
+          data['inputs'].push({'hash': txId, 'index': index, 'token': this.state.selected.uid });
```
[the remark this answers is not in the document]
pedroferreira1: Humm `hash` is more generic, right? I agree, changed back to `tx_id` but also changed in the lib, which was expecting `hash` (will need to change this in the headless as well)

src/components/TokenHistory.js
```diff
@@ -43,7 +50,9 @@ class TokenHistory extends React.Component {
   }
 
   componentDidUpdate = (prevProps) => {
-    if (this.props.history !== prevProps.history) {
```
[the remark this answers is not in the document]
pedroferreira1: This is a token history screen, not a tx screen. And this `if` is because re-render was not being called when a new tx arrived in the redux.

```
  return { 
    tokensHistory: state.tokensHistory,
  };
};
```

The redux array that must be checked depend on the component props, otherwise I would be able to do something like this

```
  return { 
    tokensHistory: state.tokensHistory[this.props.tokenUID],
  };
};
```

But I don't have access to `this.props`. Is there a way to do it?

src/components/tokens/TokenAction.js
```diff
@@ -112,11 +112,9 @@ class TokenAction extends React.Component {
    * Method executed after PIN is written
    * If success when preparing the tx, return the send tx object
    * Otherwise, show error message
-   *
-   * @param {String} pin PIN written by the user
    */
-  onPrepareSendTransaction = (pin) => {
```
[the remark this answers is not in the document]
pedroferreira1: `HathorWallet` object already has the PIN

src/components/WalletBalance.js
```diff
@@ -24,21 +28,25 @@ class WalletBalance extends React.Component {
   render = () => {
     const token = this.props.tokens.find((token) => token.uid === this.props.selectedToken);
     const symbol = token ? token.symbol : '';
+    const balance = {
```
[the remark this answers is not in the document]
pedroferreira1: Actually not. The way is implemented, the moment we unlock the screen we redirect the user to the home screen (which shows de balance). Then we show the loading screen address above it as a priority, until the wallet is loaded but while it's loading `this.props.tokenBalance` might be null.

src/components/tokens/TokenAction.js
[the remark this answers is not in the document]
msbrogli: Same thing here. In this wallet, I don't think it should store the pin.

src/components/ModalSendTx.js
```diff
@@ -89,7 +89,7 @@ class ModalSendTx extends React.Component {
       if (hathorLib.wallet.isPinCorrect(pin)) {
         $('#pinModal').data('bs.modal')._config.backdrop = 'static';
         $('#pinModal').data('bs.modal')._config.keyboard = false;
-        this.sendTransaction = this.props.prepareSendTransaction(pin);
+        this.sendTransaction = this.props.prepareSendTransaction();
```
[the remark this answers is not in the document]
msbrogli: As we discussed before, I agree with doing it in a separate PR.

src/components/WalletAddress.js
```diff
@@ -66,7 +82,6 @@ class WalletAddress extends React.Component {
       if (hathorLib.wallet.canGenerateNewAddress()) {
         wallet.generateNewAddress();
       } else {
-        this.refs.alertError.show(3000);
```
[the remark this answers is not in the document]
msbrogli: Ok.

src/components/ModalSendTx.js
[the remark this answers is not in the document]
pedroferreira1: Done and already integrated with this PR

src/components/tokens/TokenAction.js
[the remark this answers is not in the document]
pedroferreira1: Done and already integrated with this PR

src/components/ModalSendTx.js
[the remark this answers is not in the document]
msbrogli: Great. Thanks!

src/components/TokenHistory.js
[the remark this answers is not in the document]
msbrogli: I guess you have access to `props`. Check it out here: https://react-redux.js.org/using-react-redux/connect-mapstate

src/components/WalletBalance.js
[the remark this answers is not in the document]
msbrogli: Make sense. I'm wondering whether we should create this component without a balance but maybe it is something we should change in a separate PR. What do you think?

src/components/tokens/TokenAction.js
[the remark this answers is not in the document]
msbrogli: Great! Thanks!

src/App.js
```diff
@@ -227,9 +118,6 @@ const returnLoadedWalletComponent = (Component, props, rest) => {
   // Check version
   if (reduxState.isVersionAllowed === undefined && !isServerScreen) {
     const promise = version.checkApiVersion();
-    promise.then(() => {
-      wallet.localStorageToRedux();
```
msbrogli, line 231: No need to call it anymore?

src/components/SendTokensOne.js
```diff
@@ -208,8 +191,8 @@ class SendTokensOne extends React.Component {
     }
 
     const renderBalance = () => {
-      const balance = hathorLib.wallet.calculateBalance(Object.values(this.props.historyTransactions), this.state.selected.uid);
-      return <span className="ml-3">({t`Balance available: `}{hathorLib.helpers.prettyValue(balance.available)})</span>;
+      const availableBalance = this.state.selected.uid in this.props.tokensBalance ? this.props.tokensBalance[this.state.selected.uid].available : 0;
```
msbrogli, line 194: As we are in the code, I feel it is way better to use an if-else block instead of the conditional ternary operator. What do you think?

src/components/TokenAdministrative.js
```diff
@@ -17,13 +17,13 @@ import hathorLib from '@hathor/wallet-lib';
 import PropTypes from 'prop-types';
 
 const mapStateToProps = (state) => {
-  const balance = hathorLib.wallet.calculateBalance(
-    Object.values(state.historyTransactions),
-    hathorLib.constants.HATHOR_TOKEN_CONFIG.uid
-  );
+  const HTR_UID = hathorLib.constants.HATHOR_TOKEN_CONFIG.uid;
+  const htrBalance = HTR_UID in state.tokensBalance ? state.tokensBalance[HTR_UID].available : 0;
```
msbrogli, line 21: Same thing here about the conditional ternary operator.

src/components/TokenAdministrative.js
```diff
@@ -59,10 +57,8 @@ class TokenAdministrative extends React.Component {
   }
 
   componentDidUpdate = (prevProps) => {
-    if (this.props.historyTransactions !== prevProps.historyTransactions) {
-      if (this.props.token.uid !== prevProps.token.uid) {
-        this.cleanStates();
-      }
+    if ((this.props.tokensHistory && !prevProps.tokensHistory) ||
+        (this.props.tokensHistory[this.props.token.uid].length !== prevProps.tokensHistory[this.props.token.uid].length)) {
```
msbrogli, line 61: I feel we can avoid this whole method passing just the selected token history to the component.

src/constants.js
```diff
@@ -130,3 +131,8 @@ export const IPC_RENDERER = ipcRenderer;
  * Flag to hide/show elements after ledger integration is done
  */
 export const LEDGER_ENABLED = false;
+
+/**
+ * Store to be used in the wallet and all HathorWallet objects created
+ */
+export const STORE = new HybridStore();
```
msbrogli, line 138: I don't like putting this in the constants file. I feel we should create a `storeSingleton` in the `./storage.js` and import from there. Or we can create a `./storeInstance.js`, for instance.

src/screens/LoadWallet.js
```diff
@@ -91,8 +91,7 @@ class LoadWallet extends React.Component {
     // Getting redux variables before cleaning all data
     const { pin, password } = this.props;
     // First we clean what can still be there of a last wallet
-    wallet.cleanWallet();
-    wallet.generateWallet(this.state.words, '', pin, password, true);
+    wallet.generateWallet(this.state.words, '', pin, password, this.props.history);
```
msbrogli, line 94: You changed a bool parameter (previously `true`) to a list of transactions (`this.props.history`)?

src/screens/LockedWallet.js
```diff
@@ -48,22 +48,16 @@ class LockedWallet extends React.Component {
...
         return;
       }
 
-      const promise = version.checkApiVersion();
-      promise.then((data) => {
-        // Everything is fine, so redirect to wallet
-        hathorLib.wallet.unlock();
-        // Reload wallet data
-        wallet.reloadData();
+      const promise = wallet.startWallet(null, '', pin, '', this.props.history, true);
```
msbrogli, line 58: Can you just leave a comment explaining that it is safe to pass an empty string as the password? And what is the other empty string parameter?

src/screens/NewWallet.js
```diff
@@ -104,10 +104,8 @@ class NewWallet extends React.Component {
   pinSuccess = () => {
     // Getting redux variables before cleaning all data
     const { words, pin, password } = this.props;
-    // First we clean what can still be there of a last wallet
-    wallet.cleanWallet();
     // Generate addresses and load data
-    wallet.executeGenerateWallet(words, '', pin, password, true);
+    wallet.generateWallet(words, '', pin, password, this.props.history);
```
msbrogli, line 108: Same things here about the parameter type changing from boolean to a list.

src/screens/Wallet.js
```diff
@@ -131,29 +128,11 @@ class Wallet extends React.Component {
...
-          continue;
-        }
-
-        if (hathorLib.wallet.isMintOutput(output) || hathorLib.wallet.isMeltOutput(output)) {
-          return true;
-        }
-
-      }
-    }
+    return this.props.wallet.getMintAuthority(this.props.selectedToken, { skipSpent: false }) || this.props.wallet.getMeltAuthority(this.props.selectedToken, { skipSpent: false })
```
msbrogli, line 131: I feel this is hard to read. Maybe do something like:

```js
if (condA) {
  return true;
}
if (condB) {
  return true;
}
return false;
```

What do you think?

src/utils/wallet.js
```diff
@@ -5,10 +5,10 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import { SENTRY_DSN, DEBUG_LOCAL_DATA_KEYS } from '../constants';
+import { SENTRY_DSN, DEBUG_LOCAL_DATA_KEYS, STORE } from '../constants';
 import store from '../store/index';
-import { loadingAddresses, historyUpdate, sharedAddressUpdate, reloadData, cleanData } from '../actions/index';
-import hathorLib from '@hathor/wallet-lib';
+import { setWallet, updateLoadedData, isOnlineUpdate, updateHeight, newTx, updateTx, loadingAddresses, loadWalletSuccess, historyUpdate, sharedAddressUpdate, reloadData, cleanData, changeServer } from '../actions/index';
```
msbrogli, line 10: Maybe break into multiple lines?

src/utils/wallet.js
```diff
@@ -5,10 +5,10 @@
...
+import { helpers, constants as hathorConstants, errors as hathorErrors, HathorWallet, Connection, wallet as oldWalletUtil, walletUtils, storage, tokens } from '@hathor/wallet-lib';
```
msbrogli, line 11: Maybe break into multiple lines?

src/utils/wallet.js
```diff
@@ -55,17 +55,21 @@ const wallet = {
...
    */
-  generateWallet(words, passphrase, pin, password, loadHistory) {
-    if (hathorLib.wallet.wordsValid(words).valid) {
-      return this.executeGenerateWallet(words, passphrase, pin, password, loadHistory);
-    } else {
-      return null;
+  generateWallet(words, passphrase, pin, password, routerHistory) {
+    try {
+      walletUtils.wordsValid(words);
+      return this.startWallet(words, passphrase, pin, password, routerHistory);
```
msbrogli, line 65: Maybe we should move this out of the try/except block. As far as I understood, we not handling any exceptions that might be caused by it, right?

src/utils/wallet.js
```diff
@@ -77,30 +81,120 @@ const wallet = {
...
+      } else if (state === HathorWallet.READY) {
+        // READY
+        const historyTransactions = wallet.getTxHistory();
+        store.dispatch(loadWalletSuccess(historyTransactions));
+      }
+    });
+
+    wallet.on('new-tx', (tx) => {
+      let message = '';
+      if (helpers.isBlock(tx)) {
```
msbrogli, line 151: Let's open another PR to refactor this to another file? I feel we should have an internal notification bus just like we have in the full-node. It might be useful for future features.

I'm suggesting another PR because it feels too much for this one.

msbrogli (approved):
Outstanding PR! Congratulations! The code is way simpler to read and straight to the point.

There are still some minor changes to be done and we are ready to merge it.

src/reducers/index.js
```diff
@@ -74,7 +70,7 @@ const rootReducer = (state = initialState, action) => {
     case 'reload_data':
```
andreabadesso, line 74: *comment partially related to the line number*

I suggest using immerjs (https://immerjs.github.io/immer/) for state mutations

Maybe not on this PR, but I like to use immer to make sure that our state only mutates on reducers (that use immer)

src/reducers/index.js
```diff
@@ -89,15 +85,293 @@ const rootReducer = (state = initialState, action) => {
...
+
+    const newHistory = [...currentHistory];
+    newHistory[txIndex] = getTxHistoryFromTx(tx, tokenUid, tokenTxBalance)
+    updatedHistoryMap[tokenUid] = newHistory;
+  }
+
+  const newTokensHistory = Object.assign({}, state.tokensHistory, updatedHistoryMap);
+  const newTokensBalance = Object.assign({}, state.tokensBalance, updatedBalanceMap);
+
+  return Object.assign({}, state,{
```
andreabadesso, line 306: Missing space after `state,`

andreabadesso (changes requested):

src/components/WalletBalance.js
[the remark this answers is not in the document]
pedroferreira1: I agree we shouldn't change this component in this PR. Maybe it's not the best solution to have the loading screen just above the rest of the wallet but if we decide to change it's not a simple refactor.

src/App.js
pedroferreira1: No need to. The main goal of this method is to save in redux the registered tokens of the user (because it's the data we can't get from the full node).

We are already doing that in `startWallet` method here: `store.dispatch(reloadData({ tokens: dataToken }));`

src/screens/LoadWallet.js
pedroferreira1: Yes. The previous method used to receive `loadHistory` as the last parameter, which was responsible to load/not load the data from the full node when starting the wallet. This is not needed anymore and now I need this `routerHistory` (not a list of transactions) to handle the notification in case of a new tx event.

src/components/SendTokensOne.js
pedroferreira1: Done

src/screens/LockedWallet.js
pedroferreira1: Done.

src/screens/Wallet.js
pedroferreira1: Done

src/utils/wallet.js
pedroferreira1: Yes, you're right. Moved out.

src/utils/wallet.js
pedroferreira1: Created the issue: https://github.com/HathorNetwork/hathor-wallet/issues/194

src/components/tokens/TokenDestroy.js
```diff
@@ -64,9 +66,9 @@ class TokenDestroy extends React.Component {
    * @return {string} Error message, in case of form invalid. Nothing, otherwise.
    */
   destroy = () => {
-    if (this.state.destroyQuantity > this.props.authorityOutputs.length) {
+    if (this.state.destroyQuantity > this.props.authoritiesLength) {
       const type = this.props.action === 'destroy-mint' ? t`mint` : t`melt`;
-      return `You only have ${this.props.authorityOutputs.length} ${type} ${hathorLib.helpers.plural(this.props.authorityOutputs.length, 'output', 'outputs')} to destroy.`;
+      return `You only have ${this.props.authoritiesLength} ${type} ${hathorLib.helpers.plural(this.props.authoritiesLength, 'output', 'outputs')} to destroy.`;
```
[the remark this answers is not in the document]
pedroferreira1: Thanks!

src/App.js
```diff
@@ -227,9 +118,6 @@ const returnLoadedWalletComponent = (Component, props, rest) => {
   // Check version
   if (reduxState.isVersionAllowed === undefined && !isServerScreen) {
     const promise = version.checkApiVersion();
```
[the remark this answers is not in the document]
pedroferreira1: `version.checkApiVersion();` is actually a method that execute many important steps.

```
  checkApiVersion() {
    const newPromise = new Promise((resolve, reject) => {
      const libPromise = hathorLib.version.checkApiVersion();
      libPromise.then((data) => {
        // Update version allowed in redux
        store.dispatch(isVersionAllowedUpdate({allowed: hathorLib.helpers.isVersionAllowed(data.version, hathorLib.constants.MIN_API_VERSION)}));
        // Set network in lib to use the correct address byte
        let network;
        if (data.network === 'mainnet') {
          network = 'mainnet';
        } else {
          // Can we assume it will be testnet? I think it's safe
          network = 'testnet';
        }
        helpers.updateNetwork(network);
        resolve({...data, network});
      }, (error) => {
        reject();
      });
    });
    return newPromise;
  },
```

We don't need to wait for the promise to finish anymore to continue the code but we still need the rest of the code to be executed. I will just remove `const promise`, what do you think? Is this bad practices?

src/reducers/index.js
pedroferreira1: Humm I like this package and I've never seen it. Will create an issue to add this to the project in future improvements.

src/components/WalletBalance.js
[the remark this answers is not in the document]
msbrogli: Ok. Let's discuss it in a separate issue.

src/App.js
msbrogli: Ok.

src/screens/LoadWallet.js
msbrogli: Ok. I saw this change later in the changes. Thanks for the explanation.

src/App.js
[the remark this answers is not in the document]
msbrogli: I guess you have to at least add a handler in case of error.

r4mmer (commented):
Maybe we should include `eslint` in this project.

src/screens/NewWallet.js
pedroferreira1: Explained here: https://github.com/HathorNetwork/hathor-wallet/pull/188#discussion_r636906386

src/components/TokenAdministrative.js
pedroferreira1: I changed a bit your suggestion but completely removed the `componentDidUpdate` method. Token history was not used in the component, it was only used on state to understand when a new tx for the token was received to re-render the page. I used as key of component a concatenation of `tokenUid-tokenHistoryLen`. With this, every time a new tx for this token arrives the component will re-render. What do you think?

src/components/TokenHistory.js
```diff
@@ -158,74 +169,20 @@ class TokenHistory extends React.Component {
...
+          history.length === 0 ||
           (this.state.hasBefore === false && this.state.hasAfter === false)) {
         return null;
       } else {
         return (
           <nav aria-label="Token pagination" className="d-flex justify-content-center">
             <ul className="pagination">
-              <li className={(!this.state.hasBefore || this.props.history.length === 0) ? "page-item mr-3 disabled" : "page-item mr-3"}><a className="page-link" onClick={(e) => this.previousClicked(e)} href="true">{t`Previous`}</a></li>
-              <li className={(!this.state.hasAfter || this.props.history.length === 0) ? "page-item disabled" : "page-item"}><a className="page-link" href="true" onClick={(e) => this.nextClicked(e)}>{t`Next`}</a></li>
+              <li className={(!this.state.hasBefore || history.length === 0) ? "page-item mr-3 disabled" : "page-item mr-3"}><a className="page-link" onClick={(e) => this.previousClicked(e)} href="true">{t`Previous`}</a></li>
```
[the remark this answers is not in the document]
pedroferreira1: Done here: https://github.com/HathorNetwork/hathor-wallet/pull/188/commits/fa389aa2864d9ff401d50bbc24b81c3497529404

src/App.js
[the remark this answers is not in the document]
pedroferreira1: We already handle error cases in general opening the error modal. Should we do something else?

src/components/TokenHistory.js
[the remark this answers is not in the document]
pedroferreira1: The refactor was done and the issue was created: https://github.com/HathorNetwork/hathor-wallet/issues/196

src/constants.js
pedroferreira1: Done!

.github/workflows/main.yml
```diff
@@ -0,0 +1,31 @@
...
+    branches:
+    - dev
+    - master
+jobs:
+  test:
+    runs-on: 'ubuntu-latest'
+    timeout-minutes: 40  # default is 360
+    strategy:
+      matrix:
+        node-version: [8.x]
```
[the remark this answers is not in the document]
pedroferreira1: We currently fix the version only in v8.9.0 because some compatibility issues between Ledger and Electron packages. We already have an issue to handle this upgrade: https://github.com/HathorNetwork/hathor-wallet/issues/196 but it must be done carefully

src/components/TokenHistory.js
```diff
@@ -43,7 +50,9 @@ class TokenHistory extends React.Component {
   }
 
   componentDidUpdate = (prevProps) => {
-    if (this.props.history !== prevProps.history) {
+    if (this.props.tokensHistory && this.props.tokensHistory[this.props.selectedToken] &&
```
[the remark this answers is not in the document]
pedroferreira1: This code actually changed a bit in a newer version but the idea is that the `selectedToken` is the key of the component, then I don't need to handle this manually, I only need to trigger a dataUpdate when the data in redux (`tokensHistory`) changes.

src/components/TokenHistory.js
```diff
@@ -107,7 +117,8 @@ class TokenHistory extends React.Component {
    * Calculates the index of the reference hash in the history list
    */
   getReferenceIndex = () => {
-    const idxReference = this.props.history.findIndex((tx) =>
+    const history = this.props.tokensHistory[this.props.selectedToken];
+    const idxReference = history.findIndex((tx) =>
```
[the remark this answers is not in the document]
pedroferreira1: I added this validation but actually I am throwing an error because this method should never be called when `this.state.reference = null`. Every place we call it in the code we are validating this before. What do you think?

src/components/tokens/TokenDelegate.js
```diff
@@ -29,16 +36,24 @@ class TokenDelegate extends React.Component {
...
    * SendTransaction object that emit events while the tx is being sent and promise resolves when the sending is done
    * In case of error, an object with {success: false, message}
    */
   prepareSendTransaction = (pin) => {
-    const output = this.props.authorityOutputs[0];
     const type = this.props.action === 'delegate-mint' ? t`Mint` : t`Melt`;
-    return hathorLib.tokens.delegateAuthority(output.tx_id, output.index, output.decoded.address, this.props.token.uid, this.delegateAddress.current.value, this.delegateCreateAnother.current.checked, type.toLowerCase(), pin);
+    return this.props.wallet.delegateAuthority(
+      this.props.token.uid,
+      type.toLowerCase(),
```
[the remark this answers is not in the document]
pedroferreira1: The `dev` branch of wallet lib has been updated a few days ago, the changes haven't been propagated to `master` yet. Can you double check the arguments in the `dev` branch?

```
  delegateAuthority(tokenUid, type, destinationAddress, options = {}) {
    storage.setStore(this.store);
    const newOptions = Object.assign({ createAnother: true, startMiningTx: true, pinCode: null }, options);
    ...
```

src/components/tokens/TokenDestroy.js
```diff
@@ -28,25 +35,20 @@ class TokenDestroy extends React.Component {
...
-      data.push({
-        'tx_id': array[i].tx_id,
-        'index': array[i].index,
-        'address': array[i].decoded.address,
-        'token': this.props.token.uid
-      });
-    }
-    return hathorLib.tokens.destroyAuthority(data, pin);
+    const type = this.props.action === 'destroy-mint' ? 'mint' : 'melt';
+    return this.props.wallet.destroyAuthority(
```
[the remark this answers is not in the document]
pedroferreira1: The `dev` branch of wallet lib has been updated a few days ago, the changes haven't been propagated to `master` yet. Can you double check the arguments in the `dev` branch?

```
  destroyAuthority(tokenUid, type, count, options = {}) {
    storage.setStore(this.store);
    const newOptions = Object.assign({ startMiningTx: true, pinCode: null }, options);
    ...
```

andreabadesso (approved):

src/components/TokenHistory.js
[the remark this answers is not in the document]
r4mmer: The way the pagination is made, throwing an error on `this.state.reference = null` is the right thing.
I think the pagination could be simpler, but it's not something for this issue.

r4mmer (approved):

src/components/TokenAdministrative.js
msbrogli: It seems a clever idea. I've never seen the key being used this way but it should work nicely. I like it.

msbrogli (approved):

PR #207, opened 2021-06-15
fix: TokenBar balance crashes when opened if has no transactions for token without tx

pedroferreira1:


src/components/TokenBar.js
```diff
@@ -105,7 +105,11 @@ class TokenBar extends React.Component {
    */
   getTokenBalance = (uid) => {
     const balance = this.props.tokensBalance[uid];
-    const total = balance.available + balance.locked;
+    let total = 0;
+    if (balance) {
+      // If we don't have any transaction for the token, balance will be undefined
```
msbrogli, line 110: Shouldn't we ensure that balance will never be undefined (instead of handling this case here)?

msbrogli (changes requested):

src/components/TokenBar.js
pedroferreira1: We calculate the token balance when we receive a transaction for this token. We could start the redux tokenBalance variable with HTR balance as 0 but we would have a problem with new registered tokens that don't have any transactions. We could also have a default balance of 0 for any new registered token that does not have txs.

Not sure what's the best solution, I would go with the one in this PR, given that we already check if balance is undefined in other parts of the code but I can do the changes I said in the last paragraph. What do you think?

msbrogli: I agree this is the best solution given the current structure. Thanks for the explanation.

msbrogli (approved):

andreabadesso (approved):

PR #210, opened 2021-07-13
feat: Add support for multiple wallets

msbrogli:
# Acceptance criteria

- Manage multiple wallets (hardware and software) with different seeds (or passwords)
- Show user which wallet is currently in use
- Create a screen to change the wallet in use
- Allow user to change the wallet on the lock screen

# UI/UX

## Navbar

- Show current wallet on the top right

Screenshot from 2022-03-29 10-54-35

## Settings screen

- Add button to enter "Wallet list" screen

Screenshot from 2022-03-29 10-32-52

## Wallet List screen

- List all known registered wallets
- Clicking on a wallet will go to the unlock screen with the wallet selected
- Form to create a new wallet

Screenshot from 2022-03-29 10-33-18

## Lock screen

- Add dropdown to select which wallet to use
- Only show software wallets on this screen
    - Hardware wallets should be selected from the "Wallet List" screen

Screenshot from 2022-03-29 10-34-21

## Demo Video

https://user-images.githubusercontent.com/15909384/160683184-e6be5d7d-1e4f-4f4d-9d38-6cfa8e0ff1b1.mp4


src/screens/WalletList.js
```diff
@@ -0,0 +1,125 @@
...
+      // Always start with current wallet
+      return -1;
+    }
+    return key1 > key2;
+  }).map(([prefix, walletInfo]) => {
+    return (
+      <tr key={prefix}>
+        <td><a href="true" onClick={(e) => goToWallet(e, prefix)}>{prefix === props.walletPrefix ? t`Current`+": "+walletInfo.name : walletInfo.name}</a></td>
+      </tr>
+    );
```
andreabadesso, line 89:
```suggestion
    const walletTitle = prefix === props.walletPrefix ? t`Current` + ': ' + walletInfo.name : walletInfo.name;
    const onClick = (e) => goToWallet(e, prefix);

    return (
      <tr key={prefix}>
        <td>
          <a href="true" onClick={onClick}>
            {walletTitle}
          </a>
        </td>
      </tr>
    );
```

andreabadesso, line 89:
```suggestion
    const walletTitle = prefix === props.walletPrefix ? t`Current` + ': ' + walletInfo.name : walletInfo.name;
    const onClick = (e) => goToWallet(e, prefix);

    return (
      <tr key={prefix}>
        <td>
          <a href="true" onClick={onClick}>
            {walletTitle}
          </a>
        </td>
      </tr>
    );
```

andreabadesso (changes requested):

src/screens/WalletList.js
r4mmer: I'm resolving this since it's a duplicate of the one above

r4mmer: Done on [30b78f1](https://github.com/HathorNetwork/hathor-wallet/pull/210/commits/30b78f14c1ba8c1665304590bcce1eb58d7e849f)

src/screens/WalletList.js
```diff
@@ -0,0 +1,125 @@
...
+  const onNewWalletClick = (e) => {
+    e.preventDefault();
+    setError(null);
+    if (!walletNameRef.current) {
+      return;
+    }
+    const walletName = walletNameRef.current?.value;
+    try {
+      hathorLib.storage.store.addWallet(walletName, walletName);
+      // walletNameRef.current?.value = '';
```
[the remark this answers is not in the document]
r4mmer: Removed on [30b78f1](https://github.com/HathorNetwork/hathor-wallet/pull/210/commits/30b78f14c1ba8c1665304590bcce1eb58d7e849f)

src/screens/Settings.js
```diff
@@ -62,8 +75,17 @@ class Settings extends React.Component {
...
+      // Remove from list of wallets
+      hathorLib.storage.store.removeWallet(walletPrefix);
+    }
+    // Set to the default prefix
+    this.props.setWalletPrefix('');
     wallet.resetWalletData();
-    this.props.history.push('/welcome/');
+    hathorLib.wallet.lock();
+    this.props.history.push('/');
+    // this.props.history.push('/welcome/');
```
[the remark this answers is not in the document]
r4mmer: Removed on [30b78f1](https://github.com/HathorNetwork/hathor-wallet/pull/210/commits/30b78f14c1ba8c1665304590bcce1eb58d7e849f)

src/components/WalletStatus.js
```diff
@@ -0,0 +1,41 @@
...
+  };
+};
+
+/**
+ * Component that shows the current wallet name in use
+ *
+ * @memberof Components
+ */
+function WalletStatus(props) {
+  // if (props.isLocked === true) {
```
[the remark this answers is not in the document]
r4mmer: Removed on [30b78f1](https://github.com/HathorNetwork/hathor-wallet/pull/210/commits/30b78f14c1ba8c1665304590bcce1eb58d7e849f)

andreabadesso (approved):

pedroferreira1:
Before starting the code review I have some feedback about UI/UX:

1. We should limit the max quantity of chars in the wallet name (maybe to 50 chars?) and reduce the max width of the wallet name presentation, so it doesn't reduce usability. See images below:

image
image

2. When you reset the wallet and you go through the same process of starting a wallet as we have nowadays, the wallet will always be added as "Default". Shouldn't we give the possibility for the user to select a wallet name?

src/screens/Settings.js
```diff
@@ -62,8 +75,16 @@ class Settings extends React.Component {
    */
   handleReset = () => {
```
pedroferreira1, line 76: I don't think I like this new Reset flow. As far as I understood it's the following:

1. If it's a default wallet, then you don't remove it from the list (i.e., you always need at least one wallet)
2. If it's not a default wallet, then you remove it from the list and you go to the welcome screen (why did you change the `push` to `/` instead of `welcome`?) And you did reset a custom wallet but when you start from the beginning you are required to start a new wallet that will substitute the default one.

Am I missing something here? I feel we should discuss this UX

src/screens/LockedWallet.js
```diff
@@ -77,6 +89,16 @@ class LockedWallet extends React.Component {
     $('#confirmResetModal').modal('show');
   }
 
+  /**
+   * When user clicks to change the wallet, it will change the current selected wallet
+   *
+   * @param {Object} e Event of when the link is clicked
+   */
+   changeWalletClicked = (e) => {
```
pedroferreira1, line 97: I feel this method name is misleading. There is no click, it's when the select changes. The param docstring should be changed as well.

src/screens/Settings.js
r4mmer: By default we will go to the `Default` wallet once we reset a wallet. Then we enter one of three cases:
- If the `Default` wallet was reset we should go to `/welcome`
- if it already exists we should go to `/locked`
- if it's a Hardware wallet we should go to `/wallet_type`

This method to choose which path happens naturally when going to the root path. That's why I chose to change from the `/welcome` to `/`.

The default wallet is the wallet that uses no prefix on the keys (this will make upgrading from an older version work seamlessly) but once it is gone we can only create wallets with a prefix.I thought we should keep the `Default` wallet as always existing because it creates a better experience when resetting a wallet, the user can choose which is his default wallet and he will always get that behavior.
This could change if we rework the init of a hardware wallet though, If the `/locked` was the default start screen of all wallets we could change to whatever wallet and the experience wouldn't change.

src/screens/LockedWallet.js
r4mmer: Changed to `changeWalletHandler`

src/storage.js
```diff
@@ -53,9 +53,42 @@ class LocalStorageStore {
...
+
   constructor() {
     this.memStore = new hathorLib.MemoryStore();
     this.persistentStore = new LocalStorageStore();
+    this.prefix = '';
+    this.nonPrefixedKeys = new Set(HybridStore.nonPrefixedKeyList);
+  }
+
+  getListOfWallets() {
+    return this.getItem('list-of-wallets') || {'': { name: 'Default' }};
```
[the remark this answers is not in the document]
r4mmer: Created a new method for setting `list-of-wallets` so we don't have to use it on other places

src/storage.js
```diff
@@ -53,9 +53,42 @@ class LocalStorageStore {
  * It's not a problem if 'wallet:data' is not persisted, as this data is always loaded when we connect to the server.
  */
 class HybridStore {
+  static nonPrefixedKeyList = ['list-of-wallets'];
```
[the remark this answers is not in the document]
r4mmer: Added

src/screens/WalletList.js
```diff
@@ -0,0 +1,126 @@
...
+        <td><a href="true" onClick={onclick}>{walletTitle}</a></td>
+      </tr>
+    );
+  });
+
+  const newWalletForm = (
+    <form ref={newWalletFormRef} className="w-100">
+      <div className="row">
+        <div className="col-6">
+          <input ref={walletNameRef} type="text" autoComplete="off" placeholder={t`Wallet Name`} className="form-control" required />
```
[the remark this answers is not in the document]
r4mmer: Done

src/screens/WalletList.js
```diff
@@ -0,0 +1,126 @@
...
+      return;
+    }
+    const walletName = walletNameRef.current?.value;
+    try {
+      hathorLib.storage.store.addWallet(walletName, walletName);
+      // This will trigger a state change which will reload the component.
+      setListOfWallets(hathorLib.storage.store.getListOfWallets());
+    } catch (ex) {
+      // wallet already exists
+      console.log(ex);
```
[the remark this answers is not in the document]
r4mmer: Removed

src/screens/WalletList.js
```diff
@@ -0,0 +1,126 @@
...
+    setError(null);
+    if (!walletNameRef.current) {
+      return;
+    }
+    const walletName = walletNameRef.current?.value;
+    try {
+      hathorLib.storage.store.addWallet(walletName, walletName);
+      // This will trigger a state change which will reload the component.
+      setListOfWallets(hathorLib.storage.store.getListOfWallets());
+    } catch (ex) {
```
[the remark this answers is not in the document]
r4mmer: Done

src/storage.js
```diff
@@ -53,9 +53,42 @@ class LocalStorageStore {
...
+  }
+
+  removeWallet(prefix) {
+    const listOfWallets = this.getListOfWallets();
+    if (listOfWallets[prefix] !== undefined) {
+      delete listOfWallets[prefix];
+      this.setItem('list-of-wallets', listOfWallets);
+    } else {
+      // The wallet being deleted does not exist
+      throw new Error('Wallet does not exist.');
```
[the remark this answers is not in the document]
r4mmer: Done

src/storage.js
```diff
@@ -53,9 +53,42 @@ class LocalStorageStore {
...
+    }
+  }
+
+  addWallet(name, prefix) {
+    const listOfWallets = this.getListOfWallets();
+    if (listOfWallets[prefix] !== undefined) {
+      throw new Error("This wallet prefix is already in use.");
```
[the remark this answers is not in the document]
r4mmer: Done

src/errors.js
```diff
@@ -0,0 +1,27 @@
...
+export class WalletDoesNotExistError extends Error {
+  constructor(prefix) {
+    super(`Wallet does not exist: ${prefix}`);
+    this.name = "WalletDoesNotExistError";
+  }
+};
+
+export class WalletAlreadyExistError extends Error {
+  constructor() {
+    super('Prefix already in use');
```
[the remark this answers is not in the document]
r4mmer: As discussed privately this message will be shown to the user above the input with the prefix.

obiyankenobi:
Made changes to the UI/UX. I still need to push the code changes, but wanted to get feedback if this seems like a good UX for the user.

The user does not have a default wallet now. And naming the wallet will be part of its initialization process. Apart from that, hardware wallets cannot be named/saved. They will always be temporary.

https://user-images.githubusercontent.com/3236756/179064173-a0028b51-fd17-45ba-be52-cfd4735ddb14.mov

Missing tasks:
- [ ] test wallet upgrade ('sanitize' store), when moving from a wallet that does not support multiple wallets;
- [ ] check if there is another wallet with same name when creating a new wallet;
- [ ] add modal when clicking change wallet in the Settings screen;

@trondbjoroy I'd like your input for a text to be shown on the modal before proceeding with changing the wallet;

trondbjoroy:
@obiyankenobi How about adding a modal header called "Change wallet" (alternatively, just as a headline right above the main text) 

And the following explanatory text: "To change to a different software wallet, select one from the list below or add a new one. You may also connect to a hardware wallet."

Small typo:
typo

obiyankenobi:
@trondbjoroy I actually mean a modal before showing the screen to choose wallets.

image

src/screens/Settings.js
obiyankenobi: Closing as the UX changed greatly.

obiyankenobi:
UI/UX has changed completely. 

https://user-images.githubusercontent.com/3236756/184239776-00a7e418-39ae-41dd-852f-04fba267f7d7.mov

@trondbjoroy what do you think about the UX? I think having multiple wallets greatly complicates things for users, but not sure there's a better way to do it. I think users with multiple wallets will be a minority and it may become more complex for average users.

src/storage.js
```diff
@@ -53,9 +60,116 @@ class LocalStorageStore {
...
+  /*
+   * Upgrades an old version of the storage, which did not support multiple wallets
+   *
+   * @param {String} walletPrefix Prefix used to identify the wallet.
+   * @param {String} walletName Name used to identify the wallet.
+   *
+   * @memberof HybridStore
+   * @inner
+   */
+  upgradeStorage(walletPrefix, walletName) {
```
r4mmer, line 81: Should we have a `storage:version` key to check if we have already upgraded this storage?

src/utils/wallet.js
```diff
@@ -1060,6 +1061,69 @@ const wallet = {
...
+   *
+   * @param {string} name The name of the wallet
+   *
+   * @return {string} Prefix of the wallet
+   *
+   * @memberof Wallet
+   * @inner
+   */
+  setWalletPrefix(prefix) {
+    storage.store.prefix = prefix;
```
r4mmer, line 1091: Should we also dispatch an action to have this prefix on the redux state?

src/screens/Signin.js
```diff
@@ -9,35 +9,85 @@ import React from 'react';
...
 class Signin extends React.Component {
+  /**
+   * errorMessage {string} Message to be shown when an error happens
+   */
+  state = { errorMessage: '' };
+
+  /**
+   * Check wallet name is not blank or already in use
+   */
+  validateWalletName = (name) => {
```
r4mmer, line 38: Should we have a validation for forbidden characters?

Also, do we have any words we would like to reserve?
Once we deliver we will not be able to cleanly forbid words since users may have them already added.

I saw that we use `maxLength` on the input, but should we have this validation here so that we can control the error message and format?

src/screens/LockedWallet.js
```diff
@@ -120,16 +120,27 @@ class LockedWallet extends React.Component {
...
-    this.props.history.push('/welcome/');
+    // If there are other wallets, go to screen to choose wallet
+    const firstWallet = wallet.getFirstWalletPrefix();
+    if (firstWallet) {
+      wallet.setWalletPrefix(firstWallet);
+      this.props.history.push('/choose_wallet');
+    } else {
+      wallet.setWalletPrefix(null);
+      this.props.history.push('/');
+    }
```
r4mmer, line 135: Is this screen relevant?
We are adding more information to this screen but the "lock" now goes to the choose_wallet screen?
If it's not used anymore, can we remove it?

src/screens/ChooseWallet.js
```diff
@@ -0,0 +1,209 @@
...
+    if (this.state.loading) return;
+    wallet.setWalletPrefix(null);
+    hathorLib.wallet.cleanWallet({ cleanAccessData: false });
+    this.props.history.push('/signin');
+  }
+
+  /**
+   * Called when user clicks on "Hardware Wallet" button.
+   */
+  goToHardwareWallet = () => {
```
r4mmer, line 78: Does the hardware wallet back button land on this page?

pedroferreira1:
Before reviewing the code I was running the branch to see the flow and UI and I guess I've found a bug.

Steps to reproduce:

1. Load a wallet and change the server to a testnet server.
2. Settings > Change Wallet > Import another software wallet
3. This new wallet won't load (this happened to me)
4. Close and open the wallet, it will have two loaded wallets

Now you will be able to load any of the wallets but, once you load the first, you won't be able to load the second.

obiyankenobi:
> Before reviewing the code I was running the branch to see the flow and UI and I guess I've found a bug.
> 
> Steps to reproduce:
> 
> 1. Load a wallet and change the server to a testnet server.
> 2. Settings > Change Wallet > Import another software wallet
> 3. This new wallet won't load (this happened to me)
> 4. Close and open the wallet, it will have two loaded wallets
> 
> Now you will be able to load any of the wallets but, once you load the first, you won't be able to load the second.

Fixed on https://github.com/HathorNetwork/hathor-wallet/pull/210/commits/601076c6a909107a2a04d67c4f63fe845cd7ef11

src/screens/Settings.js
```diff
@@ -74,8 +75,19 @@ class Settings extends React.Component {
...
     $('#confirmResetModal').modal('hide');
+    // Remove from list of wallets
```
pedroferreira1, line 78: This is pretty much the same code that we have in the `LockedWallet` file, maybe we could create a method in the wallet utils to do this, it just needs to receive the history to push.

src/screens/ChooseWallet.js
```diff
@@ -0,0 +1,215 @@
...
+ *
+ * @memberof Screens
+ */
+class ChooseWallet extends React.Component {
+  constructor(props) {
+    super(props);
+
+    // update storage in case it did not support multiple wallets (wallet upgrade)
+    const wallets = hathorLib.storage.store.getListOfWallets();
+    if (!wallets || Object.keys(wallets).length === 0){
```
pedroferreira1, line 32: If the user removes all his wallets, are we going to assume that it's a storage upgrade?

src/screens/ChooseWallet.js
```diff
@@ -0,0 +1,215 @@
...
+    
+    this.setState({ selectedWallet: prefix, errorMessage: null });
+  }
+
+  /**
+   * Called when user clicks on "Add Wallet" button.
+   */
+  addWallet = () => {
+    if (this.state.loading) return;
+    wallet.setWalletPrefix(null);
```
pedroferreira1, line 71: If the user is in the `ChooseWallet` screen, the prefix should be `null` and the wallet data should've been cleaned already, right?

(The same for the hardware wallet and the code there is really similar to this one, we should reuse part of it)

Edit: forget about the prefix, you need to select it if the user clicks to remove a wallet but the data should be cleaned already, I guess.

src/App.js
```diff
@@ -68,8 +69,17 @@ class Root extends React.Component {
   componentDidUpdate(prevProps) {
     // When Ledger device loses connection or the app is closed
     if (this.props.ledgerClosed && !prevProps.ledgerClosed) {
-      hathorLib.wallet.lock();
-      this.props.history.push('/wallet_type/');
```
pedroferreira1, line 72: In this wallet type screen we are showing an alert message if the ledger app was closed nowadays. With this change, we might never go to this screen and only to the choose wallet, so we should add this alert there as well.

src/App.js
```diff
@@ -218,6 +229,11 @@ const returnStartedRoute = (Component, props, rest) => {
     return <Redirect to={{pathname: '/wallet/'}}/>;
   }
 
+  const firstWallet = wallet.getFirstWalletPrefix();
```
pedroferreira1, line 232: Maybe add a comment here. If I understood correctly, this is used when you've just reset your wallet, is that correct?

src/screens/ChooseWallet.js
obiyankenobi: If the user removes all his wallets, he will go to the initial screens, where he will add a new one, so the list won't be empty. The only way we get to this condition on the ChooseWallet screen is during a wallet upgrade.

src/screens/LockedWallet.js
obiyankenobi: As I understand, this might be used with the wallet service. The user might be required to input his pin from time to time.

src/App.js
alexruzenhack: I didn't understand.

src/App.js
```diff
@@ -253,11 +269,16 @@ const StartedRoute = ({component: Component, ...rest}) => (
...
       wallet.cleanWallet();
+      wallet.removeHardwareWalletFromStorage();
       hathorLib.wallet.unlock();
-      return <Redirect to={{ pathname: '/wallet_type/' }} />;
+      const firstWallet = wallet.getFirstWalletPrefix();
+      if (firstWallet) {
+        return <Redirect to={{ pathname: '/choose_wallet/' }} />;
+      } else {
+        return <Redirect to={{ pathname: '/wallet_type/' }} />;
+      }
```
alexruzenhack, line 281: This structure repeats some times.

PR #212, opened 2021-07-29
feat: ledger update

r4mmer:


src/constants.js
```diff
@@ -129,4 +129,4 @@ export const IPC_RENDERER = ipcRenderer;
 /**
  * Flag to hide/show elements after ledger integration is done
  */
-export const LEDGER_ENABLED = false;
+export const LEDGER_ENABLED = true;
```
msbrogli, line 132: Not yet.

msbrogli (changes requested):

public/ledger.js
```diff
@@ -17,20 +17,68 @@ const ledgerCLA = 0xe0;
...
+        case 0xb009:  // SW_INVALID_TX
         default:
+          console.log("LedgerError: ", errorMap[e.statusCode] || 'UNKNOWN_ERROR');
           return {status: e.statusCode, message: 'Error communicating with Ledger'};
       }
     }
     return e;
   }
 
+  static formatPathData(index) {
```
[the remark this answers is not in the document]
msbrogli: Done. Thanks!

src/components/WalletAddress.js
```diff
@@ -81,7 +81,7 @@ class WalletAddress extends React.Component {
 
     if (hathorLib.wallet.isHardwareWallet()) {
       $('#ledgerAlert').modal('show');
-      ledger.checkAddress(hathorLib.transaction.intToBytes(this.props.lastSharedIndex, 4));
+      ledger.checkAddress(this.props.lastSharedIndex);
```
pedroferreira1, line 84: You must update `checkAddress` docstring. It's still saying that receives a buffer.

src/utils/ledger.js
```diff
@@ -71,12 +89,10 @@ if (IPC_RENDERER) {
       // first assemble data to be sent
       const arr = [];
       if (changeIndex > -1) {
-        // With change output
-        arr.push(hathorLib.transaction.intToBytes(1, 1));
-        // Change index of array
-        arr.push(hathorLib.transaction.intToBytes(changeIndex, 1));
+        // encode the bit indicating existence of change and change index on first byte
+        arr.push(hathorLib.transaction.intToBytes(0x80 | changeIndex, 1))
```
pedroferreira1, line 93: You use the first bit to indicate the existence of change output and then the other 7 bits you identify what's the index?

If that's the case, we can have only 128 outputs, right?

r4mmer: Thanks, this is wrong. I tested this case on a script so the ledger parses ok, but i didn't test the wallet on this. It should be 1 byte for `has_change_output + len(change_bip32_path)` another byte for index and then the actual path. since the path length can be at most 5, 7 bits is more than enough.

src/components/WalletAddress.js
r4mmer: Done.

src/screens/SendTokens.js
```diff
@@ -136,6 +136,9 @@ class SendTokens extends React.Component {
       const instance = ref.current;
       const dataOne = instance.getData();
       if (!dataOne) return;
+      if (hathorLib.wallet.isHardwareWallet()) {
+        dataOne = instance.handleInitialData(dataOne);
```
msbrogli, line 140: Why don't we do it always? I mean, why only for hardware wallets?

src/screens/SendTokens.js
```diff
@@ -234,6 +237,8 @@ class SendTokens extends React.Component {
         this.data = data;
         $('#pinModal').modal('show');
       } else {
+        // currently we only support HTR
+        data.tokens = [];
```
msbrogli, line 241: I feel we should make this verification beforehand and show an error message accordingly. Here we could assert that `data.tokens == []`.

src/components/SendTokensOne.js
```diff
@@ -132,6 +132,25 @@ class SendTokensOne extends React.Component {
     return data;
   }
 
+  /**
+   * Validate inputs and outpus
+   * 1. If inputs were not selected, select inputs from outputs amount
+   * 2. If amount of selected inputs is larger than outputs amount, we create a change output
+   * 3. If inputs were selected, check if they are valid and add address key to input
+   */
+  handleInitialData = (data) => {
```
msbrogli, line 141: As the docstring is "Validate inputs and outputs", shouldn't we name this method something like "validateInputsAndOutputs"? I feel it is more intuitive.

msbrogli (changes requested):

src/constants.js
r4mmer: Changed back

src/components/WalletAddress.js
```diff
@@ -135,7 +135,9 @@ class WalletAddress extends React.Component {
               </div>
             }
           </div>
-          <a href="true" onClick={this.seeAllAddresses} className="mt-3 ">{t`See all addresses`}</a>
+          {hathorLib.wallet.isSoftwareWallet() &&   // hide all addresses for hardware wallet
```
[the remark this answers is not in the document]
r4mmer: Ok

pedroferreira1 (approved):

src/screens/SendTokens.js
r4mmer: The software wallet uses another method called `this.props.wallet.sendManyOutputsTransaction`

src/screens/SendTokens.js
r4mmer: I don't know how this would affect the software wallet, but we could add it to `getData` method on `src/components/SendTokensOne.js`.

src/components/SendTokensOne.js
r4mmer: Done.

src/screens/SendTokens.js
msbrogli: You misunderstood me. I don't think we should do `data.tokens = []` at all. We should check whether the list of tokens is empty and raise an exception if it is not.

pedroferreira1 (approved):

msbrogli (approved):

PR #223, opened 2021-08-30
Chore: update ledger qa steps

r4mmer:
# Acceptance criteria

- [x] New steps test all ledger functionalities
- [x] New steps explain clearly the ledger app UI

### DIFF
--- a/QA.md
+++ b/QA.md
@@ -168,9 +168,9 @@ You can connect your wallet to the testnet (https://node1.foxtrot.testnet.hathor
     1. Choose 'Hardware Wallet'.
     1. Wallet should go through both steps and show 'Step 1/2' and 'Step 2/2'.
     1. On your Ledger, check it is requesting your authorization.
-    1. Deny the authorization request on Ledger (left button). It should show an error on the wallet.
+    1. Deny the authorization request on Ledger (scroll with left or right and click both buttons on "Reject" step). It should show an error on the wallet.
     1. Click 'Try again'. It goes through both steps and asks for authorization again.
-    1. Grant authorization request (right button). It will proceed to 'Loading transactions' screen.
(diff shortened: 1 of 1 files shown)


QA.md
```diff
@@ -168,17 +168,17 @@ You can connect your wallet to the testnet (https://node1.foxtrot.testnet.hathor
...
+    1. Grant authorization request (click both buttons on the "Approve" step). It will proceed to 'Loading transactions' screen.
 
 1. **Ledger wallet screen**
     1. On the main wallet screen, confirm 'Address to receive tokens' is truncated (eg: 'HGZmqThwa6...'). There should be a 'Show full address' button next to it.
     1. Confirm there's no QR Code button under the truncated address.
     1. Click on 'Show full address'. A modal will display asking you to compare the address on the wallet and on Ledger.
     1. Check that the modal cannot be closed, not even by clicking outside it.
     1. Confirm that both addresses (on wallet and Ledger) are the same. Copy this address on the computed.
-    1. Click both buttons on Ledger. Modal should close.
+    1. On the Ledger click both buttons on the "Approve" step. Modal should close.
```
luislhl, line 181: Just curious. Why all the steps are `1.`?

luislhl (approved):

QA.md
r4mmer: Had the same question, but did not want to change the structure. Maybe @pedroferreira1 knows?

pedroferreira1: You can have a numbered list only with 1's, the main benefit of using this syntax is that you can add a new element in the list in any position and you don't need to update all the list elements, e.g. if you have a list from 1 to 10 and you want to add a new element in position 3, you would need to update numbers 3 to 4, 4 to 5, .... with this syntax you just add a new element with 1. ...

pedroferreira1 (approved):

PR #224, opened 2021-09-09
fix: package lock rollback for node v8.9.0 format

pedroferreira1:
A change on package-lock.json (adding the wallet-lib dependencies inside the wallet-lib object instead of outside) created a bug when signing a transaction. The bitcore lib was not being installed correctly.

### Acceptance criteria

- Fix package-lock.json in order to install bitcore lib correctly and the input signatures continue working.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1147,64 +1147,6 @@
         "lodash": "4.17.15",
         "long": "4.0.0",
         "ws": "7.2.3"
-      },
-      "dependencies": {
-        "bitcore-lib": {
-          "version": "8.25.10",
-          "resolved": "https://registry.npmjs.org/bitcore-lib/-/bitcore-lib-8.25.10.tgz",
-          "integrity": "sha512-MyHpSg7aFRHe359RA/gdkaQAal3NswYZTLEuu0tGX1RGWXAYN9i/24fsjPqVKj+z0ua+gzAT7aQs0KiKXWCgKA==",
-          "requires": {
-            "bech32": "1.1.3",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -11406,7 +11380,9 @@
         "fsevents": {
           "version": "2.1.2",
           "resolved": "https://registry.npmjs.org/fsevents/-/fsevents-2.1.2.tgz",
-          "integrity": "sha512-R4wDiBwZ0KzpgOWetKDug1FZcYhqYnUYKtfZYt4mD5SBz76q0KR4Q9o7GIPamsVPGmW3EYPPJ0dOOjvx32ldZA=="
+          "integrity": "sha512-R4wDiBwZ0KzpgOWetKDug1FZcYhqYnUYKtfZYt4mD5SBz76q0KR4Q9o7GIPamsVPGmW3EYPPJ0dOOjvx32ldZA==",
+          "dev": true,
+          "optional": true
```
andreabadesso, line 11385: What is this?

andreabadesso (approved):

r4mmer (approved):

package-lock.json
pedroferreira1: > optional: If true then this dependency is either an optional dependency ONLY of the top level module or a transitive dependency of one. This is false for dependencies that are both an optional dependency of the top level and a transitive dependency of a non-optional dependency of the top level.

This is optional because `fsevents` won't be installed on linux/windows, only on macos (https://github.com/npm/npm/issues/17722)

PR #228, opened 2021-09-22
feat: customize wallet information if token is an NFT

pedroferreira1:
Different from the explorer PR, here I am not writing 'Loading' when I need to show an amount until the metadata is loaded. The main reason for that is because the explorer needs to request the token metadata every time it opens a screen that has a token, then the screen would always 'blink'. 

In the wallet, this is different. I load the metadata of all tokens that belong to the wallet and then I can show the information every time without 'blinking' the screen. There might be a moment on the wallet start after loading the transactions and before loading all metadatas but we always start showing HTR token, then it's unlikely to happen and, if happens, won't be much a problem.

One possible problem that could happen is sending a transaction before loading all metadata because we change the input number format for NFT tokens. Because of that, I'm blocking sending/mint/melt screens while the metadatas are not fully loaded.

### Main screen

image
image

### About token

image
image

### Token admin

image
image

### Send tokens

image

### Transaction detail

image

### NFT List

image
image
image

### DIFF
--- a/src/components/NFTListElement.js
+++ b/src/components/NFTListElement.js
@@ -1,0 +1,103 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import { t } from 'ttag'
+import { NFT_MEDIA_TYPES, VIDEO_MEDIA_TYPES_BY_EXTENSION, AUDIO_MEDIA_TYPES_BY_EXTENSION } from '../constants';
+import helpers from '../utils/helpers';
+import path from 'path';
+
+
+/**
+ * Show a NFT element in the NFT list
+ *
+ * @memberof Components
+ */
+class NFTListElement extends React.Component {
+  /**
+   * Method called when user clicked on 'See on explorer'
+   * Go to explorer token detail page
+   *
+   * @param {Object} e Event for the click
+   * @param {String} token Token uid
+   */
+  goToTokenDetail = (e, token) => {
+    e.preventDefault();
+    const url = path.join(helpers.getExplorerURL(), `token_detail/${token}`);
+    helpers.openExternalURL(url);
+  }
+
+  render = () => {
+    const nftType = this.props.nftElement.nft_media.type && this.props.nftElement.nft_media.type.toUpperCase();
+
+    const file = this.props.nftElement.nft_media.file;
+
+    const ext = helpers.getFileExtension(file);
+
+    let fileType;
+
+    if (nftType === NFT_MEDIA_TYPES.audio) {
+      fileType = AUDIO_MEDIA_TYPES_BY_EXTENSION[ext];
+    }
+
+    if (nftType === NFT_MEDIA_TYPES.video) {
+      fileType = VIDEO_MEDIA_TYPES_BY_EXTENSION[ext];
+    }
+
+    let media;
+
+    if (nftType === NFT_MEDIA_TYPES.image) {
+      media = <img src={file} width="100%" height="100%" alt="NFT Preview" />;
+    } else if (nftType === NFT_MEDIA_TYPES.video && fileType) {
+      media = (
+        <video
+          controls
+          controlsList="nodownload noremoteplayback"
+          disablePictureInPicture
+          loop={this.props.nftElement.nft_media.loop}
+        >
+          <source src={file} type={fileType} />
+          Your browser does not support html video tag.
+        </video>
+      )
+    } else if (nftType === NFT_MEDIA_TYPES.audio && fileType) {
+      media = (
+        <audio
+          controls
+          controlsList="nodownload"
+          loop={this.props.nftElement.nft_media.loop}
+        >
+          <source src={file} type={fileType} />
+          Your browser does not support the audio element.
+        </audio>
+      )
+    } else if (nftType === NFT_MEDIA_TYPES.pdf) {
+      // Toolbar to prevent showing download/print icons
+      const data = `${file}#toolbar=0`;
+      media = <object data={data} width="100%" height="100%" type="application/pdf" alt="NFT Preview" aria-label="NFT Preview" />;
+    } else {
+      media = <p> Preview Unavailable </p>
+    }
+
+    return (
+      <>
+        <div className='d-flex align-items-lg-stretch mt-4 mt-lg-0'>
+          <div className="d-flex flex-column token-nft-preview mt-5">
+            <p><strong>{this.props.nftElement.name}</strong></p>
+            <figure className="figure flex-fill p-4 d-flex align-items-center justify-content-center">
+              { media }
+            </figure>
+            <p><strong>Balance: </strong>{this.props.nftElement.balance} {this.props.nftElement.symbol}</p>
+            <p><a href="true" onClick={(e) => this.goToTokenDetail(e, this.props.nftElement.id)}>{t`See on explorer`}</a></p>
+          </div>        
+        </div>
+      </>
+    )
+  }
+}
+
+export default NFTListElement;
--- a/src/components/NFTListElement.js
+++ b/src/components/NFTListElement.js
@@ -9,7 +9,6 @@ import React from 'react';
 import { t } from 'ttag'
 import { NFT_MEDIA_TYPES, VIDEO_MEDIA_TYPES_BY_EXTENSION, AUDIO_MEDIA_TYPES_BY_EXTENSION } from '../constants';
 import helpers from '../utils/helpers';
-import path from 'path';
 
 
 /**
@@ -27,7 +26,7 @@ class NFTListElement extends React.Component {
    */
   goToTokenDetail = (e, token) => {
     e.preventDefault();
-    const url = path.join(helpers.getExplorerURL(), `token_detail/${token}`);
+    const url = helpers.getFullExplorerURL(`token_detail/${token}`);
     helpers.openExternalURL(url);
   }
 
--- a/src/components/NFTListElement.js
+++ b/src/components/NFTListElement.js
@@ -91,7 +91,7 @@ class NFTListElement extends React.Component {
             <figure className="figure flex-fill p-4 d-flex align-items-center justify-content-center">
               { media }
             </figure>
-            <p><strong>Balance: </strong>{this.props.nftElement.balance} {this.props.nftElement.symbol}</p>
+            <p><strong>Balance: </strong>{helpers.renderValue(this.props.nftElement.balance, true)} {this.props.nftElement.symbol}</p>
             <p><a href="true" onClick={(e) => this.goToTokenDetail(e, this.props.nftElement.id)}>{t`See on explorer`}</a></p>
           </div>        
         </div>
--- a/src/components/SendTokensOne.js
+++ b/src/components/SendTokensOne.js
@@ -90,6 +90,7 @@ class SendTokensOne extends React.Component {
    * Iterate through inputs and outputs to add each information typed to the data object
    */
   getData = () => {
+    const isNFT = this.state.selected.uid in this.props.tokenMetadata && this.props.tokenMetadata[this.state.selected.uid].nft;
     let data = {'outputs': [], 'inputs': []};
     for (const output of this.outputs) {
       const address = output.current.address.current.value;
@@ -97,9 +98,9 @@ class SendTokensOne extends React.Component {
 
       if (address && valueStr) {
         // Doing the check here because need to validate before doing parseInt
-        const tokensValue = wallet.decimalToInteger(valueStr);
+        const tokensValue = isNFT ? parseInt(valueStr) : wallet.decimalToInteger(valueStr);
         if (tokensValue > hathorLib.constants.MAX_OUTPUT_VALUE) {
-          this.props.updateState({ errorMessage: `Token: ${this.state.selected.symbol}. Output: ${output.current.props.index}. Maximum output value is ${hathorLib.helpers.prettyValue(hathorLib.constants.MAX_OUTPUT_VALUE)}` });
+          this.props.updateState({ errorMessage: `Token: ${this.state.selected.symbol}. Output: ${output.current.props.index}. Maximum output value is ${helpers.renderValue(hathorLib.constants.MAX_OUTPUT_VALUE, isNFT)}` });
           return null;
         }
         let dataOutput = {'address': address, 'value': parseInt(tokensValue, 10), 'token': this.state.selected.uid};
@@ -184,9 +185,11 @@ class SendTokensOne extends React.Component {
   }
 
   render = () => {
+    const isNFT = this.state.selected && this.state.selected.uid in this.props.tokenMetadata && this.props.tokenMetadata[this.state.selected.uid].nft;
+
     const renderOutputs = () => {
       return this.outputs.map((output, index) =>
-        <OutputsWrapper key={index} index={index} ref={output} addOutput={this.addOutput} />
+        <OutputsWrapper key={index} index={index} ref={output} addOutput={this.addOutput} isNFT={isNFT} />
       );
     }
 
@@ -218,22 +221,30 @@ class SendTokensOne extends React.Component {
 
     const renderBalance = () => {
       let availableBalance = 0;
-      const isNFT = this.state.selected.uid in this.props.tokenMetadata && this.props.tokenMetadata[this.state.selected.uid].nft;
       if (this.state.selected.uid in this.props.tokensBalance) {
         availableBalance = this.props.tokensBalance[this.state.selected.uid].available;
       }
       return <span className="ml-3">({t`Balance available: `}{helpers.renderValue(availableBalance, isNFT)})</span>;
     }
 
+    const renderNFTHelper = () => {
+      return (
+        <div>
+          <small className="text-muted">This is an NFT token. The amount will be an integer number, without decimal places.</small>
+        </div>
+      );
+    }
+
     return (
       <div className='send-tokens-wrapper card'>
-        <div className="mb-3">
+        <div>
           <label><strong>{t`Token:`}</strong></label>
           {this.state.selected && renderSelectToken()}
           {this.state.selected && renderBalance()}
           {this.state.selectedTokens.length !== 1 ? <button type="button" className="text-danger remove-token-btn ml-3" onClick={(e) => this.props.removeToken(this.props.index)}>{t`Remove`}</button> : null}
         </div>
-        <div className="outputs-wrapper">
+        {isNFT && renderNFTHelper()}
+        <div className="outputs-wrapper mt-3">
           <label>Outputs</label>
           {renderOutputs()}
         </div>
--- a/src/components/SendTokensOne.js
+++ b/src/components/SendTokensOne.js
@@ -14,6 +14,7 @@ import { connect } from "react-redux";
 import _ from 'lodash';
 import hathorLib from '@hathor/wallet-lib';
 import wallet from '../utils/wallet';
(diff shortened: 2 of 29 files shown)


src/components/WalletBalance.js
```diff
@@ -33,13 +35,14 @@ class WalletBalance extends React.Component {
       locked: token.uid in this.props.tokensBalance ? this.props.tokensBalance[token.uid].locked : 0,
     }
 
-    const renderBalance = () => {
+    const isNFT = this.props.selectedToken in this.props.tokenMetadata && this.props.tokenMetadata[this.props.selectedToken].nft;
```
andreabadesso, line 38: Same here

src/utils/wallet.js
```diff
@@ -211,6 +214,34 @@ const wallet = {
...
+   * @param {Array} tokens Array of token uids
+   * @param {String} network Network name
+   *
+   * @memberof Wallet
+   * @inner
+   **/
+  async fetchTokensMetadata(tokens, network) {
+    const metadatas = {};
+
+    for (const token of tokens) {
```
andreabadesso, line 230: Do we have any reason for downloading metadata in series? 

Maybe we could parallelize this using Promise.all

andreabadesso (changes requested):

src/utils/wallet.js
pedroferreira1: As we discussed in private, it's better to keep the serial download for now.

src/components/TxData.js
```diff
@@ -291,7 +310,10 @@ class TxData extends React.Component {
           return t`Unknown authority`;
         }
       } else {
-        return hathorLib.helpers.prettyValue(output.value);
+        // if it's an NFT token we should show integer value
+        const uid = this.getUIDFromTokenData(hathorLib.wallet.getTokenIndex(output.token_data));
+        const isNFT = isTokenNFT(uid);
```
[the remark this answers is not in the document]
pedroferreira1: Refactor done, can you see again, please?

src/components/SendTokensOne.js
```diff
@@ -88,16 +90,17 @@ class SendTokensOne extends React.Component {
    * Iterate through inputs and outputs to add each information typed to the data object
    */
   getData = () => {
+    const isNFT = this.state.selected.uid in this.props.tokenMetadata && this.props.tokenMetadata[this.state.selected.uid].nft;
```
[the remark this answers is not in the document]
pedroferreira1: Yeah, makes sense. Used `_.get` and did the isTokenNFT refactor. Can you see again please?

andreabadesso (approved):

src/utils/tokens.js
```diff
@@ -31,6 +31,9 @@ const tokens = {
   addToken(uid, name, symbol) {
     const tokens = hathorLib.tokens.addToken(uid, name, symbol);
     store.dispatch(newTokens({tokens, uid: uid}));
+    const reduxState = store.getState();
+    const reduxWallet = reduxState.wallet;
+    wallet.fetchTokensMetadata([uid], reduxWallet.conn.network);
```
msbrogli, line 36: What if it fails? Do we automatically retry? How many times?

src/utils/wallet.js
```diff
@@ -211,6 +214,34 @@ const wallet = {
...
+   * So we fetch the tokens metadata and store on redux
+   *
+   * @param {Array} tokens Array of token uids
+   * @param {String} network Network name
+   *
+   * @memberof Wallet
+   * @inner
+   **/
+  async fetchTokensMetadata(tokens, network) {
+    const metadatas = {};
```
msbrogli, line 228: metadata is already plural.

src/utils/wallet.js
```diff
@@ -211,6 +214,34 @@ const wallet = {
...
+
+    for (const token of tokens) {
+      try {
+        const response = await metadataApi.getDag(token, network);
```
msbrogli, line 232: Shouldn't it be `getDagMetadata(...)`?

src/utils/wallet.js
```diff
@@ -211,6 +214,34 @@ const wallet = {
...
+        if (response.data && token in response.data) {
+          const tokenMeta = response.data[token];
+          metadatas[token] = tokenMeta;
+        }
+      } catch (e) {
+        // No need to do anything, the metadata for this token was not found
```
msbrogli, line 238: What if we get a 500 or 429 return code? Or the server times out?

src/utils/wallet.js
```diff
@@ -211,6 +214,34 @@ const wallet = {
...
+   * @memberof Wallet
+   * @inner
+   **/
+  async fetchTokensMetadata(tokens, network) {
+    const metadatas = {};
+
+    for (const token of tokens) {
+      try {
+        const response = await metadataApi.getDag(token, network);
+        if (response.data && token in response.data) {
```
msbrogli, line 233: What does it mean when we do not receive a token in the response? Can an unexpected error cause it somehow?

src/screens/NFTList.js
```diff
@@ -0,0 +1,125 @@
...
+import helpers from '../utils/helpers';
+import { connect } from "react-redux";
+
+
+const mapStateToProps = (state) => {
+  const registeredTokens = state.tokens;
+  const tokenMetadata = state.tokenMetadata || {};
+  const nftList = Object.values(tokenMetadata).filter((meta) => {
+    const hasNFTMedia = meta.nft_media && meta.nft_media.file;
+    const isRegistered = state.tokens.find((registeredToken) => registeredToken.uid === meta.id);
```
msbrogli, line 23: This is an O(n^2) operation. Should we index in O(n) and check in O(1), reducing the overall complexity to O(n)? I feel the code change is simple enough to do this refactor.

src/screens/NFTList.js
```diff
@@ -0,0 +1,125 @@
...
+    return hasNFTMedia && isRegistered;
+  });
+
+  const nftData = nftList.map((meta) => {
+    const tokenConfig = state.tokens.find((registeredToken) => registeredToken.uid === meta.id);
```
msbrogli, line 28: Same thing here: This algorithm is O(n^2) and can be easily refactored to an O(n).

src/screens/NFTList.js
```diff
@@ -0,0 +1,125 @@
...
+   * @return {Number} Total number of pages of the list
+   */
+  getTotalPages = () => {
+    return Math.ceil(this.props.nftData.length / NFT_LIST_PER_PAGE);
+  }
+
+  /**
+   * Event received from pagination component after a page button in clicked
+   *
+   * @param data {Object} Data with clicked page {'selected'}
```
msbrogli, line 74: Add a note that it is 0-indexed (while `this.state.page` is 1-indexed). Maybe we should add a note about `this.state.page` right above its definition as well.

msbrogli (changes requested):

src/components/SendTokensOne.js
```diff
@@ -84,6 +86,13 @@ class SendTokensOne extends React.Component {
     this.setState({ inputsCount: newCount });
   }
 
+  /**
+   * Use helper method to check if token is an NFT
+   */
+  isNFT = () => {
+    return helpers.isTokenNFT(_.get(this.state, 'selected.uid'), this.props.tokenMetadata);
```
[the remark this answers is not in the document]
pedroferreira1: The idea of using `_.get` is because it already checks if the key exists in the object. So if I want to get `this.state.selected.uid` but I don't know if `this.state.selected` exists I need to do `this.state.selected && this.state.selected.uid`. With `_.get` I can safely do that and it handles the existence for me.

If `selected` is null it will return `false`, i.e. will continue the same old behaviour of the screen

src/components/SendTokensOne.js
```diff
@@ -219,18 +228,27 @@ class SendTokensOne extends React.Component {
...
         availableBalance = this.props.tokensBalance[this.state.selected.uid].available;
       }
-      return <span className="ml-3">({t`Balance available: `}{hathorLib.helpers.prettyValue(availableBalance)})</span>;
+      return <span className="ml-3">({t`Balance available: `}{helpers.renderValue(availableBalance, this.isNFT())})</span>;
+    }
+
+    const renderNFTHelper = () => {
+      return (
+        <div>
+          <small className="text-muted">This is an NFT token. The amount will be an integer number, without decimal places.</small>
```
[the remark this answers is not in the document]
pedroferreira1: Changed, thanks

src/components/TxData.js
```diff
@@ -599,7 +638,7 @@ class TxData extends React.Component {
           {renderBalance()}
           <div className="d-flex flex-row align-items-start mt-3 mb-3">
             <div className="d-flex flex-column align-items-start common-div bordered-wrapper mr-3">
-              <div><label>{t`Type:`}</label> {hathorLib.helpers.getTxType(this.props.transaction)}</div>
+              <div><label>Type:</label> {hathorLib.helpers.getTxType(this.props.transaction)} {isNFTCreation() && '(NFT)'}</div>
```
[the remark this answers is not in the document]
pedroferreira1: Added back, thanks. I have no idea why I removed it

src/components/tokens/TokenAction.js
```diff
@@ -125,6 +133,10 @@ class TokenAction extends React.Component {
   }
 
   render() {
+    if (!this.props.metadataLoaded) {
+      return <p>{t`Before executing token actions you must wait until the full data of your wallet is loaded.`}</p>
```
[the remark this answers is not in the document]
pedroferreira1: Yeah, I agree. It's better. Changed it

src/utils/wallet.js
pedroferreira1: Yeah, I had this change locally already. Thanks

src/components/NFTListElement.js
```diff
@@ -0,0 +1,103 @@
...
+  /**
+   * Method called when user clicked on 'See on explorer'
+   * Go to explorer token detail page
+   *
+   * @param {Object} e Event for the click
+   * @param {String} token Token uid
+   */
+  goToTokenDetail = (e, token) => {
+    e.preventDefault();
+    const url = path.join(helpers.getExplorerURL(), `token_detail/${token}`);
```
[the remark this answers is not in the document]
pedroferreira1: Do you mean to create a method like this and use it? (I've done that, just confirming if that's what you said)

```
  getFullExplorerURL(urlPath) {
    return path.join(this.getExplorerURL(), urlPath);
  },
```

src/components/TxData.js
```diff
@@ -324,15 +345,20 @@ class TxData extends React.Component {
       });
     }
 
-    const renderDecodedScript = (decoded) => {
-      switch (decoded.type) {
+    const renderDecodedScript = (output) => {
+      switch (output.decoded.type) {
```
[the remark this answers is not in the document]
pedroferreira1: Right now we never return null from the API, we return `decoded: {}` but I agree it's better to add this protection. Just did.

src/utils/helpers.js
```diff
@@ -69,6 +69,55 @@ const helpers = {
...
+   * If token is an NFT checking metadata
+   *
+   * @param {string} uid Token uid
+   * @param {Object} metadatas Metadatas of registered tokens
+   *
+   * @return {boolean} if token is an NFT
+   * @memberof Helpers
+   * @inner
+   */
+  isTokenNFT(uid, metadatas) {
```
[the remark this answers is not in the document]
pedroferreira1: Renamed to `metadata_per_token`, ok? I thought it was better

src/utils/wallet.js
pedroferreira1: Changed to `metadataPerToken`

src/screens/NFTList.js
pedroferreira1: Done.

src/screens/NFTList.js
pedroferreira1: Yeah, makes sense. Refactor done here: https://github.com/HathorNetwork/hathor-wallet/pull/228/commits/96bb6e497a99d75557e1b89587c624677e15dbd5

src/screens/NFTList.js
pedroferreira1: Actually this was not needed anymore, given that we are now only storing metadata of registered tokens. Removed this part here: https://github.com/HathorNetwork/hathor-wallet/pull/228/commits/96bb6e497a99d75557e1b89587c624677e15dbd5

src/utils/wallet.js
r4mmer, line 232: Use `Promise.all` so it loads faster.
Do you think it's better to create a get method to a list of tokens?

src/utils/wallet.js
r4mmer: We could make the api return 200 with empty body instead of 404, so this catch can be only for unexpected errors

src/components/tokens/TokenMint.js
```diff
@@ -110,6 +121,8 @@ class TokenMint extends React.Component {
   }
 
   render() {
+    const isNFT = this.props.token.uid in this.props.tokenMetadata && this.props.tokenMetadata[this.props.token.uid].nft;
```
r4mmer, line 124: Is this being used?

src/utils/tokens.js
pedroferreira1: I've refactored this part of code to have an error handling and auto retry. Can you see if it's good? https://github.com/HathorNetwork/hathor-wallet/pull/228/commits/9e150da7e7f5923df00911378430609f9345f9aa

src/utils/wallet.js
pedroferreira1: I've refactored this part of code to have an error handling and auto retry. Can you see if it's good? https://github.com/HathorNetwork/hathor-wallet/pull/228/commits/9e150da7e7f5923df00911378430609f9345f9aa

src/utils/wallet.js
pedroferreira1: I've refactored this part of code to have an error handling and auto retry. Can you see if it's good? https://github.com/HathorNetwork/hathor-wallet/pull/228/commits/9e150da7e7f5923df00911378430609f9345f9aa

src/components/tokens/TokenMint.js
pedroferreira1: It's not. Thanks!

src/utils/wallet.js
pedroferreira1: As we discussed privately it's better not to do it right now

src/components/NFTListElement.js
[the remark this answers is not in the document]
msbrogli: Exactly! Thanks!

src/utils/helpers.js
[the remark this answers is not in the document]
msbrogli: Sure! Well done. Thanks!

src/components/SendTokensOne.js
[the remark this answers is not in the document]
msbrogli: What does `isTokenNFT(...)` expect for the first parameter? Is a `false` an expected type? If yes, I'm good to resolve this conversation.

src/components/TxData.js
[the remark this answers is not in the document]
msbrogli: Well done. Thanks!

src/screens/NFTList.js
msbrogli: Ok. Thanks!

src/screens/NFTList.js
msbrogli: Well done. Thanks!

src/utils/wallet.js
```diff
@@ -211,6 +215,74 @@ const wallet = {
...
+        }, DOWNLOAD_METADATA_RETRY_INTERVAL);
+        // Will break the for
+        didBreak = true;
+        return true;
+      }
+    }
+
+    for (const [index, token] of tokens.entries()) {
+      try {
+        const response = await metadataApi.getDagMetadata(token, network);
```
msbrogli, line 251: I feel that this retry logic should be inside `getDagMetadata(...)` which might receive an optional parameter `maxNumberOfRetries`. It should even handle the issue of getting a response with no data.

src/utils/wallet.js
msbrogli: Well done. Thanks!

src/utils/wallet.js
msbrogli: You can move the current implementation of `getDagMetadata(...)` to `_getRawDagMetadata(...)` so we can use it for a single fetch (even though I feel we won't).

src/utils/wallet.js
```diff
@@ -211,6 +215,74 @@ const wallet = {
...
+  async fetchTokensMetadata(tokens, network, downloadRetry = 0) {
+    const metadataPerToken = {};
+    let didBreak = false;
+
+    const handleRetry = (index) => {
+      if (downloadRetry === METADATA_RETRY_LIMIT) {
+        // Should we throw an error and prevent wallet usage only because metadata was not loaded correctly?
+        // Right now I'm continuing and will download as many as we can
+        // Won't break the for
+        return false;
```
msbrogli, line 238: Should we report this problem to Sentry? Maybe we should log it as well since one can run the wallet with debugging tools.

msbrogli, line 238: Should we mark the error in the redux global state? We can use it to show the error on the screens.

msbrogli (approved):

src/components/SendTokensOne.js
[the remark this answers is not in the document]
pedroferreira1: I think I was not clear with the words. If `selected` is `null`, then `isTokenNFT` will return `false`.

If `selected` is `null`,  `_.get` will return `undefined`, then the first thing we do is `uid in metadata` in this case `undefined in metadata`, then returns `false`.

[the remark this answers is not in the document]
msbrogli: I got it. `_.get(...)` will return `undefined`. And `isTokenNFT(undefined, ...)` will return false. Ok, then!

src/utils/wallet.js
msbrogli: After a second refactor, this code moved to the hathor-wallet-lib.

src/utils/tokens.js
msbrogli: After a second refactor, this logic moved to the hathor-wallet-lib.

src/utils/wallet.js
msbrogli: Done! Thanks!

src/reducers/index.js
```diff
@@ -45,6 +45,12 @@ const initialState = {
   // Height of the best chain of the network arrived from ws data
   height: 0,
   wallet: null,
+  // Metadata of tokens
+  tokenMetadata: {},
+  // Token list of uids that had errors when loading metadata
+  tokenMetadataErrors: [],
```
msbrogli, line 51: Should it be a set or a dict with uids as keys and more information (e.g. latest retry time) as values?

msbrogli, line 51: It seems we are not using this piece of information yet, right?

msbrogli (approved):

src/utils/wallet.js
pedroferreira1: This is an old code but I don't think report to Sentry is good. Request may fail for many client reasons. I'm logging in the wallet so we can know when using a debug tool

pedroferreira1: Added the redux state with the errors but we've decided not to use it for now

src/reducers/index.js
pedroferreira1: As we discussed privately we won't use this state for now, so I will keep the array and when we decide to show those errors we can improve them.

pedroferreira1: As we discussed privately we won't use this state for now

PR #257, opened 2021-12-13
fixing losing of redux's storage info when browser reload or close app

roremeol:
### Acceptance Criteria
- Fixing losing redux's storage when refresh browser or closing the app

### Security Checklist
- [ x ] Make sure you do not include new dependencies in the project unless strictly necessary and do not include dev-dependencies as production ones. More dependencies increase the possibility of one of them being hijacked and affecting us.

### DIFF
--- a/src/components/TokenHistory.js
+++ b/src/components/TokenHistory.js
@@ -234,7 +234,10 @@ class TokenHistory extends React.Component {
     }
 
     const renderHistoryData = () => {
-      const keys = hathorLib.wallet.getWalletData().keys;
+      const data = hathorLib.wallet.getWalletData()
+      if(data==null)
+        return hathorLib.wallet.lock();
+      const keys = data.keys;
       const isNFT = helpers.isTokenNFT(get(this.props, 'selectedToken'), this.props.tokenMetadata);
       return this.state.transactions.map((tx, idx) => {
         let statusElement = '';


pedroferreira1:
@roremeol can you please change the base branch to `dev`?

And I will just request some code refactoring to be in the same style we have:

```
   const data = hathorLib.wallet.getWalletData();
   if (data === null) {
        return hathorLib.wallet.lock();
   }
   const keys = data.keys;
```

Besides that, I still need to test the code but seems to me we would need to add a 
`this.props.history.push('/locked/');` after the method to lock, doesn't it?

roremeol:
> @roremeol can you please change the base branch to `dev`?
> 
> 
> 
> And I will just request some code refactoring to be in the same style we have:
> 
> 
> 
> ```
> 
>    const data = hathorLib.wallet.getWalletData();
> 
>    if (data === null) {
> 
>         return hathorLib.wallet.lock();
> 
>    }
> 
>    const keys = data.keys;
> 
> ```
> 
> 
> 
> Besides that, I still need to test the code but seems to me we would need to add a 
> 
> `this.props.history.push('/locked/');` after the method to lock, doesn't it?

Pedro, during my tests, this code gotten work.

I think it is happening because the security of the user. You doesn't store some user info ( pin, ... ) and wen do you reload de browser this storages is lost and the wallet stops to work. When you lock the wallet, the user have to insert the pin again and the wallet recoverers it self.

I think your sugest works too.

PR #259, opened 2021-12-21
feat: new command handlers

r4mmer:
### Acceptance Criteria
- Implement command methods for new ledger commands

### Security Checklist

### DIFF
--- a/public/ledger.js
+++ b/public/ledger.js
@@ -387,7 +387,7 @@ class Ledger {
    *   List of data with the format:
    *     version + uid + len(symbol) + symbol + len(name) + name + len(signature) + signature
    *
-   * @return {Promise} Promise resolved when signature is received
+   * @return {Promise} Promise resolved when all requests finish
    */
   sendTokens = async (tokens) => {
     const values = [];
@@ -413,7 +413,7 @@ class Ledger {
    * @param {Object} data Data to send to Ledger
    *   version + uid + len(symbol) + symbol + len(name) + name + len(signature) + signature
    *
-   * @return {Promise} Promise resolved when signature is received
+   * @return {Promise} Promise resolved when Ledger request finishes
    */
   verifyTokenSignature = async (data) => {
     try {
@@ -427,15 +427,12 @@ class Ledger {
   /**
    * Reset token signatures
    *
-   * @param {Object} data Data to send to Ledger
-   *   version + uid + len(symbol) + symbol + len(name) + name + len(signature) + signature
-   *
-   * @return {Promise} Promise resolved when signature is received
+   * @return {Promise} Promise resolved when signatures are reset on Ledger
    */
   resetTokenSignatures = async () => {
     try {
       const transport = await this.getTransport();
-      await this.sendToLedgerOrQueue(transport, this.commands.RESET_TOKEN_SIGNATURES, 0, 0, data);
+      await this.sendToLedgerOrQueue(transport, this.commands.RESET_TOKEN_SIGNATURES, 0, 0);
     } catch (e) {
       throw Ledger.parseLedgerError(e);
     }
--- a/public/ledger.js
+++ b/public/ledger.js
@@ -13,6 +13,7 @@ const constants = require('./constants');
 
 const maxLedgerBuffer = 255;
 const ledgerCLA = 0xe0;
+const ledgerOK = Buffer.from('9000', 'hex');
 
 class Ledger {
   
@@ -32,6 +33,7 @@ class Ledger {
       0xb007:  "SW_BAD_STATE",
       0xb008:  "SW_SIGNATURE_FAIL",
       0xb009:  "SW_INVALID_TX",
+      0xb00a:  "SW_INVALID_SIGNATURE",
     };
     if (e.name && e.name === 'TransportStatusError') {
       switch (e.statusCode) {
@@ -51,6 +53,7 @@ class Ledger {
         case 0xb007:  // SW_BAD_STATE
         case 0xb008:  // SW_SIGNATURE_FAIL
         case 0xb009:  // SW_INVALID_TX
+        case 0xb00a:  // SW_INVALID_SIGNATURE
         default:
           console.log("LedgerError: ", errorMap[e.statusCode] || 'UNKNOWN_ERROR');
           return {status: e.statusCode, message: 'Error communicating with Ledger'};
@@ -139,6 +142,10 @@ class Ledger {
       'ADDRESS': 0x04,
       'PUBLIC_KEY_DATA': 0x05,
       'SEND_TX': 0x06,
+      'SIGN_TOKEN': 0x07,
+      'SEND_TOKEN': 0x08,
+      'VERIFY_TOKEN_SIGNATURE': 0x09,
+      'RESET_TOKEN_SIGNATURES': 0x0a,
     }
 
     // Queue of commands to send to ledger so we don't send them in paralel
@@ -217,7 +224,11 @@ class Ledger {
           "getPublicKeyData",
           "checkAddress",
           "sendTx",
-          "getSignatures"
+          "getSignatures",
+          "signToken",
+          "sendTokens",
+          "verifyTokenSignature",
+          "resetTokenSignatures",
         ];
         this.transport.decorateAppAPIMethods(this, this.methods, "HTR");
         resolve(transport);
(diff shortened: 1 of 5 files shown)


src/utils/ledger.js
```diff
@@ -125,6 +157,61 @@ if (IPC_RENDERER) {
...
+     * Sign token info on ledger
+     *
+     * @param {Object} token
+     *  with uid (hex), symbol and name
+     *
+     * @memberof Ledger
+     * @inner
+     */
+    signToken(token) {
+      const data = Buffer.concat(serializeTokenInfo(token, false));
```
luislhl, line 171: I've seen in https://github.com/HathorNetwork/hathor-ledger-app/pull/38 that we will have limits on the token name and symbol lengths.

Should we add a validation for this here before sending the token to Ledger?

Or this validation already exists somewhere? (like when minting the token)

src/utils/ledger.js
```diff
@@ -125,6 +157,61 @@ if (IPC_RENDERER) {
...
+     * Send tokens to ledger to create context for sign_tx
+     *
+     * @param {Object} tokens
+     *  each with uid (hex), symbol, name and signature
+     *
+     * @memberof Ledger
+     * @inner
+     */
+    sendTokens(tokens) {
+      const data = tokens.map(t => Buffer.concat(serializeTokenInfo(t, true)));
```
luislhl, line 186: Should we add a validation here on the number of tokens to avoid hitting the `TX_MAX_TOKENS` in the Ledger App?

src/utils/ledger.js
```diff
@@ -125,6 +157,61 @@ if (IPC_RENDERER) {
...
+    /**
+     * Sign token info on ledger
+     *
+     * @param {Object} token
+     *  with uid (hex), symbol and name
+     *
+     * @memberof Ledger
+     * @inner
+     */
+    signToken(token) {
```
luislhl, line 170: Question: The expected flow here would be to:
1. sign a number of tokens
2. send them to ledger
3. verify their signatures
4. send a tx using the tokens (which is not implemented here yet, right?)

Is that right?

In which cases do we use the `resetTokenSignatures`?

src/utils/ledger.js
r4mmer: There is a PR for the sign token modal with this validation, it's better to leave it on the component context since it can better show where/what went wrong.

src/utils/ledger.js
r4mmer: This validation is done on another PR that calls this method, actually the user won't be able to add more than 10 tokens to the transaction if it's a hardware wallet.

src/utils/ledger.js
r4mmer: The flow to sign a tx will be:

1. Individually sign custom tokens that will be used
2. When user clicks to send transaction
3. Send custom tokens if they exist (if not, go to 5)
4. If they fail, show which failed
5. Send the `signTx` command 

The `sendToken` actually verifies the token signature and prepares the context for the `signTx`. So we don't need to call verify here. It will only be used on startup to check the signatures on storage are still valid.

src/utils/ledger.js
luislhl: I think adding a validation here is something you should still consider.

Since it's a utils method to interact with ledger, we could use it in other places in the future besides the component, and we should make sure the validation is done every time.

src/utils/ledger.js
luislhl: Same thing I commented in the other comment.

Maybe we want to make sure there is a validation here as well, in case someone uses this method somewhere else besides your component.

src/utils/tokens.js
```diff
@@ -118,6 +118,33 @@ const tokens = {
...
+   * Returns the token signatures on storage
+   * This is only used for hw wallets.
+   *
+   * @memberof Tokens
+   * @inner
+   */
+  getTokenSignatures() {
+    const tokenSignatures = hathorLib.storage.getItem('wallet:token:signatures');
+    if (!tokenSignatures) return {};
+    // if (tokenSignatures === null) return {};
```
pedroferreira1, line 132: Remove?

src/utils/tokens.js
```diff
@@ -118,6 +118,33 @@ const tokens = {
   getNFTFee() {
     return 1;
   },
+
+  /**
+   * Returns the token signatures on storage
+   * This is only used for hw wallets.
+   *
```
pedroferreira1, line 125: Missing return type in the docstring

src/utils/tokens.js
```diff
@@ -118,6 +118,33 @@ const tokens = {
...
+    const tokenSignatures = hathorLib.storage.getItem('wallet:token:signatures');
+    if (!tokenSignatures) return {};
+    // if (tokenSignatures === null) return {};
+    return tokenSignatures;
+  },
+
+  /**
+   * Add a token signature to storage, overwriting if exists
+   * This is only used for hw wallets.
+   *
```
pedroferreira1, line 139: Missing params and return types in the docstring

public/ledger.js
```diff
@@ -350,6 +361,85 @@ class Ledger {
...
+
+  /**
+   * Get signature for token on ledger
+   *
+   * @param {Object} data Data to send to Ledger
+   *   version + uid + len(symbol) + symbol + len(name) + name
+   *
+   * @return {Promise} Promise resolved when signature is received
+   */
+  signToken = async (data) => {
```
pedroferreira1, line 373: Should we change to `getTokenSignature`? I think it's more explicit what the method does, just like you wrote in the docstring. If we change this, then we should change the command as well

public/ledger.js
```diff
@@ -350,6 +361,85 @@ class Ledger {
...
+
+  /**
+   * Reset token signatures
+   *
+   * @param {Object} data Data to send to Ledger
+   *   version + uid + len(symbol) + symbol + len(name) + name + len(signature) + signature
+   *
+   * @return {Promise} Promise resolved when signature is received
+   */
+  resetTokenSignatures = async () => {
```
pedroferreira1, line 435: Where is the `data` parameter? Is it working?

pedroferreira1, line 435: We can reset only one signature, right? Should we change the name to singular? `resetTokenSignature`

src/utils/ledger.js
```diff
@@ -34,6 +34,38 @@ const formatPathData = (index) => {
   return buffer;
 };
 
+const intToByte = (number) => {
```
pedroferreira1, line 37: I guess you have a method exactly like this in the lib at `hathorLib.helpersUtils.intToBytes(value, 1)`, where 1 is the number of bytes.

You can use both but using that method is more generic if we need to use more of int to bytes in the future.

src/utils/ledger.js
```diff
@@ -34,6 +34,38 @@ const formatPathData = (index) => {
...
+
+export const serializeTokenInfo = (token, hasSignature) => {
+  // version + uid + len(symbol) + symbol + len(name) + name
+  const uidBytes = Buffer.from(token.uid, "hex");
+  const symbolBytes = Buffer.from(token.symbol, "utf8");
+  const nameBytes = Buffer.from(token.name, "utf8");
+  const arr = [];
+
+  // 0: token version = 1 (always)
+  arr.push(intToByte(1));
```
pedroferreira1, line 51: Maybe we should create this in the constants file

src/utils/ledger.js
```diff
@@ -34,6 +34,38 @@ const formatPathData = (index) => {
   return buffer;
 };
 
+const intToByte = (number) => {
+  const buf = Buffer.alloc(1);
+  buf[0] = number;
+  return buf;
+}
+
+export const serializeTokenInfo = (token, hasSignature) => {
```
pedroferreira1, line 43: Missing docstring

public/ledger.js
```diff
@@ -350,6 +361,85 @@ class Ledger {
...
+  }
+
+  /**
+   * Send tokens and signatures to ledger before a tx
+   *
+   * @param {Object} tokens List of data to send to ledger
+   *   List of data with the format:
+   *     version + uid + len(symbol) + symbol + len(name) + name + len(signature) + signature
+   *
+   * @return {Promise} Promise resolved when signature is received
```
pedroferreira1, line 390: Is this correct?

public/ledger.js
```diff
@@ -350,6 +361,85 @@ class Ledger {
...
+    }
+  }
+
+  /**
+   * Verify a token and signature with ledger
+   *
+   * @param {Object} data Data to send to Ledger
+   *   version + uid + len(symbol) + symbol + len(name) + name + len(signature) + signature
+   *
+   * @return {Promise} Promise resolved when signature is received
```
pedroferreira1, line 416: Is this correct?

public/ledger.js
```diff
@@ -350,6 +361,85 @@ class Ledger {
...
+    }
+  }
+
+  /**
+   * Reset token signatures
+   *
+   * @param {Object} data Data to send to Ledger
+   *   version + uid + len(symbol) + symbol + len(name) + name + len(signature) + signature
+   *
+   * @return {Promise} Promise resolved when signature is received
```
pedroferreira1, line 433: Is this correct?

pedroferreira1 (changes requested):

public/ledger.js
r4mmer: Yes

r4mmer: We can't reset a single signature, only all signatures

src/utils/ledger.js
r4mmer: Started using it

public/ledger.js
r4mmer: Fixed

public/ledger.js
r4mmer: Fixed

src/utils/ledger.js
r4mmer: Added

src/utils/ledger.js
r4mmer: Added

public/ledger.js
r4mmer: Fixed

public/ledger.js
r4mmer: Maybe, but I think it's good to maintain consistency with the method name and the command name

src/utils/tokens.js
r4mmer: added

src/utils/tokens.js
r4mmer: Done

src/utils/tokens.js
r4mmer: Done

public/ledger.js
pedroferreira1: I agree. And to change the command name we must change the ledger app code?

public/ledger.js
pedroferreira1: Yeah, I saw in the other PR but this was the first one

pedroferreira1: I was a bit confused with the data parameter, it's good now.

src/constants.js
```diff
@@ -193,3 +193,8 @@ export const NFT_GUIDE_URL = 'https://hathor.network/nft-guide';
  * How many token meta we are downloading concurrently
  */
 export const METADATA_CONCURRENT_DOWNLOAD = 10;
+
+/**
+ * Token version
+ */
+export const TOKEN_VERSION = 1;
```
pedroferreira1, line 200: I think the parameter name should be more explicit, e.g. `LEDGER_TOKEN_SIGNATURE_VERSION` and the docstring should describe in more detail for what we use it.

src/utils/ledger.js
```diff
@@ -34,6 +34,40 @@ const formatPathData = (index) => {
   return buffer;
 };
 
+
+/**
+ * Serialize token information to an array of Buffer
+ *
```
pedroferreira1, line 40: Missing parameter types

pedroferreira1 (changes requested):

public/ledger.js
r4mmer: Yes.

src/utils/ledger.js
r4mmer: Done.

src/constants.js
r4mmer: Changed to `LEDGER_TOKEN_VERSION` since it's used on other commands other than signature.

src/utils/ledger.js
r4mmer: I think you're right that we should have a "centralized" validation, but I don't think this is the place to do this.

I have an idea to remove the `public/ledger.js` module into a separate lib, we could later do all validation on this module and have the callers catch validation errors. This seems better than having duplicated validations here.

luislhl: Ok, I think that makes sense.

But I just think you should create an issue explaining this idea, if there isn't one already.

luislhl (approved):

PR #264, opened 2021-12-30
feat(ledger) sign tx with custom tokens

r4mmer:
### Acceptance Criteria
- Send custom tokens to ledger before signing a tx

### Security Checklist


src/screens/SendTokens.js
```diff
@@ -289,14 +357,19 @@ class SendTokens extends React.Component {
...
-   * Checks if still have a known token available that is not selected yet  
+   * Executed when user clicks to add another token to this transaction
+   * Checks if still have a known token available that is not selected yet
    * Create a new child reference with this new token
    */
   addAnotherToken = () => {
     if (hathorLib.wallet.isHardwareWallet()) {
-      $('#notSupported').modal('show');
-      return;
+      if (this.state.txTokens.length === 10) {
```
luislhl, line 366: Maybe move `10` to a constant

src/screens/SendTokens.js
```diff
@@ -393,14 +471,37 @@ class SendTokens extends React.Component {
       }
     }
 
+    const renderSignTokensAlertBody = () => {
```
luislhl, line 474: It seems this is not being used

src/components/SendTokensOne.js
```diff
@@ -156,12 +156,27 @@ class SendTokensOne extends React.Component {
...
       input.tx_id = input.txId;
     }
+    for (let output of data.outputs) {
+      if (hathorLib.tokens.isHathorToken(this.state.selected.uid)) {
+        output.tokenData = 0;
+        continue;
+      }
+      // adding tokenData information
+      console.log('allTokens', this.state.selectedTokens.filter(t => !hathorLib.tokens.isHathorToken(t.uid)));
+      output.tokenData = 1 + this.state.selectedTokens
```
luislhl, line 166: Where will this `tokenData` field be used?

src/screens/SendTokens.js
```diff
@@ -203,6 +237,39 @@ class SendTokens extends React.Component {
     this.setState({ errorMessage: message, ledgerStep: 0 });
   }
 
+  beforeSendLedger = () => {
```
luislhl, line 240: Could you add a docstring on this method?

luislhl:
At this point are you already able to record a video of the new features in use?

src/screens/SendTokens.js
```diff
@@ -393,14 +471,37 @@ class SendTokens extends React.Component {
...
+
+      return <ul>{rows}</ul>
+    }
+
+    const renderLedgerModals = () => {
+      if (hathorLib.wallet.isSoftwareWallet()) return null;
+
+      return (
+        <div>
+          <ModalAlertNotSupported />
```
pedroferreira1, line 490: I think we don't need this modal anymore, right?

src/components/SendTokensOne.js
```diff
@@ -156,12 +156,27 @@ class SendTokensOne extends React.Component {
     for (let input of data.inputs) {
       input.tx_id = input.txId;
     }
+    for (let output of data.outputs) {
```
pedroferreira1, line 159: This method here was created in the past to speed up hardware wallet integration and is using some old lib methods that will be deprecated soon

I think this is a nice opportunity to refactor this hardware wallet part to use the same methods of the software wallet, instead of continuing adding custom things here just to work specifically for hardware wallet. What do you think?

src/screens/SendTokens.js
r4mmer: Removed

src/components/SendTokensOne.js
```diff
@@ -156,12 +156,27 @@ class SendTokensOne extends React.Component {
...
+      if (hathorLib.tokens.isHathorToken(this.state.selected.uid)) {
+        output.tokenData = 0;
+        continue;
+      }
+      // adding tokenData information
+      console.log('allTokens', this.state.selectedTokens.filter(t => !hathorLib.tokens.isHathorToken(t.uid)));
```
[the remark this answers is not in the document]
r4mmer: Removed

src/screens/SendTokens.js
r4mmer: Removed

src/components/SendTokensOne.js
```diff
@@ -156,12 +156,27 @@ class SendTokensOne extends React.Component {
...
+      output.tokenData = 1 + this.state.selectedTokens
+                              .filter(t => !hathorLib.tokens.isHathorToken(t.uid))
+                              .findIndex((t) => t.uid === this.state.selected.uid);
+      console.log('tokenData', output.tokenData);
```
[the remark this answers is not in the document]
r4mmer: Removed

src/components/SendTokensOne.js
r4mmer: Created a [PR](https://github.com/HathorNetwork/hathor-wallet-lib/pull/306) on the wallet-lib to make the necessary changes for this.

src/screens/SendTokens.js
```diff
@@ -126,25 +154,20 @@ class SendTokens extends React.Component {
...
       if (!dataOne) return;
-      if (hathorLib.wallet.isHardwareWallet()) {
-        dataOne = instance.validateInputsAndOutputs(dataOne);
-        if (!dataOne) return;
-        // currently we only support HTR
-        data['tokens'] = [];
-      }
       data['inputs'] = [...data['inputs'], ...dataOne['inputs']];
       data['outputs'] = [...data['outputs'], ...dataOne['outputs']];
+      data['tokens'] = [...data['tokens'], ...dataOne['tokens']];
```
pedroferreira1, line 170: `getData` method does not add any key `'tokens'` to the returned object. I don't think we should have this line here, the tokens array is handled in the lib with the token data.

src/screens/SendTokens.js
```diff
@@ -203,33 +215,73 @@ class SendTokens extends React.Component {
...
+    // get all custom tokens, and include the signatures
+    const tokenSignatures = tokens.getTokenSignatures();
+    const missingSigs = txTokens.filter(t => !tokenSignatures.hasOwnProperty(t.uid));
+    if (missingSigs.length !== 0) {
+      // there are tokens without signatures, missingSigs
+      // set tittle and content
+      this.ledgerAlertModalTitle = t`Unverified custom tokens`
+      this.ledgerAlertModalBody = this.renderAlertTokenList(missingSigs);
+      $("#ledgerAlertModal").modal('show');
+      return;
```
pedroferreira1, line 243: If the user is trying to send a tx and a custom token does not have a signature, I thought we would give the possibility for the user to sign it while sending. Isn't it the design?

src/screens/SendTokens.js
```diff
@@ -203,33 +215,73 @@ class SendTokens extends React.Component {
...
     // Complete data with default values
-    hathorLib.transaction.completeTx(data);
+    hathorLib.transaction.completeTx(this.data);
+
+    this.sendTransaction = hathorLib.SendTransaction({ outputs: this.data.outputs, inputs: this.data.inputs, network: this.props.wallet.getNetworkObject() });
+    this.data = this.sendTransaction.prepareTxData();
 
     // Find if tx has change output
-    const changeIndex = data.outputs.findIndex((v) => v.isChange === true);
+    const changeIndex = this.data.outputs.findIndex((v) => v.isChange === true);
```
pedroferreira1, line 270: With multiple tokens we can have more than one change output, one for each token. How are handling this in ledger? As far as I can see, `ledger.sendTx(this.data, changeIndex, changeKeyIndex);` method allows for only one change output, is that correct? Do we need to add this support on ledger as well?

src/screens/SendTokens.js
```diff
@@ -289,14 +341,19 @@ class SendTokens extends React.Component {
...
    * Create a new child reference with this new token
    */
   addAnotherToken = () => {
     if (hathorLib.wallet.isHardwareWallet()) {
-      $('#notSupported').modal('show');
-      return;
+      if (this.state.txTokens.filter(t => !hathorLib.tokens.isHathorToken(t.uid)).length === LEDGER_TX_CUSTOM_TOKEN_LIMIT) {
+        // limit is 10 custom tokens per tx
+        this.ledgerAlertModalTitle = t`Token limit reached`
+        this.ledgerAlertModalBody = <p>{t`Ledger has a limit of 10 different tokens per transaction.`}</p>
```
pedroferreira1, line 353: This message number 10 should use the constant var

pedroferreira1 (changes requested):

src/screens/SendTokens.js
r4mmer: Yes, this is a problem we have to fix.
There was no way to keep backwards compatibility when adding multiple change outputs.
We need a separate project to update the "sign tx" method on ledger.

src/screens/SendTokens.js
r4mmer: Changed

src/screens/SendTokens.js
r4mmer: Removed

src/screens/SendTokens.js
pedroferreira1: We need to discuss this, even create another design if needed. I don't think we should merge this PR like this because the implementation has bugs in this part.

src/screens/SendTokens.js
```diff
@@ -289,14 +345,24 @@ class SendTokens extends React.Component {
...
+   * Checks if still have a known token available that is not selected yet
    * Create a new child reference with this new token
    */
   addAnotherToken = () => {
     if (hathorLib.wallet.isHardwareWallet()) {
-      $('#notSupported').modal('show');
-      return;
+      if (!version.isLedgerCustomTokenAllowed()) {
+        // Custom token not allowed for this Ledger version
+        $('#notSupported').modal('show');
```
luislhl, line 356: This modal will show the following message: `Unfortunately this feature is currently not supported when using a hardware wallet. [...]`

Shouldn't we show a different message in this case? Saying that the Ledger version is not supported, instead of saying that hardware wallet is not supported.

src/components/SendTokensOne.js
```diff
@@ -216,8 +193,8 @@ class SendTokensOne extends React.Component {
     const renderSelectToken = () => {
       return (
         <select className="ml-3" value={this.state.selected.uid} onChange={this.changeSelect}
-          title={hathorLib.wallet.isHardwareWallet() ? t`This feature is disabled for hardware wallet` : t`Select token`} 
-          disabled={hathorLib.wallet.isHardwareWallet() ? true : null}>
+          title={version.isLedgerCustomTokenAllowed() ? t`Select token` : t`This feature is disabled for hardware wallet`}
```
luislhl, line 196: Is this message `This feature is disabled for hardware wallet` still accurate? 

Shouldn't we change it to say it is disabled for the current ledger version?

luislhl (approved):
I've tested this with Ledger and things are working fine.

The only problem I had was that my token signatures were invalidated when I installed a new version of the app in Ledget, and my tx was failing to send. The reason for this and the necessary fix were not clear to the user.

public/ledger.js
```diff
@@ -383,28 +383,27 @@ class Ledger {
...
-          values.push(data);
-        }
+    const transport = await this.getTransport();
+    let i = 0;
+    for (let [uid, data] of Object.entries(tokenMap)) {
+      try {
+        await this.sendToLedgerOrQueue(transport, this.commands.SEND_TOKEN, i, 0, data);
+      } catch (e) {
+        // only return failures
+        values.push(uid);
```
pedroferreira1, line 401: Ledger now is throwing an error instead of returning the not ok value like we had before?
`if (ledgerOK.compare(value) !== 0) {`

public/ledger.js
```diff
@@ -383,28 +383,27 @@ class Ledger {
   /**
    * Send tokens and signatures to ledger before a tx
    *
-   * @param {Object} tokens List of data to send to ledger
-   *   List of data with the format:
+   * @param {Object} tokenMap of data to send to Ledger
+   *   keys are the token uid and values are the token info in the format:
```
pedroferreira1, line 387: Should we write that token info is '...token info serialized as buffer in the format:'? 

(both comments in this method are applied to the method below as well

src/components/ModalAlertNotSupported.js
```diff
@@ -19,6 +19,7 @@ import ModalAlert from './ModalAlert.js';
 class ModalAlertNotSupported extends React.Component {
   render() {
     const renderBody = () => {
+      if (this.props.body) return this.props.body;
```
pedroferreira1, line 22: We already had this in `this.props.children`. Was it not enough?

```
return (
      <ModalAlert
        id="notSupported"
        title={this.props.title || t`Action not supported`}
        buttonName={t`Close`}
        body={this.props.children || renderBody()}
      />
    );
```

src/utils/version.js
```diff
@@ -64,6 +64,24 @@ const version = {
     } else {
       return false;
     }
+  },
+
+  /**
+   * Checks if custom tokens are allowed.
```
pedroferreira1, line 70: on ledger

src/utils/version.js
```diff
@@ -64,6 +64,24 @@ const version = {
...
+   *
+   * @param {Object} routerHistory History to push new path if we find no version on hw wallet.
```
pedroferreira1, line 72: WHat's this param?

src/constants.js
```diff
@@ -198,3 +198,24 @@ export const METADATA_CONCURRENT_DOWNLOAD = 10;
...
+/**
+ * Min Ledger app version supported by this wallet
+ */
+export const LEDGER_MIN_VERSION = '1.0.0';
+
+/**
+ * Max Ledger app version supported by this wallet
+ * Obs: max not inclusive, so we do NOT support this version
+ */
+export const LEDGER_MAX_VERSION = '1.2.0';
```
pedroferreira1, line 216: With this you are saying that you support v1.1.1, which is false as well, right? Why are you using this? And why making this not inclusive?

src/utils/helpers.js
```diff
@@ -133,6 +133,33 @@ const helpers = {
...
+   *
+   * @param {string} v1, first version string
+   * @param {string} v2, second version string
+   *
+   * @returns {Number} 0 if equal, 1 if v1 > v2 and -1 if v2 > v1
+   *
+   * @memberof Helpers
+   * @inner
+   */
+  cmpVersionString (v1, v2) {
```
pedroferreira1, line 150: Can you remove the space between the method name and the parentesis?

src/screens/WalletType.js
```diff
@@ -70,6 +70,17 @@ class WalletType extends React.Component {
    */
   handleVersion = (event, arg) => {
     if (arg.success) {
+      // compare ledger version with our min version
+      const version = arg.data.slice(3, 6).join('.');
+      hathorLib.storage.setItem('ledger:version', version);
+      if (
+        helpers.cmpVersionString(version, LEDGER_MIN_VERSION) < 0 ||
+        helpers.cmpVersionString(version, LEDGER_MAX_VERSION) >= 0
```
pedroferreira1, line 78: I'm thinking about this. Maybe we shouldn't do this check here. We still don't know if this wallet version will support the next ledger version. If we improve some ledger UX but the wallet continue working well, it will not support the ledger version?

src/utils/ledger.js
```diff
@@ -122,23 +122,43 @@ if (IPC_RENDERER) {
...
+          arr.push(hathorLib.transaction.intToBytes(0x80 | changeBuffer[0], 1))
+          // change output index on the second
+          arr.push(hathorLib.transaction.intToBytes(change.outputIndex, 1))
+          // Change key path of the address
+          arr.push(changeBuffer.slice(1));
+        } else {
+          // no change output
+          arr.push(hathorLib.transaction.intToBytes(0, 1));
+        }
+        initialData = Buffer.concat(arr);
```
pedroferreira1, line 146: We can move this outside this if/else like before, no? And move the `initialData` instantiation as well

src/constants.js
r4mmer: The idea is that any update that would break backwards compatibility would at least bump the version to 1.2.0 (since we are at 1.1.0). So I could assume we also support 1.1.1 and not have the wallet reject future Ledger apps (if they're compatible).

src/utils/ledger.js
r4mmer: Done in [86d6ee0](https://github.com/HathorNetwork/hathor-wallet/pull/264/commits/86d6ee02563d136387db484d7848ea2da6a5a3a4)

src/utils/helpers.js
r4mmer: Done in [86d6ee0](https://github.com/HathorNetwork/hathor-wallet/pull/264/commits/86d6ee02563d136387db484d7848ea2da6a5a3a4)

src/utils/version.js
r4mmer: It was removed.

src/utils/version.js
r4mmer: Done in [86d6ee0](https://github.com/HathorNetwork/hathor-wallet/pull/264/commits/86d6ee02563d136387db484d7848ea2da6a5a3a4)

src/components/ModalAlertNotSupported.js
r4mmer: Did not see the children, changed to use it and removed the body prop

public/ledger.js
r4mmer: Done in [86d6ee0](https://github.com/HathorNetwork/hathor-wallet/pull/264/commits/86d6ee02563d136387db484d7848ea2da6a5a3a4)

src/constants.js
pedroferreira1: I'm thinking about this. My main concern is that we might add another feature, bump to v1.2.0 and this new feature does not break anything in this wallet version. But we would require the other users to updated the wallet.

Maybe this is not a bad thing to be honest.

luislhl: Maybe we need a third kind of version, which would be the protocol version (discussed in other moments).

We would have to check if the protocol version used by the wallet and the ledger app match, instead of checking the Ledger App version.

Both the Wallet and Ledger App versions could be bumped without changing the Protocol version, if they don't break compatibility, which would make new versions still work together.

pedroferreira1 (approved):

PR #268, opened 2022-01-11
feat(ledger): check ledger version

r4mmer:
### Acceptance Criteria
- Include min and max ledger version
- Validate ledger version and show error message if invalid

### Security Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "react-router-dom": "^5.0.1",
     "react-scripts": "^3.0.1",
     "redux": "^4.0.0",
+    "semver-compare": "^1.0.0",
     "ttag": "^1.7.22",
     "viz.js": "^2.1.2"
   },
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -47,6 +47,7 @@
     "react-router-dom": "^5.0.1",
     "react-scripts": "^3.0.1",
     "redux": "^4.0.0",
+    "semver-compare": "^1.0.0",
```
luislhl, line 50: This lib is just a function with 13 lines. Maybe we should just copy it to our codebase instead of adding this dependency.

> Make sure you do not include new dependencies in the project unless strictly necessary and do not include dev-dependencies as production ones. More dependencies increase the possibility of one of them being hijacked and affecting us.

r4mmer: Done.

luislhl (approved):

PR #270, opened 2022-01-19
feat: add context menu to copy/paste using only the mouse

### LINKED ISSUE #211: Enable copy/paste right click on text
It'd be good to enable users to copy/paste by opening the traditional right click menu.

https://github.com/electron/electron/issues/4068#issuecomment-170911307
https://github.com/sindresorhus/electron-context-menu

tuliomir:
### Acceptance Criteria
- Shows a context menu on mouse right click on all text input fields allowing copy and paste
- Does not allow copy on password fields
- Resolves #211 

### Security Checklist

 ### UI/UX

https://user-images.githubusercontent.com/3298774/150235177-42f953f5-4b95-41d1-97fc-3f590820a36f.mov

### DIFF
--- a/public/contextMenu.js
+++ b/public/contextMenu.js
@@ -27,8 +27,6 @@ const webContents = win => win.webContents || (win.id && win);
  * @returns {*}
  */
 const removeUnusedMenuItems = menuTemplate => {
-  let notDeletedPreviousElement;
-
   return menuTemplate
     .filter(menuItem => {
       // First rules to remove an item
--- a/public/contextMenu.js
+++ b/public/contextMenu.js
@@ -22,237 +22,237 @@ const electron = require('electron');
 const webContents = win => win.webContents || (win.id && win);
 
 const decorateMenuItem = menuItem => {
-    return (options = {}) => {
-        if (options.transform && !options.click) {
-            menuItem.transform = options.transform;
-        }
+  return (options = {}) => {
+    if (options.transform && !options.click) {
+      menuItem.transform = options.transform;
+    }
 
-        return menuItem;
-    };
+    return menuItem;
+  };
 };
 
 const removeUnusedMenuItems = menuTemplate => {
-    let notDeletedPreviousElement;
+  let notDeletedPreviousElement;
 
-    return menuTemplate
-        .filter(menuItem => menuItem !== undefined && menuItem !== false && menuItem.visible !== false && menuItem.visible !== '')
-        .filter((menuItem, index, array) => {
-            const toDelete = menuItem.type === 'separator' && (!notDeletedPreviousElement || index === array.length - 1 || array[index + 1].type === 'separator');
-            notDeletedPreviousElement = toDelete ? notDeletedPreviousElement : menuItem;
-            return !toDelete;
-        });
+  return menuTemplate
+    .filter(menuItem => menuItem !== undefined && menuItem !== false && menuItem.visible !== false && menuItem.visible !== '')
+    .filter((menuItem, index, array) => {
+      const toDelete = menuItem.type === 'separator' && (!notDeletedPreviousElement || index === array.length - 1 || array[index + 1].type === 'separator');
+      notDeletedPreviousElement = toDelete ? notDeletedPreviousElement : menuItem;
+      return !toDelete;
+    });
 };
 
 const create = (win, options) => {
-    const handleContextMenu = (event, props) => {
-        if (typeof options.shouldShowMenu === 'function' && options.shouldShowMenu(event, props) === false) {
-            return;
+  const handleContextMenu = (event, props) => {
+    if (typeof options.shouldShowMenu === 'function' && options.shouldShowMenu(event, props) === false) {
+      return;
+    }
+
+    const {editFlags} = props;
+    const hasText = props.selectionText.trim().length > 0;
+    const can = type => editFlags[`can${type}`] && hasText;
+
+    const defaultActions = {
+      separator: () => ({type: 'separator'}),
+      cut: decorateMenuItem({
+        id: 'cut',
+        label: 'Cu&t',
+        enabled: can('Cut'),
+        visible: props.isEditable,
+        click(menuItem) {
+          const target = webContents(win);
+
+          if (!menuItem.transform && target) {
+            target.cut();
+          } else {
+            props.selectionText = menuItem.transform ? menuItem.transform(props.selectionText) : props.selectionText;
+            electron.clipboard.writeText(props.selectionText);
+          }
         }
+      }),
+      copy: decorateMenuItem({
+        id: 'copy',
+        label: '&Copy',
+        enabled: can('Copy'),
+        visible: props.isEditable || hasText,
+        click(menuItem) {
+          const target = webContents(win);
 
-        const {editFlags} = props;
-        const hasText = props.selectionText.trim().length > 0;
-        const can = type => editFlags[`can${type}`] && hasText;
-
-        const defaultActions = {
-            separator: () => ({type: 'separator'}),
-            cut: decorateMenuItem({
-                id: 'cut',
-                label: 'Cu&t',
-                enabled: can('Cut'),
-                visible: props.isEditable,
-                click(menuItem) {
-                    const target = webContents(win);
-
-                    if (!menuItem.transform && target) {
-                        target.cut();
-                    } else {
-                        props.selectionText = menuItem.transform ? menuItem.transform(props.selectionText) : props.selectionText;
-                        electron.clipboard.writeText(props.selectionText);
-                    }
-                }
-            }),
-            copy: decorateMenuItem({
-                id: 'copy',
-                label: '&Copy',
-                enabled: can('Copy'),
-                visible: props.isEditable || hasText,
-                click(menuItem) {
-                    const target = webContents(win);
-
-                    if (!menuItem.transform && target) {
-                        target.copy();
-                    } else {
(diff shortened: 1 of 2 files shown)


public/contextMenu.js
```diff
@@ -0,0 +1,409 @@
+/*
+MIT License
+
+Copyright (c) Sindre Sorhus <sindresorhus@gmail.com> (https://sindresorhus.com)
+Original from: https://github.com/sindresorhus/electron-context-menu
```
luislhl, line 5: Maybe we should make it explicit (here or in some other place) the version of the original on which ours is based.

This will make it easier if we need to update the code with newer versions

public/contextMenu.js
```diff
@@ -0,0 +1,409 @@
+/*
```
luislhl, line 1: Did you make any changes from the original?

public/contextMenu.js
```diff
@@ -0,0 +1,409 @@
...
+
+    The above copyright notice and this permission notice shall be included in all copies or substantial portions of the
+    Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE
+    WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS
+    OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+    OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+const electron = require('electron');
```
pedroferreira1, line 20: I think we should keep only the minimum required for our feature and not the whole file. Besides that, would be interesting to add support for `debugMode`. We have this in our `public/electron.js` and this could add the 'Inspect element' in the context menu.

pedroferreira1 (changes requested):

public/contextMenu.js
tuliomir: Yes, I removed the features that requires new dependencies to be installed.
Basically commented lines 21 to 23 and then commented every code instance that used them.

public/contextMenu.js
tuliomir: The features we did not use were now removed from our code.

The overall library architecture was kept despite being a bit over-sized for our needs because it is already a stable code, handles edge cases and has good maintenance.

public/contextMenu.js
tuliomir: Added a reference to the specific release we based our code on on the url itself.

public/electron.js
```diff
@@ -41,6 +42,18 @@ const debugMode = (
   process.argv.indexOf('--debug') >= 0
 );
 
+// Creates the context menu on mouse right-click for allowing copy/paste and other configurable actions
+contextMenu({
+  menu: (actions, props, browserWindow, dictionarySuggestions) => {
+    return [
+        ...dictionarySuggestions,
```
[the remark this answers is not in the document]
tuliomir: That was the recommended way to call this library, as per the author's `readme` file.

It was one of the features removed on the code cleanup, this function call is a lot more straightforward now.

public/contextMenu.js
tuliomir: A new cleanup was made, removing many features we didn't need.
Our version is now significantly different from the original.

luislhl (approved):

public/contextMenu.js
```diff
@@ -0,0 +1,258 @@
+/*
```
andreabadesso, line 1: **Comment unrelated to line number**

I know we don't have a linter on this project yet, but I think we should refactor this entire file to be more consistent with the code style from the other files by using 2 spaces indentation

andreabadesso (changes requested):

public/contextMenu.js
tuliomir: Good point! The file was refactored.

public/contextMenu.js
```diff
@@ -0,0 +1,258 @@
...
+    .filter((menuItem, index, array) => {
+      const toDelete = menuItem.type === 'separator' && (!notDeletedPreviousElement || index === array.length - 1 || array[index + 1].type === 'separator');
+      notDeletedPreviousElement = toDelete ? notDeletedPreviousElement : menuItem;
+      return !toDelete;
+    });
+};
+
+const create = (win, options) => {
+  const handleContextMenu = (event, props) => {
+    if (typeof options.shouldShowMenu === 'function' && options.shouldShowMenu(event, props) === false) {
```
pedroferreira1, line 48: What's this `options.shouldShowMenu`? I think we don't use it.

public/contextMenu.js
```diff
@@ -0,0 +1,258 @@
...
+
+    const {editFlags} = props;
+    const hasText = props.selectionText.trim().length > 0;
+    const can = type => editFlags[`can${type}`] && hasText;
+
+    const defaultActions = {
+      separator: () => ({type: 'separator'}),
+      cut: decorateMenuItem({
+        id: 'cut',
+        label: 'Cu&t',
```
pedroferreira1, line 60: What's this &?

public/contextMenu.js
```diff
@@ -0,0 +1,258 @@
...
+        enabled: can('Cut'),
+        visible: props.isEditable,
+        click(menuItem) {
+          const target = webContents(win);
+
+          if (!menuItem.transform && target) {
```
pedroferreira1, line 66: What's this `if` doing?

public/contextMenu.js
```diff
@@ -0,0 +1,258 @@
...
+
+    let dictionarySuggestions = [];
+
+    let menuTemplate = [
+      defaultActions.cut(),
+      defaultActions.copy(),
+      defaultActions.paste(),
+      defaultActions.separator(),
+      shouldShowInspectElement && defaultActions.inspect(),
+      defaultActions.separator()
```
pedroferreira1, line 130: If don't should the inspect element, then show two separators in a row?

public/contextMenu.js
```diff
@@ -0,0 +1,258 @@
...
+    ];
+
+    if (options.menu) {
```
pedroferreira1, line 133: The only options parameter we are using is `showInspectElement`, right? Should we remove all those `if`'s?

public/contextMenu.js
```diff
@@ -0,0 +1,258 @@
...
+
+    if (options.append) {
+      const result = options.append(defaultActions, props, win, event);
+
+      if (Array.isArray(result)) {
+        menuTemplate.push(...result);
+      }
+    }
+
+    menuTemplate = removeUnusedMenuItems(menuTemplate);
```
pedroferreira1, line 153: What's this doing?

public/contextMenu.js
```diff
@@ -0,0 +1,258 @@
...
+
+    for (const menuItem of menuTemplate) {
```
pedroferreira1, line 155: Are we using this?

public/contextMenu.js
tuliomir: & is a syntax for the context menu to know where to put the underscore on the word. And then, what "hotkey" will be assigned to that button.

In that case, a right click shows the context menu and if the user types `T` on the keyboard, the system runs the "Cut" operation.

public/contextMenu.js
tuliomir: It is an optional function to allow the context menu to understand if, under runtime conditions, it should not be drawn.

We don't use it, I will remove it as well.

public/contextMenu.js
tuliomir: Another generic structure that allows custom behavior when the operation is executed.
It is out of our scope, I'll remove it.

public/contextMenu.js
tuliomir: The lib was structured in a way that it first includes all menuItems and then runs a `can` function to remove those that are not relevant for the runtime.

Since we won't handle a dynamic amount of menuItems we could do it the other way around, checking if the elements are needed before adding, but I don't think it will improve the code in a meaningful way.

I'll add some inline documentation to improve the code readability.

public/contextMenu.js
tuliomir: No, removing this section as well.

public/contextMenu.js
tuliomir: The "removeUsedMenuItems" had a logic to avoid two consecutive separators or a separator as a first/last element.
Since we also removed dynamic menuItems, this was further simplified.

tuliomir:
@pedroferreira1 , a complete jsdoc was implemented on the contextMenu library and its scope was further restricted to our current use of Copy/Paste/Cut.
This should help with code clarity.

public/contextMenu.js
tuliomir: Removed :+1:

public/contextMenu.js
```diff
@@ -0,0 +1,249 @@
...
+
+const webContents = win => win.webContents || (win.id && win);
+
+/**
+ * Removes menuItems that are not being used
+ * @param menuTemplate
+ * @returns {*}
+ */
+const removeUnusedMenuItems = menuTemplate => {
+  let notDeletedPreviousElement;
```
pedroferreira1, line 30: This is not used, right? Should we delete it?

pedroferreira1 (approved):

PR #272, opened 2022-01-21
feat: displaying correct script data on tx details screen

### LINKED ISSUE #265: Show data output script decoded in Transaction Detail screen
### Motivation

We are already showing in the explorer the data script (used in NFT creation tx) decoded but in the wallet we are still showing 'Unable to decode' and some users don't understand it.

### Acceptance criteria

Data scripts should be shown correctly and not 'Unable to decode'.

tuliomir:
### Acceptance Criteria
- On "Transactrion Details" screen, data scripts should be shown correctly and not 'Unable to decode'.
- Specifically NFT creation transactions must show their script output correctly
- Resolves #265 

**Old screen**
Screenshot from 2022-01-20 21-52-22

**New screen**
Screenshot from 2022-01-20 21-43-57

This feature is similar to the Explorer's PR 141: HathorNetwork/hathor-explorer#141

### Other adjustments
Hathor Wallet Lib dependency updated to 0.29

### Security Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   "version": "0.21.1",
   "private": true,
   "dependencies": {
-    "@hathor/wallet-lib": "^0.26.0",
+    "@hathor/wallet-lib": "^0.29.0",
     "@ledgerhq/hw-transport-node-hid": "^4.73.4",
     "@sentry/electron": "^0.17.0",
     "babel-polyfill": "^6.26.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1135,9 +1135,9 @@
       }
     },
     "@hathor/wallet-lib": {
-      "version": "0.26.0",
-      "resolved": "https://registry.npmjs.org/@hathor/wallet-lib/-/wallet-lib-0.26.0.tgz",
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -24,7 +24,7 @@
   "version": "0.21.1",
   "private": true,
   "dependencies": {
-    "@hathor/wallet-lib": "^0.26.0",
+    "@hathor/wallet-lib": "^0.29.0",
```
andreabadesso, line 27: Is this version upgrade safe?

I don't remember what changed from version 0.26.0 to 0.29.0, maybe only wallet-service related stuff, but I think it is worth checking if any breaking changes happened

andreabadesso (changes requested):

package.json
tuliomir: From what I could read on [the releases page](https://github.com/HathorNetwork/hathor-wallet-lib/releases) and on [the difference between the codes](https://github.com/HathorNetwork/hathor-wallet-lib/compare/v0.26.0...v0.29.1) I could find no breaking change notification. Also, no change to dependencies.

All tests made locally on the wallet were successful.
Do you have any pointers to edge cases that may have been broken and must be tested?

pedroferreira1 (approved):

andreabadesso (approved):

package.json
andreabadesso: I don't, if you think this is safe, we can run the QA before merging to master

PR #283, opened 2022-02-07
feat: check signatures on startup

r4mmer:
### Acceptance Criteria
During hardware wallet startup we should check all token signatures to make sure they're valid.
The util used to start hardware wallet (and where we were checking the signatures) is not being used so we have to move this to the actual place where we start the hardware wallet and remove the unused util.

### Security Checklist

### DIFF
--- a/src/screens/WalletType.js
+++ b/src/screens/WalletType.js
@@ -11,6 +11,7 @@ import { t } from 'ttag'
 import logo from '../assets/images/hathor-logo.png';
 import wallet from '../utils/wallet';
 import ledger from '../utils/ledger';
+import version from '../utils/version';
 import helpers from '../utils/helpers';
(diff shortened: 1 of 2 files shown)


src/screens/WalletType.js
```diff
@@ -118,6 +119,27 @@ class WalletType extends React.Component {
 
       wallet.startWallet(null, '', null, '', this.props.history, false, xpub);
       hathorLib.wallet.markBackupAsDone();
+
+      const tokenSignatures = hathorLib.storage.getItem('wallet:token:signatures');
```
luislhl, line 123: Maybe we could keep this logic in a utils file (wallet or ledger).

But that's just a suggestion.

luislhl (approved):

pedroferreira1 (approved):

PR #298, opened 2022-04-08
feat: get mint melt authority from new facade methods

andreabadesso:
## Motivation

The wallet-service did not have the `getMintAuthority` and `getMeltAuthority` methods, they have been added on **this PR**, so this PR is the implementation of those methods on the wallet-desktop

### Acceptance Criteria
- We should use the `getMintAuthority` and `getMeltAuthority` from the [new normalized method on both facades](https://github.com/HathorNetwork/hathor-wallet-lib/pull/344)

### Security Checklist

### DIFF
--- a/src/components/TokenAdministrative.js
+++ b/src/components/TokenAdministrative.js
@@ -98,34 +98,16 @@ class TokenAdministrative extends React.Component {
   /**
    * Update token state after didmount or props update
    */
-  updateWalletInfo = () => {
-    if (!this.props.token || !this.props.token.uid) {
+  updateWalletInfo = async () => {
+    if (!this.props.token && !this.props.token.uid) {
       return;
     }
 
-    const { uid } = this.props.token;
+    const mintUtxos = await this.props.wallet.getMintAuthority(this.props.token.uid, { many: true });
+    const meltUtxos = await this.props.wallet.getMeltAuthority(this.props.token.uid, { many: true });
 
-    let mintCount = 0;
-    let meltCount = 0;
-
-    const tokenBalance = this.props.tokensBalance[uid];
-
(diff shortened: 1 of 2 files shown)


r4mmer (approved):

andreabadesso:
@pedroferreira1 @r4mmer 

I had to refactor a little the `Wallet` screen component because there was a bug with the `shouldShowAdministrativeTab` method

Can you please check again?

Thanks

src/components/TokenAdministrative.js
```diff
@@ -98,34 +98,16 @@ class TokenAdministrative extends React.Component {
...
-      meltCount = canMeltUtxos ? 1 : 0;
-    } else {
-      // Old Facade
-      const mintUtxos = this.props.wallet.getMintAuthority(this.props.token.uid, { many: true });
-      const meltUtxos = this.props.wallet.getMeltAuthority(this.props.token.uid, { many: true });
-
-      mintCount = mintUtxos ? mintUtxos.length : 0;
-      meltCount = meltUtxos ? meltUtxos.length : 0;
-    }
+    const mintCount = mintUtxos ? mintUtxos.length : 0;
```
pedroferreira1, line 109: Now that you changed the return to always return an array (empty when we were returning `null` before), then we can directly use here:

```
const mintCount = mintUtxos.length;
const meltCount = meltUtxos.length;
```

pedroferreira1 (approved):

src/components/TokenAdministrative.js
andreabadesso: Done in 3d1e2fa0cb04a84fc42213ed7c5e95e72f4969d6

Thanks

r4mmer (approved):

PR #313, opened 2022-05-20
Docs/Update QA for "hide zero balance tokens" feature

tuliomir:
### Acceptance Criteria
- There should be new steps for testing the "Hide zero-balance tokens" on the QA document

### Security Checklist

### DIFF
--- a/QA.md
+++ b/QA.md
@@ -111,10 +111,37 @@ You can connect your wallet to the testnet (https://node1.foxtrot.testnet.hathor
     1. Melt 20.00 tokens. The total supply and your balance must be 130.00 now.
     1. Click on 'Balance & History' and copy your address. Then go back to 'Administrative Tools' and click to delegate mint.
     1. Type your address and delegate. Now you must have 2 mint outputs.
-    1. Click to delegate melt and deselect the option to 'Create another melt output for you?'. Delegate the authority to the address WZ7pDnkPnxbs14GHdUFivFzPbzitwNtvZo and now you must have no melt outputs.
+    1. Click to delegate melt and deselect the option to 'Create another melt output for you?'. Delegate the authority to the address `WZ7pDnkPnxbs14GHdUFivFzPbzitwNtvZo` and now you must have no melt outputs.
     1. Go back to the 'Administrative Tools', click to destroy mint and select 3. You must see an error message that you don't have 3 mint authority outputs.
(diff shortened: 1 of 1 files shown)


QA.md
```diff
@@ -140,22 +169,22 @@ You can connect your wallet to the testnet (https://node1.foxtrot.testnet.hathor
 1. **Register tokens with same name**
     1. Unregister TST token but copy its configuration string and save it.
     1. Click on 'Custom tokens' and then 'Create a new token'.
-    1. Type 'Test token' in the token name, 'Test' in the symbol and 200 in the amount.
+    1. Type 'Test token' in the token name, 'TST' in the symbol and 200 in the amount.
```
andreabadesso, line 172: For this step, the symbol don't necessarily needs to be the same as the previous one as the check is on the token name

I'm not agains't this change tho

QA.md
```diff
@@ -140,22 +169,22 @@ You can connect your wallet to the testnet (https://node1.foxtrot.testnet.hathor
...
+    1. Click on the first transaction with amount 0.00.
     1. Copy the first input from HTR and from your address (copy the transaction ID and the index).
     1. Click on 'Send tokens'.
     1. Paste the address and 1.00 as the amount. Deselect 'Choose inputs automatically' and paste the transaction ID and index copied before.
     1. Must see an error that the output is already spent.
 
 1. **Create NFT**
     1. Click on 'Custom tokens', then 'Create an NFT'.
-    1. Fill NFT data as 'test data', name and symbol as 'NFT Test' and 'NFTT', and amount as 100. Check only the melt authority checkbox.
+    1. Fill NFT data as `ipfs://test`, name and symbol as 'NFT Test' and 'NFTT', and amount as 100. Check only the melt authority checkbox.
```
andreabadesso, line 187: Oh yea, that would work too

But I think we can remove the ipfs:// from the next step that uses it

andreabadesso (changes requested):

QA.md
tuliomir: If only the name needs to be the same, I would rather revert my change to keep this clear. :+1:

QA.md
tuliomir: My intent here was just to keep the data consistent with the validation we make a few steps after.
Keeping both `ipfs://test` or `test data` would suffice.

pedroferreira1 (approved):

PR #320, opened 2022-06-03
chore: upgrade nodejs to v14 and dependencies accordingly

pedroferreira1:
### Acceptance Criteria

The main goal of this PR is to be able to generate wallet executables with newer versions of nodejs with the minimum changes possible. We were requiring nodejs v8 for development before.

Using nodejs v16 brought many problems that would not be easily fixed, so I decided to go with v14 in a first upgrade project. This version is still one of the active versions and we can do another upgrade from v14 -> v16 later.

Given that we are only 2 major versions behind the latest react version, I decided to upgrade electron and electron builder versions. 

**Electron builder**

I updated to the latest version.

image

**Electron**

I first tried to update to the latest version (v19) but we were in the v5. Because of this difference in major version, many problems happened, then I decided to do this in smaller steps and tried intermediary versions. v13 was one of the versions with more downloads in the previous week.

image

**Node Sass Chokidar**

This package was deprecated and we were not being able to use it anymore in newer versions of electron-builder. This is the other "big" change, that I switched to use `sass` instead of `node-sass-chokidar` and updated the scripts accordingly.

**Other packages**

The other packages that were upgraded were because the versions that we were using were not compatible with the new electron version.

**Send tx handler**

There was a small change in the prop types in the `src/components/SendTxHandler.js` file to remove a deprecation warning.

**Patch package**

The wallet uses our `wallet-lib`, which uses `bitcore-lib`, which uses `elliptic`, which uses `bn.js` package. In this new nodejs version, the `Buffer` module is available as a global variable and the `bn.js` is not ready for that, then we ended up having an error when using it. Because of this problem in this package (and because it's in this dependency chain, then we can't just fork the repo and simply use the fork), we had to use this `patch-package` package to add a patch to the node modules.

### Security Checklist


msbrogli (approved):
image

msbrogli (approved):

r4mmer (approved):

PR #322, opened 2022-06-10
chore: enable require on renderer process

r4mmer:
### Acceptance Criteria
- Enable Ledger to work with upgraded dependencies

### Explanation

To communicate with Ledger we use a library to send/receive data to the USB port and the version we currently use of this library cannot run on the renderer process. So we run any call to Ledger on the main process and use the `electron.IpcRenderer` to send and receive events between the main process and the renderer process.

On electron version 12 a breaking change was introduced ([changelog here](https://www.electronjs.org/docs/latest/breaking-changes#default-changed-contextisolation-defaults-to-true)) that changed the default of the `contextIsolation` which (among other things) enables or disables the use of `require` to import node packages on the renderer process, including the 'electron' package itself.

Once we disabled the `contextIsolation`  there was a second issue, in some places the binary type `Uint8Array` was being used instead of `Buffer` so we converted to `Buffer` where needed to use a consistent api for binary data.

### Security Checklist

### DIFF
--- a/public/electron.js
+++ b/public/electron.js
@@ -60,6 +60,7 @@ function createWindow () {
     icon: path.join(__dirname, iconOS),
     webPreferences: {
       nodeIntegration: true,
+      contextIsolation: false,
       preload: path.join(__dirname, 'preload.js')
     }
   })
--- a/src/components/ModalLedgerSignToken.js
+++ b/src/components/ModalLedgerSignToken.js
@@ -111,7 +111,8 @@ function ModalLedgerSignToken({token, modalId, cb}) {
     setWaitingLedger(false);
     if (arg.success) {
       // save token signature on storage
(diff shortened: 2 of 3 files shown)


r4mmer:
> The reason being that if we know that the Buffer implementation differs on javascript environments, we would at least centralize the refactoring points if they were to change again.

The wallet-lib already has a buffer utils, but the issue we are solving (in this PR) is that the IPC (the brigde between main and renderer process) where the Ledger data passes uses `Uint8Array`.
I tried to force the main process to send a `Buffer` but the IPC converted it to `Uint8Array` anyway.
So we use `Buffer` everywhere (main process, renderer, lib) but electron changes it on IPC which is only used for Ledger.

pedroferreira1 (approved):

PR #325, opened 2022-06-16
chore: refactor hardware wallet init screens into own file

obiyankenobi:
### Acceptance Criteria
- Hardware wallet initialization is its own file, with separate route;
- Wallet continues to function was before;

### Security Checklist

### Additional notes
This PR is needed for the multiple wallets change, as we need an easy way to go to the beginning of the hardware initialization steps.

### DIFF
--- a/src/screens/StartHardwareWallet.js
+++ b/src/screens/StartHardwareWallet.js
@@ -1,0 +1,231 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import { t } from 'ttag'
+
+import logo from '../assets/images/hathor-logo.png';
+import wallet from '../utils/wallet';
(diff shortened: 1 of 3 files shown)


src/screens/StartHardwareWallet.js
```diff
@@ -0,0 +1,231 @@
...
+import InitialImages from '../components/InitialImages';
+import hathorLib from '@hathor/wallet-lib';
+
+
+/**
+ * Screen used to select between hardware wallet or software wallet
+ *
+ * @memberof Screens
+ */
+class WalletType extends React.Component {
```
pedroferreira1, line 26: Should we call this class `StartHardwareWallet`?

pedroferreira1 (changes requested):

src/screens/StartHardwareWallet.js
obiyankenobi: Yes! Just fixed it.

pedroferreira1 (approved):

r4mmer (approved):

Issue #328, opened 2022-06-17
Token bar scroll not working in v0.23.2


pedroferreira1:

pedroferreira1:
Fixed on https://github.com/HathorNetwork/hathor-wallet/pull/329

PR #329, opened 2022-06-17
fix: add scroll to token bar component

pedroferreira1:
### Summary

With the upgrade of sass in the new version, the scroll in the token bar stopped working. The solution for this new version is to add a style `overflow: auto;` to the parent element to the one that needs to have a scroll.

### Acceptance Criteria
- Add scroll to token bar
- Add this test to the QA

#### Before the fix

image

#### After the fix

image

### Security Checklist


pedroferreira1:
Closes #328

r4mmer (approved):

tuliomir (approved):

PR #330, opened 2022-06-22
wip: add modal to register all tokens with information

pedroferreira1:
### Summary

The idea in this feature is to split the tokens in different categories, (i) the tokens with balance, which are always selected, (ii) the tokens with zero balance, that the user can select to be registered as well, and (iii) tokens that have conflict (same name/symbol as other token). Tokens with conflict can't be registered using this feature but the user can see a list of conflicts.

image

image

image

### Acceptance Criteria

- TODO

### Security Checklist


pedroferreira1:
This is cancelled for now

PR #331, opened 2022-06-29
feat: add feature to export configuration strings for registered tokens

pedroferreira1:
### Summary

Many users have hundreds of tokens and, when they reset their wallets, or change devices, they must register all of them again. We already have in the desktop wallet the possibility of a bulk import in the Unknown Tokens screen, we only needed a feature to export all registered tokens, so the users won't have to register all of them again.

### Acceptance Criteria
- Users should be able to export the configuration strings of all registered tokens

https://user-images.githubusercontent.com/3298774/176432891-98d8d429-b338-43b0-8463-4fd6c613615d.mov

- Following @tuliomir's suggestion (https://github.com/HathorNetwork/hathor-wallet/pull/331#issuecomment-1170330708) I've added a message in the reset modal about the export tokens feature.

image

### Security Checklist

### DIFF
--- a/src/screens/Settings.js
+++ b/src/screens/Settings.js
@@ -27,6 +27,7 @@ import { connect } from "react-redux";
 const mapStateToProps = (state) => {
   return {
     useWalletService: state.useWalletService,
+    registeredTokens: state.tokens,
   };
 };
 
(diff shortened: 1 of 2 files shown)


luislhl (approved):

luislhl (approved):

src/screens/Settings.js
```diff
@@ -95,6 +96,35 @@ class Settings extends React.Component {
...
+    // Create the hidden a element to trigger the download
+    const element = document.createElement('a');
+    const file = new Blob([text], {
+      type: 'text/plain'
+    });
+    element.href = URL.createObjectURL(file);
+    element.download = 'Hathor Wallet - Tokens.txt';
+    document.body.appendChild(element);
+    element.click();
+    element.remove();
```
r4mmer, line 125: I know this is required when downloading a blob from a browser, but why can't we use the `fs` module?

r4mmer (approved):

Issue #332, opened 2022-07-06
Wallet does not release Ledger connection when not in use


r4mmer:
# Description

The wallet opens a connection to the Ledger device (USB connection using node-hid) and does not release while not in use.
Not being in use means that the wallet is not listening for a response to a command sent.
This makes any other software that wants to interact with the device not able to do so while the wallet is open, e.g. Ledger Live.

# Related

[Node HID integration docs](https://developers.ledger.com/docs/transport/node-electron-hid/)
[Line that opens the connection](https://github.com/HathorNetwork/hathor-wallet/blob/f83cc9d8ef89974e958111b5ae4cfad6f4eb806c/public/ledger.js#L219)

PR #333, opened 2022-07-08
feat: release ledger transport when not in use

r4mmer:
### Acceptance Criteria

- Close transport when not in use
- Do not change behavior for current Hardware wallet

### Behavior

https://user-images.githubusercontent.com/15909384/178019501-d1458fe6-d3c0-4ae2-87c3-94d527bdb43c.mp4

When opening the manager, no interaction with the Ledger was made.
Ledger Live manager closed the Hathor app, which signaled to the wallet to close, then asked confirmation to trust the manager. 

### Security Checklist

### DIFF
--- a/public/ledger.js
+++ b/public/ledger.js
@@ -128,8 +128,10 @@ class Ledger {
           this.hathorAppOpened = false;
           // If we don't close it when changing the app we get an error
           // https://github.com/LedgerHQ/ledgerjs/issues/22
-          this.transport.device.close();
-          this.transport = null;
+          if (this.transport) {
+            this.transport.device.close();
+            this.transport = null;
+          }
         }
       },
       error: error => {},
@@ -170,6 +172,11 @@ class Ledger {
         this.checkSendQueue();
       });
     } else {
+      // no more commands on queue, we can release the transport connection
+      if (this.transport) {
+        this.transport.device.close();
+        this.transport = null;
+      }
       this.canSend = true;
     }
   }


r4mmer:
closes #332

src/App.js
```diff
@@ -191,7 +191,7 @@ const returnStartedRoute = (Component, props, rest) => {
   if (hathorLib.wallet.loaded()) {
     // Wallet is locked, go to locked screen
     if (hathorLib.wallet.isLocked()) {
-      return <Redirect to={{pathname: '/locked/'}}/>;
+      return <Redirect to={{pathname: '/locked/', hash: rest.location.hash}}/>;
```
r4mmer, line 194: Before redirecting to `/wallet_type/#ledger:closed` we lock the wallet and this is the actual route taken.
Below the `returnDefaultComponent` method will intercept the `/locked/` route and redirect back to `/wallet_type/` for hardware wallets, we need to pass the hash in both instances so that it reaches the actual wallet type screen.

src/actions/index.js
```diff
@@ -166,3 +166,8 @@ export const resolveLockWalletPromise = (pin) => ({ type: "resolve_lock_wallet_p
  * This will reset the selected token if the one selected has been hidden because of zero balance
  */
 export const resetSelectedTokenIfNeeded = () => ({ type: "reset_selected_token_if_needed" });
+
+/**
+ * This will be used then the Ledger app closes to warn the user, and after the user is notified.
```
pedroferreira1, line 171: Typo when*

Also missing the parameter type

src/screens/LoadingAddresses.js
```diff
@@ -43,7 +43,7 @@ class LoadingAddresses extends React.Component {
   /**
    * canRedirect {boolean} set if can already redirect to the screen after loading
    */
-  state = { canRedirect: false };
+  state = { canRedirect: false, timeoutId: null };
```
pedroferreira1, line 46: Missing this new parameter in the docstring

src/screens/LoadingAddresses.js
```diff
@@ -61,9 +61,15 @@ class LoadingAddresses extends React.Component {
     }
     // To prevent only a blink in this screen when user loads the addresses really fast
     // I set that the user will see this screen at least for 2 seconds
-    setTimeout(() => {
+    const timeoutId = setTimeout(() => {
       this.setState({ canRedirect: true });
     }, 2000);
+    this.setState({ timeoutId });
```
pedroferreira1, line 67: Maybe this `timeoutId` could be `this.timeoutId` because it's not being used to render anything, only to be used in the class. What do you think?

src/screens/LoadingAddresses.js
```diff
@@ -61,9 +61,15 @@ class LoadingAddresses extends React.Component {
...
+  }
+
+  componentWillUnmount = () => {
+    clearTimeout(this.state.timeoutId);
```
pedroferreira1, line 71: Is there any chance we might get here before setting `timeoutId`? I would add a `if (this.timeoutId)` just to guarantee we don't get any errors here.

src/screens/LoadingAddresses.js
r4mmer: Removed timeoutId from state.

src/screens/LoadingAddresses.js
r4mmer: `componentWillUnmount` will never run before `componentDidMount`, but if the `setTimeout` callback has already been called this `clearTimeout` will be a no-op.

src/screens/LoadingAddresses.js
r4mmer: Done

src/actions/index.js
r4mmer: Added

pedroferreira1 (approved):

src/App.js
r4mmer: Outdated

tuliomir (approved):

PR #340, opened 2022-07-28
fix: ledger queue

r4mmer:
### Acceptance Criteria
- queue executions should manage transport open/close to avoid sync issues.

### Security Checklist

### DIFF
--- a/public/ledger.js
+++ b/public/ledger.js
@@ -16,7 +16,7 @@ const ledgerCLA = 0xe0;
 const ledgerOK = Buffer.from('9000', 'hex');
 
 class Ledger {
-  
+
   static parseLedgerError(e) {
     const errorMap = {
       0x6a86:  "WRONG_P1P2",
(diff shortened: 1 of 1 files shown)


public/ledger.js
```diff
@@ -154,62 +154,64 @@ class Ledger {
...
    *
    * @return {Promise} Promise resolved when there's a response from Ledger
    */
-  sendToLedgerOrQueue = (transport, command, p1, p2, data) => {
-    if (this.canSend) {
-      this.canSend = false;
-      const promise = transport.send(ledgerCLA, command, p1, p2, data);
-      promise.finally(() => {
+  sendToQueue = (command, p1, p2, data) => {
+    console.log('sending to Ledger:', command, p1, p2);
```
pedroferreira1, line 195: Here we send to queue only, right?

pedroferreira1 (approved):

public/ledger.js
```diff
@@ -154,62 +154,64 @@ class Ledger {
...
+
+    const { command, p1, p2, data, resolve, reject } = this.sendQueue.shift();
+    const promise = this.getTransport().then(transport => transport.send(ledgerCLA, command, p1, p2, data));
+    promise.then((response) => {
+      resolve(response);
+    }, (error) => {
+      reject(error);
+    }).finally(() => {
+      // When done, check the queue again
+      this.checkSendQueue();
```
tuliomir, line 184: I'm slightly worried about a recursion when handling queue consumption, but since we're dealing with very slow inputs ( user manipulating a small physical device ) and small queues, I think it's no reason to make any changes to this approach.

tuliomir (approved):

public/ledger.js
```diff
@@ -154,62 +154,64 @@ class Ledger {
...
+    });
```
tuliomir, line 185: As a code style suggestion:
```suggestion
const { command, p1, p2, data, resolve, reject } = this.sendQueue.shift();
    this.getTransport()
      .then((transport) => transport.send(ledgerCLA, command, p1, p2, data))
      .then((response) => resolve(response), (error) => reject(error))
      .finally(() => {
        // When done, check the queue again
        this.checkSendQueue();
      });
```

PR #344, opened 2022-10-17
fix: upgrade wallet lib and other fixes

andreabadesso:
### Acceptance Criteria
- Fixes an issue on the UnknownTokens page where the tokens would load history indefinetely (1c3e3c118526258b6270584a20bb74c897c0b962)
- Fixes an issue on the hideZeroBalances feature that wouldn't display any tokens because we didn't load their balances (2a90bb694b5591313ac64959e2892b0d395a295b)
- Fixes an issue where the wallet would get stuck on the loading screen after the connection was down and up again (c9499b158971c6221477f9a15ddd2bd92b9887fd and c4f06a8d0b30ab67802adab06112e05773609039)
- Fixes `startWallet` while on Ledger (497dd83da24c949d8b6fe64ce3a2a841a3dbc9b7)

### Security Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4439,9 +4439,9 @@
       }
     },
     "bitcore-lib": {
-      "version": "8.25.40",
-      "resolved": "https://registry.npmjs.org/bitcore-lib/-/bitcore-lib-8.25.40.tgz",
-      "integrity": "sha512-mb6kvfhoiIdoyFsDlhIFVst3HpeGjGYBf0XDxTdZ+H07EC4JuiViA3bnQ5uZbZjHFngEl0GTPaoK1Zaolutw4A==",
+      "version": "8.25.28",
+      "resolved": "https://registry.npmjs.org/bitcore-lib/-/bitcore-lib-8.25.28.tgz",
+      "integrity": "sha512-UrNHh0Ba8GUiHUYRmm2IKlb8eomsbvk/Z6oQdaOPQoLiamiKnu45pAMqtcHg06wMDF8at54oIdoD2WEU+TQujw==",
       "requires": {
         "bech32": "=2.0.0",
         "bip-schnorr": "=0.6.4",
@@ -4454,11 +4454,11 @@
       }
     },
     "bitcore-mnemonic": {
-      "version": "8.25.40",
-      "resolved": "https://registry.npmjs.org/bitcore-mnemonic/-/bitcore-mnemonic-8.25.40.tgz",
-      "integrity": "sha512-+4tkEwt+F0G1rLGLKCe9m8BhirUu0KNqzvjXvlxmO5S65MelHyAns5/R1CeKH3jEZJDQLghgKPwITfIbe+/uIg==",
+      "version": "8.25.28",
+      "resolved": "https://registry.npmjs.org/bitcore-mnemonic/-/bitcore-mnemonic-8.25.28.tgz",
+      "integrity": "sha512-z4EB7r1JYqJCl31rpfm9qZX24tJ+PPvgAk7xvD6x8pqyv60j7hDTn1zt2mDQu0IxGlGMSnQmGmddZf3UcQXQ8w==",
       "requires": {
-        "bitcore-lib": "^8.25.40",
+        "bitcore-lib": "^8.25.28",
         "unorm": "^1.4.1"
       }
     },
--- a/package-lock.json
+++ b/package-lock.json
@@ -1520,9 +1520,9 @@
       }
     },
     "@hathor/wallet-lib": {
-      "version": "0.40.1",
-      "resolved": "https://registry.npmjs.org/@hathor/wallet-lib/-/wallet-lib-0.40.1.tgz",
-      "integrity": "sha512-HyzeN0Iar2AximKRKHXo/ffLhwLYOMjWwqGG2dGOHeQOWgrOx5wU0myVkeG0faKT6x+7AxT7oWtykoiK8FwmeA==",
+      "version": "0.41.0",
+      "resolved": "https://registry.npmjs.org/@hathor/wallet-lib/-/wallet-lib-0.41.0.tgz",
+      "integrity": "sha512-YXQeqP6v+hnN1o2iHELSvT2X6GD86lWt6RWmEbw1nazWPB4R88Yx4GMBtf4yH9gr6r0LZ001aOnIF6EA8uM0Ww==",
       "requires": {
         "axios": "^0.21.4",
         "bitcore-lib": "^8.25.10",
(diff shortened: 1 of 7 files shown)


r4mmer (approved):

package-lock.json
```diff
@@ -4439,9 +4439,9 @@
       }
     },
     "bitcore-lib": {
-      "version": "8.25.28",
-      "resolved": "https://registry.npmjs.org/bitcore-lib/-/bitcore-lib-8.25.28.tgz",
-      "integrity": "sha512-UrNHh0Ba8GUiHUYRmm2IKlb8eomsbvk/Z6oQdaOPQoLiamiKnu45pAMqtcHg06wMDF8at54oIdoD2WEU+TQujw==",
+      "version": "8.25.40",
```
pedroferreira1: We should revert this update

src/actions/index.js
```diff
@@ -292,6 +294,15 @@ export const walletReloadData = () => ({
   type: types.WALLET_RELOAD_DATA,
 });
 
+export const storeRouterHistory = (routerHistory) => ({
```
pedroferreira1, line 297: Missing parameter docstring

src/screens/UnknownTokens.js
```diff
@@ -249,8 +249,10 @@ class UnknownTokens extends React.Component {
 
       return unknownTokens.map((token, index) => {
         const isNFT = helpers.isTokenNFT(token.uid, this.props.tokenMetadata);
-        const tokenBalance = get(this.props.tokensBalance, `${token.uid}`, { status: TOKEN_DOWNLOAD_STATUS.LOADING });
-        const tokenHistory = get(this.props.tokensHistory, `${token.uid}`, { status: TOKEN_DOWNLOAD_STATUS.LOADING });
+
+        // Note: We can't default to LOADING here otherwise we will never show the `Show History` button
```
pedroferreira1, line 253: Why? This would be "LOADING" only until the balance/history if fetched.

src/sagas/wallet.js
```diff
@@ -287,12 +291,17 @@ export function* loadTokens() {
...
-  // and continue, loading the tokens asynchronously
-  for (const token of registeredTokens) {
+  // Dispatch actions to asynchronously load the balances of each token the wallet has
+  // ever interacted with. The `put` effect will just dispatch and continue, loading
+  // the tokens asynchronously.
+  //
+  // Note: We need to download the balance of all the tokens from the wallet so we can
+  // hide zero-balance tokens
+  for (const token of allTokens) {
     yield put(tokenFetchBalanceRequested(token));
```
pedroferreira1, line 301: After we finish fetching the balance, do we fetch the history automatically? If that's the case, then we could fetch the history only from the registered tokens.

src/sagas/wallet.js
```diff
@@ -531,10 +545,37 @@ export function* onWalletConnStateUpdate({ payload }) {
...
+    // Store all tokens on redux as we might have lost tokens during the disconnected
+    // period.
+    const { allTokens } = yield call(loadTokens);
+
+    // Load success, we can send the user back to the wallet screen
+    yield put(loadWalletSuccess(allTokens));
+    routerHistory.replace('/wallet/');
+    yield put(loadingAddresses(false));
+  } catch (e) {
+    yield put(startWalletFailed());
```
pedroferreira1, line 569: In the `startWallet` method you have

```
yield cancel(threads);
```

Should we cancel any thread here as well?

We also have some different code after success

```
  yield take(types.START_WALLET_REQUESTED);
  yield cancel(threads);
```

Why are they different? Can we reuse the same code?

src/sagas/wallet.js
```diff
@@ -531,10 +545,37 @@ export function* onWalletConnStateUpdate({ payload }) {
   yield put(isOnlineUpdate({ isOnline }));
 }
 
+export function* walletReloading() {
+  const wallet = yield select((state) => state.wallet);
+  const routerHistory = yield select((state) => state.routerHistory);
+
+  // Since we close the channel after a walletReady event is received,
```
pedroferreira1, line 552: Are we closing this channel in this method after the reload is completed?

src/screens/UnknownTokens.js
andreabadesso: No, since we only load the history after the user clicks the "Show history" button and we hide that button when the history is already loading, the loadHistory action will never be triggered and the UI will show "Loading history..." forever

src/sagas/wallet.js
andreabadesso: We only fetch the history when the user clicks the token to see more details, we don't pre-fetch token history

src/sagas/wallet.js
andreabadesso: > Should we cancel any thread here as well?

The wallet is not being initialized again, we are only re-downloading the token balances, so the `take` effect on the `startWallet` saga is still waiting for a `START_WALLET_REQUESTED` action to cancel the threads that still need to be running in background, so we don't have to do anything here

But I just noticed that we might have lost transactions during the downtime, so we should invalidate the token history for all tokens (except hathor that is already downloaded on the `loadTokens` saga)

Done in e03c9bfae103c50bf899995c28bf7b81a0e7be50

src/sagas/wallet.js
andreabadesso: No, the channel closes itself after the wallet is ready or erroed, so we don't have to

src/actions/index.js
andreabadesso: Added in bb823f3176d1a311548edc28bf12922d331e65ce

src/sagas/wallet.js
andreabadesso: By the way, this might happen on the wallet-mobile as well, we should also apply this commit to it

pedroferreira1: Ok, did you create a KTLO for that?

src/sagas/wallet.js
```diff
@@ -531,10 +546,47 @@ export function* onWalletConnStateUpdate({ payload }) {
   yield put(isOnlineUpdate({ isOnline }));
 }
 
+export function* walletReloading() {
+  console.log('Wallet reloading.');
```
pedroferreira1, line 550: Remove

pedroferreira1 (approved):

src/sagas/wallet.js
andreabadesso: Done in 2a6522688c8969fbd09a2c81ce76093d51ac53d1

src/sagas/wallet.js
andreabadesso: https://github.com/HathorNetwork/hathor-wallet-mobile/issues/200

PR #351, opened 2022-10-21
refactor: global modal

andreabadesso:
## Motivation

This PR is a refactor but also a bugfix for the wallet-desktop

Currently, since we are rendering modals on the screen components render method, every time a re-render happens (because of props being changed), the modals automatically close and in some cases, the modal backdrop stays there preventing the user from using the wallet.

The idea here is to have a global modal that is rendered on the top level of the component tree and a [Context](https://reactjs.org/docs/context.html) that can be accessed by all the components down the tree, so we can better control when to show and hide the modal

### Acceptance Criteria
- We should create a provider that will be able to render all modal types on demand
- We should only be able to render one modal at a time for now
- We should reduce the number of re-renders on the Wallet screen (0813b37dbcae7d577663c4d218861655fb955944)
- Fixes an error where the destroy authority action would not reload the token info (019b7bc3b2b378bd1b3ccccc9eeb6ebf13ba1e1c)

### Security Checklist


src/components/ModalEditToken.js
```diff
@@ -1,105 +0,0 @@
...
-              <button onClick={this.handleSave} type="button" className="btn btn-hathor">{t`Save`}</button>
-            </div>
-          </div>
-        </div>
-      </div>
-    )
-  }
-}
-
-export default ModalEditToken;
```
andreabadesso, line 105: This component was not being used anymore, so I removed it

src/App.js
```diff
@@ -271,13 +271,6 @@ const returnDefaultComponent = (Component, props) => {
           <Navigation {...props}/>
           <Component {...props} />
           <RequestErrorModal {...props} />
-           {/* At first I added this ModalAlert in the Version component (where I think it should be)
-             * however this component is inside the Navigation component, that has position fixed.
-             * The bootstrap modal does not work fine inside a wrapper with position fixed
-             * so the backdrop was being rendered above the modal.
-             * That's why the best solution was to add this modal here (so I can use in all screens that have the Navigation)
-             */}
-          <ModalAlert title='Software wallet warning' body={<SoftwareWalletWarningMessage />} buttonName='Ok' id='softwareWalletWarningModal' />
```
andreabadesso, line 280: No need for this anymore, the GlobalModal will be in charge of rendering the modals

src/components/ModalAddManyTokens.js
```diff
@@ -45,6 +45,7 @@ class ModalAddManyTokens extends React.Component {
   }
 
   componentDidMount = () => {
+    $('#addManyTokensModal').modal('show');
```
andreabadesso, line 48: The idea here is that the components won't need to use jquery directly, every time the component gets rendered it will show and when they are destroyed, they should hide

We also have a common prop `onClose` passed by the GlobalModal that should be called by the modals when the `hide.bs.modal` event is dispatched so the GlobalModal can properly remove it from the DOM

src/components/GlobalModal.js
```diff
@@ -0,0 +1,114 @@
...
+  [MODAL_TYPES.RESET_TOKEN_SIGNATURES]: ModalLedgerResetTokenSignatures,
+  [MODAL_TYPES.RESET_ALL_DATA]: ModalResetAllData,
+  [MODAL_TYPES.LEDGER_SIGN_TOKEN]: ModalLedgerSignToken,
+  [MODAL_TYPES.CONFIRM_TESTNET]: ModalConfirmTestnet,
+  [MODAL_TYPES.SEND_TX]: ModalSendTx,
+};
+
+export const GlobalModalContext = createContext(initialState);
+
+export const useGlobalModalContext = () => useContext(GlobalModalContext);
```
andreabadesso, line 54: This is not being used yet as we only have Class components, but would be used like:

```javascript
const globalModal = useGlobalModalContext();

globalModal.showModal(...);
```

on function components

src/screens/Wallet.js
```diff
@@ -75,61 +73,108 @@ class Wallet extends React.Component {
...
 
-    $('#wallet-div').on('show.bs.tab', 'a[data-toggle="tab"]', (e) => {
-      // On tab show we reload token data from the server for each component
-      if (e.target.id === 'token-tab') {
-        this.generalInfoRef.current.updateTokenInfo();
-      } else if (e.target.id === 'administrative-tab') {
-        this.administrativeRef.current.updateTokenInfo();
-      }
-    });
+    await this.updateTokenInfo(this.props.selectedToken);
```
pedroferreira1, line 92: We shouldn't execute this request and the next one if it's HTR.

src/screens/Wallet.js
```diff
@@ -75,61 +73,108 @@ class Wallet extends React.Component {
...
 
+    this.updateWalletInfo();
     // First time the screen is mounted we must also check if we should show administrative tab
     this.shouldShowAdministrativeTab(this.props.selectedToken);
   }
 
-  componentWillReceiveProps(nextProps) {
```
pedroferreira1, line 107: You moved this code to `componentDidUpdate`, and`componentDidUpdate` is not called in the initial render. Could this lead to bugs? Should we call this `getTokenSignature` piece of code also in the `componentDidMount`?

If yes, maybe we should create a method to wrap all the calls inside `componentDidMount` and `componentDidUpdate`

src/screens/Wallet.js
```diff
@@ -195,16 +254,23 @@ class Wallet extends React.Component {
...
+      hasTokenSignature: value,
+    });
+  }
+
   retryDownload = (e, tokenId) => {
     e.preventDefault();
     const balanceStatus = get(
-      this.props.tokensBalance,
-      `${tokenId}.status`,
+      this.props.tokenBalance,
```
pedroferreira1, line 267: Why did you change this? It looks like it's `tokensBalance`

```
  return {
    selectedToken: state.selectedToken,
    entireState: state,
    tokensHistory: state.tokensHistory,
    tokensBalance: state.tokensBalance,
    tokens: state.tokens,
    wallet: state.wallet,
    useWalletService: state.useWalletService,
  };
};
```

src/screens/SendTokens.js
```diff
@@ -132,11 +128,11 @@ class SendTokens extends React.Component {
...
       const tokenList = this.state.txTokens.filter(t => arg.data.includes(t.uid))
-      this.setState({
-        ledgerAlertModalTitle: t`Invalid custom tokens`,
-        ledgerAlertModalBody: this.renderAlertTokenList(tokenList),
-      });
-      $("#ledgerAlertModal").modal('show');
+      this.context.showModal(MODAL_TYPES.ALERT, {
+        id: 'ledgerAlertModal',
+        title: t`Invalid custom tokens`,
+        body: this.renderAlertTokenList(tokenList),
```
pedroferreira1, line 134: There's. a `buttonName={t`Close`}` missing.

src/screens/SendTokens.js
```diff
@@ -186,7 +182,16 @@ class SendTokens extends React.Component {
         arr.push(Buffer.from(signatures[i]));
       }
       this.sendTransaction.prepareTxFrom(arr);
-      this.setState({ ledgerStep: 1, ledgerModalTitle: t`Sending transaction` });
+      this.setState({
+        ledgerStep: 1,
+      }, () => {
+        this.context.showModal(MODAL_TYPES.ALERT, {
```
pedroferreira1, line 188: Why are you showing this modal here? I don't see anything showing this modal before.

src/screens/SendTokens.js
```diff
@@ -244,11 +249,12 @@ class SendTokens extends React.Component {
...
       // there are tokens without signatures, missingSigs
       // set tittle and content
-      this.setState({
-        ledgerAlertModalTitle: t`Unverified custom tokens`,
-        ledgerAlertModalBody: this.renderAlertTokenList(missingSigs),
+      this.context.showModal(MODAL_TYPES.ALERT, {
+        id: 'ledgerAlertModal',
+        title: t`Unverified custom tokens`,
+        body: this.renderAlertTokenList(missingSigs),
+        buttonName: 'Ok',
```
pedroferreira1, line 256: The `buttonName` before was 'Cancel', or not?

src/screens/SendTokens.js
```diff
@@ -382,17 +397,24 @@ class SendTokens extends React.Component {
...
         // limit is 10 custom tokens per tx
         const modalBody = <p>{t`Ledger has a limit of ${LEDGER_TX_CUSTOM_TOKEN_LIMIT} different tokens per transaction.`}</p>
-        this.setState({
-          ledgerAlertModalTitle: t`Token limit reached`,
-          ledgerAlertModalBody: modalBody,
+        this.context.showModal(MODAL_TYPES.ALERT, {
+          id: 'ledgerAlertModal',
+          title: t`Token limit reached`,
+          body: modalBody,
+          buttonName: 'Ok',
```
pedroferreira1, line 416: The same for the `buttonName`

src/components/ModalAddressQRCode.js
```diff
@@ -83,7 +89,7 @@ class ModalAddressQRCode extends React.Component {
           <div className="modal-content">
             <div className="modal-header">
               <h5 className="modal-title" id="exampleModalLabel">{t`Address to receive tokens`}</h5>
-              <button type="button" className="close" data-dismiss="modal" aria-label="Close">
+              <button type="button" className="close" onClick={this.close} data-dismiss="modal" aria-label="Close">
```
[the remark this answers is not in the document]
andreabadesso: It should be `this.props.onClose`, fixed in 708d6e29ab5bda9ac7800232fd669c2762351ef5

It was working though, because of the event and `data-dismiss`

src/components/ModalBackupWords.js
```diff
@@ -74,6 +75,8 @@ class ModalBackupWords extends React.Component {
       if (this.refs.password) {
         this.refs.password.value = '';
       }
+
+      this.props.onClose();
```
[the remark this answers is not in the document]
andreabadesso: But it is inside the  `on` event

src/components/ModalConfirmTestnet.js
```diff
@@ -0,0 +1,59 @@
...
+
+export default function ModalConfirmTestnet({ onClose, success }) {
+  const [confirmText, setConfirmText] = useState('');
+  const [testnetError, setTestnetError] = useState('');
+
+  useEffect(() => {
+    $('#modalConfirmTestnet').modal('show');
+    $('#modalConfirmTestnet').on('hidden.bs.modal', onClose);
+
+    return () => $('#modalConfirmTestnet').modal('hide');
```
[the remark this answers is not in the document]
andreabadesso: Yes we should, thanks!

708d6e29ab5bda9ac7800232fd669c2762351ef5

src/components/ModalConfirmTestnet.js
```diff
@@ -0,0 +1,59 @@
...
+                </span>
+              </div>
+            </div>
+          </div>
+          <div className="modal-footer">
+            <button onClick={confirmTestnetServer} type="button" className="btn btn-secondary">
+              {t`Connect to testnet`}
+            </button>
+            <button onClick={onClose} type="button" className="btn btn-hathor">
+              { t`Ok` }
```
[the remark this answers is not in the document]
andreabadesso: You are right, fixed in 307846368a0bb68b29c437195c90d68237a7ec64

src/components/ModalLedgerResetTokenSignatures.js
```diff
@@ -17,14 +17,15 @@ import { t } from 'ttag';
...
-  const modalId = "resetTokenSignatures";
+function ModalLedgerResetTokenSignatures({ onClose }) {
+  const modalId = 'resetTokenSignatures';
   const [waitingLedger, setWaitingLedger] = useState(false);
   const [showOk, setOk] = useState(false);
   const [errorMessage, setError] = useState(null);
 
   const closeModal = () => {
     $(`#${modalId}`).modal('hide');
+    onClose();
```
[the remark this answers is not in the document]
andreabadesso: Yes, it will be called twice

This is not a problem though, as the `onClose` that comes as prop from the GlobalModal is idempotent

src/screens/CustomTokens.js
```diff
@@ -35,27 +35,32 @@ const mapStateToProps = (state) => {
...
+    this.context.hideModal();
     this.refs.alertSuccess.show(1000);
   }
 
   /**
    * Triggered when user clicks to do the register a token, then opens the new token modal
    */
   registerTokenClicked = () => {
-    $('#addTokenModal').modal('show');
+    this.context.showModal('MODAL_ADD_TOKEN', {
```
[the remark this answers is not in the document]
andreabadesso: Thanks, fixed in 3b6d232079865c3dba33503bce9bfa70d6b6bb60

src/screens/SendTokens.js
andreabadesso: Fixed in b66bd3ad0a3a763a971a36fe9fb177bb040ed65f

src/screens/SendTokens.js
andreabadesso: Because we could change the modal body using `setState` before, now we can't 

So what this is doing is removing the old modal and replacing it by a new one with the `renderAlertBody` rendered with `ledgerStep` set to 1

src/screens/SendTokens.js
andreabadesso: We're both wrong, the buttonName should be `Closed`

Fixed in 4155a44519e9e2a4de667e7564f976f7753f02af

src/screens/SendTokens.js
andreabadesso: 4155a44519e9e2a4de667e7564f976f7753f02af fixes it as well

src/screens/Wallet.js
andreabadesso: You are right, fixed in 9aa96889c9eadbbbb2bc9515f39e00206731267d

src/screens/Wallet.js
andreabadesso: This is indeed a bug, fixed in 2f91a8ad674689fffb0ee4645163f8012c438a2e

src/components/ModalAddToken.js
```diff
@@ -50,15 +52,18 @@ class ModalAddToken extends React.Component {
...
       });
-    })
+    });
 
     $('#addTokenModal').on('shown.bs.modal', (e) => {
       this.refs.config.focus();
-    })
+    });
+
+    $('#addTokenModal').on('hidden.bs.modal', this.props.onClose);
```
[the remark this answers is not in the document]
andreabadesso: The difference between the two events are only that the `hidden.bs.modal` event waits for the transition to finish before being fired, the `hide.bs.modal` will fire right away

I added the `hidden.bs.modal` event listener to the `ModalAddManyTokens` to keep consistent in c6d98af11ca8082688343878a7dc28d90f1e27c3

src/components/ModalAddToken.js
```diff
@@ -145,7 +150,12 @@ class ModalAddToken extends React.Component {
           <div className="modal-content">
             <div className="modal-header">
               <h5 className="modal-title" id="exampleModalLabel">{t`Register a new token`}</h5>
-              <button type="button" className="close" data-dismiss="modal" aria-label="Close">
+              <button
+                type="button"
+                className="close"
+                data-dismiss="modal"
+                aria-label="Close"
+                onClick={this.props.onClose}>
```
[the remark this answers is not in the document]
andreabadesso: Yes it will and yes it is idempotent, it will just setState with null as the modalType, which will render null no matter how many times it was called

src/sagas/wallet.js
```diff
@@ -430,7 +430,15 @@ export function* handleTx(action) {
   }
 
   // find tokens affected by the transaction
-  const balances = yield call(wallet.getTxBalance.bind(wallet), tx);
+  const affectedTokens = new Set();
+
+  for (const output of tx.outputs) {
```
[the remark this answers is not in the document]
andreabadesso: I explored a little how we could do this and the problem right now is that we don't have a `isAddressMine` method working with the wallet-service facade, and even if we did, it would involve a request to a lambda to check

So I suggest we accept that we could be downloading more data for now and open a KTLO to handle this better later.

What do you think?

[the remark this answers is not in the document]
andreabadesso: https://github.com/HathorNetwork/hathor-wallet/issues/355

src/screens/Wallet.js
andreabadesso: I agree with the refactor, done in aad8472e8c7ac6edac7d81148f9276723bac2887

I also refactored it to reset the state when the user changes token and added a check to prevent a situation where the user changes tokens before the download completes causing the screen to show the administrative tab before the next request is complete

I also created an issue to refactor the mechanism for loading token data to cache locally instead of re-downloading every time the user changes token:

https://github.com/HathorNetwork/hathor-wallet/issues/356

src/components/TokenGeneralInfo.js
```diff
@@ -210,7 +181,7 @@ class TokenGeneralInfo extends React.Component {
       return (
         <p className="mt-2 mb-4">
           <strong>{t`Always show this token:`}</strong> {
-          this.state.alwaysShow
```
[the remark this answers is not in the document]
andreabadesso: Yes it is, I guess that since the `renderAlwaysShowTokenCheckbox` is not memoized, react will run through it and since it is returning a different value, it will re-render with the new value

src/App.js
```diff
@@ -271,13 +271,6 @@ const returnDefaultComponent = (Component, props) => {
           <Navigation {...props}/>
           <Component {...props} />
           <RequestErrorModal {...props} />
```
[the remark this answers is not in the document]
andreabadesso: I agree we should change this modal as well, but I would suggest doing this in a separate PR as the refactor is a little more complex than it looks since the `RequestErrorModal` is displayed from inside a `axiosInstance` which has no access to the `GlobalModalContext`. We will probably have to create some mechanism of displaying the `GlobalModal` from a redux action

Also, it is not affected by the same problem as it is rendered on `App.js`

src/components/ModalAddToken.js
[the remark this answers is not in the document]
pedroferreira1: I understand. But why did you add this on click here if we already have the event handler to do the same thing?

src/components/ModalBackupWords.js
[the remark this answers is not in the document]
pedroferreira1: Ops, my bad

src/App.js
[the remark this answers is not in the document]
pedroferreira1: Ok. Do we have any other component in this situation? And do we have a KTLO/issue to handle this?

src/sagas/wallet.js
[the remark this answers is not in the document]
pedroferreira1: Yeah, the wallet service implementation won't work 100% without this. We can do this in another PR but we should do it before rolling out

src/actions/index.js
```diff
@@ -294,14 +294,6 @@ export const walletReloadData = () => ({
   type: types.WALLET_RELOAD_DATA,
 });
 
-/**
- * routerHistory: The RouterHistory object to store
- */
-export const storeRouterHistory = (routerHistory) => ({
```
pedroferreira1, line 300: Why did you remove this?

pedroferreira1 (approved):

src/components/ModalAddToken.js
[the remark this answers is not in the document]
andreabadesso: That's a good question, I'm not sure

Removed it in bcb6d9bc6be85c52d313e9396c6f67b3afd8fd67

r4mmer (approved):

Issue #357, opened 2022-10-28
[Design] Automated tests


tuliomir:
:construction: Under development

This issue describes the research being done on automating tests for the Wallet Desktop application.

tuliomir:
Reviewers: @pedroferreira1 , @alexruzenhack

alexruzenhack:
> It is possible to test UI without Puppeteer, we can use [React Testing Library](https://testing-library.com/docs/react-testing-library/intro/) + Jest. Did you hear about this library? With this combo, the components can be rendered and navigated by DOM API and then asserted.

Now I'm reviewing the PoC and see you use the React Testing Library already. Nice!

pedroferreira1:
> As such, possible paths for this UI test would be:
>
> Update all react dependencies to use react-scripts@5 and try again

We've talked about trying to upgrade only the react-scripts and build a poc of one test suite using pupeteer, even if some parts of the app end up broken. Did you try to do it?

tuliomir:
> ...we can use [React Testing Library](https://testing-library.com/docs/react-testing-library/intro/) + Jest.

Added a _Guide Level Description_ section documenting that these libs are being used on the PoC. :+1:

pedroferreira1:
> To avoid unnecessary consumption of HTR tokens, all tests should, at the end, melt every custom token generated to ensure the original HTR balance will be preserved - exception made for the NFTs that have a necessary fee. Those funds should finally be sent back to the source wallet to provide for future tests.

Why don't we run the tests using the testnet, so we don't really need to worry about losing funds?

tuliomir:
> ...have you validated how to run those tests using ledger? 

Added considerations about the hardware wallet / ledger testing: they will be done in a future moment and are described on the _Future Possibilities_  section.

---

> ...we don't really need to worry about losing funds?

About the amount of funds, while we have a lot of funds available on the `testnet`, we also have to refill the funds on the source wallet manually. Since the tests are automated and may be executed very often, I think It's better to preserve these funds only to delay as much as possible the manual refilling process that will be done by the _on-call_.

What do you think?

---
Also, added a section on mocking and intercepting HTTP requests on the _Future Possibilities_ section to cover a comment on the PoC PR.

pedroferreira1:
[...]

PR #358, opened 2022-11-07
test: Testing PoC

tuliomir:
This PR implements a proof of concept for the testing design on #357.

It aims to create a minimum of file structure and code base for this application, which developers can use as a reference to implement other tests quickly.

New libs were added as `dev-dependencies` to improve the test quality by offering high level tools to interact with the generated component DOMs.


src/components/WalletAddress.js
```diff
@@ -36,7 +36,7 @@ const mapStateToProps = (state) => {
  *
  * @memberof Components
  */
-class WalletAddress extends React.Component {
+export class WalletAddress extends React.Component {
```
tuliomir, line 39: For testing, it is recommended to `export` the class in isolation, besides the already existing `export default` that also connects the class with the redux state.

package.json
```diff
@@ -62,7 +63,7 @@
     "start": "npm-run-all -p watch-css start-js",
     "build-js": "react-scripts build",
     "build": "npm-run-all build-css build-js",
-    "test": "react-scripts test --env=node",
+    "test": "react-scripts test",
```
tuliomir, line 66: Redundant parameter, as the `react-scripts` already have `env=node` as a default.

src/__tests__/components/WalletAddress.test.js
```diff
@@ -0,0 +1,127 @@
...
+  container = document.createElement("div");
+  document.body.appendChild(container);
+
+  // Mock the oldWallet object
+  oldWallet = hathorLib.wallet;
+  hathorLib.wallet = {
+    isHardwareWallet: () => false
+  }
+  hathorLib.wallet.isSoftwareWallet =
+    () => !hathorLib.wallet.isHardwareWallet()
```
alexruzenhack, line 34: This one could have been as follows, to take advantage of the object literal:
```
  // Mock the oldWallet object
  const falsy = () => false;
  oldWallet = hathorLib.wallet;
  hathorLib.wallet = {
    isHardwareWallet: falsy,
    isSoftwareWallet: () => !falsy()
  }
```
I believe it would improve readability as well.

alexruzenhack (approved):

src/__tests__/components/WalletAddress.test.js
tuliomir: This specific code would not work because we need to dynamically change the results on `isHardwareWallet` and `isSoftwareWallet`. Each is called on different places within the tested component, and they must return the same "answer" as to if the wallet is hardware or software.

The actual code of `isSoftwareWallet` is almost exactly this:
https://github.com/HathorNetwork/hathor-wallet-lib/blob/master/src/wallet.js#L2711-L2721

To improve legibility, I refactored it to be a "mini" class on on [58d6a76](https://github.com/HathorNetwork/hathor-wallet/pull/358/commits/58d6a763d77cd32852ab809f54fb09f00d4915fa). Do you think this makes it clearer what is the purpose of this mock?

alexruzenhack: Interessante! Ficou bem mais legivel. Nice!

tuliomir:
The PoC of this PR will be reviewed and merged via the #365 PR.

PR #363, opened 2022-11-21
feat: check addresses mine

andreabadesso:
Depends on https://github.com/HathorNetwork/hathor-wallet-lib/pull/429 and https://github.com/HathorNetwork/hathor-wallet-service/pull/305

### Acceptance Criteria
- We should use the new facade method `checkAddressesMIne` to decide whether to display the wallet address badge on TxData screen

### Security Checklist

### DIFF
--- a/src/components/TxData.js
+++ b/src/components/TxData.js
@@ -301,7 +301,10 @@ class TxData extends React.Component {
   }
 
   isAddressMine = (address) => {
-    return address in this.state.walletAddressesMap;
+    return (
+      address in this.state.walletAddressesMap
+      && this.state.walletAddressesMap[address]
+    );
   }
 
   render() {
--- a/src/components/TxData.js
+++ b/src/components/TxData.js
@@ -101,7 +101,7 @@ class TxData extends React.Component {
     const inputs = this.props.transaction.inputs;
     const outputs = this.props.transaction.outputs;
 
-    const addressWithoutUndecoded = (acc, io) => {
+    const getDecodedAddresses = (acc, io) => {
       const address = get(io, 'decoded.address');
 
       if (!address) {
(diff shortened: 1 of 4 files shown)


src/components/TxData.js
```diff
@@ -277,6 +300,10 @@ class TxData extends React.Component {
     this.props.history.push('/wallet/');
   }
 
+  isAddressMine = (address) => {
+    return address in this.state.walletAddressesMap;
```
tuliomir, line 304: I think this would return true even for entries that are not supposed to.

For example, for this [integration test scenario](https://github.com/HathorNetwork/hathor-wallet-lib/blob/a73a4634025c963ec97f13851d32774472c41681/__tests__/integration/hathorwallet_others.test.js#L240-L250), the walletAddressMap would be:
```js
{
  address1: true,
  address2: true,
  address3: true,
  'invalid-address': false,
}
```

This `isAddressMine` would return a _truthy_ response for a `invalid-address` parameter in this case, right?

andreabadesso: Thanks, fixed in f108d5b44a0cfb7c4ee80e45891c0a7ef6c3e149

src/components/TxData.js
```diff
@@ -93,6 +96,26 @@ class TxData extends React.Component {
     });
   }
 
+  fetchWalletAddressesMap = async () => {
+    const inputs = this.props.transaction.inputs;
+    const outputs = this.props.transaction.outputs;
+
+    const addresses = [
+      ...inputs.map((input) => input.decoded.address),
+      ...outputs.map((output) => output.decoded.address),
```
pedroferreira1, line 105: When we have a data script output (we don't have addresses on it), will this throw an error?

andreabadesso: It would, luckily I fixed it in a73171da5314901a014f1e417bb497f45e3779dd

src/components/TxData.js
```diff
@@ -93,6 +97,36 @@ class TxData extends React.Component {
     });
   }
 
+  fetchWalletAddressesMap = async () => {
+    const inputs = this.props.transaction.inputs;
+    const outputs = this.props.transaction.outputs;
+
+    const addressWithoutUndecoded = (acc, io) => {
```
pedroferreira1, line 104: Why don't you call it `getDecodedAddresses`?

andreabadesso: Good idea, done in 8886be7fd9049b1078bf4efa3dc8aee7d43a1d5d

pedroferreira1 (approved):

PR #365, opened 2022-11-29
test: PoC for automated tests

tuliomir:
# Summary
This PR implements a proof of Concept for automated tests on the desktop wallet, as per the design #357 .
- Unit tests using `jest`
- E2E tests using cypress

It aims to create a minimum of file structure and code base for this application, which developers can use as a reference to implement other tests quickly.

# Acceptance Criteria
- The application behavior should remain unchanged

# Lib changes

New libs were added as `dev-dependencies` to improve the test quality by offering high level tools to interact with the generated component DOMs and the e2e platform itself.

Testing Library added on commit [3908138](https://github.com/HathorNetwork/hathor-wallet/pull/365/commits/3908138e051a8f478141a9481933ee8dfd82ea46#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519)
- testing-library/jest-dom
- testing-library/react
- testing-library/user-event

Cypress and its Testing Library were added on commit [4c0c55a](https://github.com/HathorNetwork/hathor-wallet/pull/365/commits/4c0c55aa3318825ba25cb135f19de53e67cfde89)
- testing-library/cypress
- cypress
- eslint-plugin-cypress

All those changes were double-checked to only affect the `dev-dependencies` on the `package-lock.json` file.


src/components/WalletAddress.js
```diff
@@ -36,7 +36,7 @@ const mapStateToProps = (state) => {
  *
  * @memberof Components
  */
-class WalletAddress extends React.Component {
+export class WalletAddress extends React.Component {
```
tuliomir, line 39: For testing, it is recommended to `export` the class in isolation, besides the already existing `export default` that also connects the class with the redux state.

package.json
```diff
@@ -62,7 +63,8 @@
     "start": "npm-run-all -p watch-css start-js",
     "build-js": "react-scripts build",
     "build": "npm-run-all build-css build-js",
-    "test": "react-scripts test --env=node",
+    "test": "react-scripts test",
```
tuliomir, line 66: Redundant parameter, as the `react-scripts` already have `env=node` as a default.

package.json
```diff
@@ -45,7 +46,7 @@
     "react-paginate": "^6.3.2",
     "react-redux": "^7.1.0",
     "react-router-dom": "^5.0.1",
-    "react-scripts": "^3.0.1",
+    "react-scripts": "4.0.3",
```
tuliomir, line 49: It was possible to upgrade to the 2021 version of the react-scripts.
Further upgrading encountered errors described on the design document and were discarded.

alexruzenhack (approved):

package.json
pedroferreira1: I don't think we should update any version in this PR. We have a project to upgrade dependencies and this might break the building of executable, ledger integration, only in one specific OS. If we have plans to merge this PR, it should be only for the test foundation

cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "Using fixtures to represent data",
```
pedroferreira1, line 2: Do we need this file in the repository? If yes, should we adapt it to one example for us?

package.json
```diff
@@ -80,10 +82,16 @@
   "devDependencies": {
     "@sentry/browser": "^5.0.5",
     "@sentry/cli": "^1.40.0",
+    "@testing-library/cypress": "^8.0.7",
```
pedroferreira1, line 85: Have you checked the `package-lock.json` to see if all new lines there are dependencies/subdependencies from those new packages and added only for dev?

cypress/fixtures/example.json
tuliomir: At the moment we don't. It would require more complex tests already implemented to be used properly.

Removed this file on [6f10f0f](https://github.com/HathorNetwork/hathor-wallet/pull/365/commits/6f10f0ffa1e4397c6794058f13d7a9129768ee19) and added a section on the design's _Future Possibilities_ describing how to use this feature later, as it is a good test practice to have.

package.json
tuliomir: I double-checked the changes to the `package-lock.json and added a new section _Lib changes_ on this PR's description with my findings.

I believe all the added test dependencies are safe, and the other updates also.

package.json
tuliomir: It's a valid point. Reverted back the `react-scripts` to the `^3.0.1` version :+1:

pedroferreira1 (approved):

PR #371, opened 2022-12-09
refactor: global token bar

andreabadesso:
### Acceptance Criteria
- We should refactor the TokenBar component to be a Function component
- We should refactor the TokenBar to be global and render only on specific routes

### Security Checklist

### DIFF
--- a/src/components/TokenBar.js
+++ b/src/components/TokenBar.js
@@ -32,7 +32,7 @@ export default function TokenBar () {
   const tokens = useSelector((state) => state.tokens);
   const allTokens = useSelector((state) => state.allTokens);
   const selectedToken = useSelector((state) => state.selectedToken);
-  const tokensBalance= useSelector((state) => state.tokensBalance);
+  const tokensBalance = useSelector((state) => state.tokensBalance);
   const tokenMetadata = useSelector((state) => state.tokenMetadata);
 
   // If the current route is not in the whitelist, we should not render the tokenbar
--- a/src/components/TokenBar.js
+++ b/src/components/TokenBar.js
@@ -5,9 +5,9 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React from 'react';
+import React, { useState } from 'react';
 import { t } from 'ttag';
-import { connect } from "react-redux";
+import { useSelector, useDispatch } from 'react-redux';
 import { selectToken } from '../actions/index';
 import { get } from 'lodash';
 import hathorLib from '@hathor/wallet-lib';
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -24,6 +25,7 @@ ReactDOM.render(
   <Provider store={store}>
     <GlobalModal>
       <Router>
+        <TokenBar />
```
andreabadesso, line 28: This needs to be rendered under the `Router` provider so we can get the `location` and `history` objects from the `useLocation` and `useHistory` hooks

src/components/TokenBar.js
```diff
@@ -5,220 +5,204 @@
...
-    this.state = {
-      opened: false,
-      selected: 'HTR',
-    }
+import { useHistory, useLocation } from 'react-router-dom';
+
+// Routes that should display the TokenBar
+const ROUTE_WHITELIST = [
+  '/unknown_tokens/',
+  '/wallet/',
```
tuliomir, line 23: This won't match with `/wallet/send_tokens`, right?

tuliomir (approved):

src/components/TokenBar.js
```diff
@@ -5,220 +5,204 @@
...
+export default function TokenBar () {
+  const [opened, setOpened] = useState(false);
+  const dispatch = useDispatch();
+  const history = useHistory();
+  const location = useLocation();
+
+  const tokens = useSelector((state) => state.tokens);
+  const allTokens = useSelector((state) => state.allTokens);
+  const selectedToken = useSelector((state) => state.selectedToken);
+  const tokensBalance= useSelector((state) => state.tokensBalance);
```
pedroferreira1, line 35:
```suggestion
  const tokensBalance = useSelector((state) => state.tokensBalance);
```

pedroferreira1 (approved):

src/components/TokenBar.js
andreabadesso: It won't!

src/components/TokenBar.js
```diff
@@ -5,220 +5,204 @@
...
+  '/wallet/',
+];
+
+export default function TokenBar () {
+  const [opened, setOpened] = useState(false);
+  const dispatch = useDispatch();
+  const history = useHistory();
+  const location = useLocation();
+
+  const tokens = useSelector((state) => state.tokens);
```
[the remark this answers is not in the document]
andreabadesso: Yes! Done in c80e7e678367c921e152938981b51374c2da173e

PR #372, opened 2022-12-11
fix: restart after wallet service error using saga

andreabadesso:
### Acceptance Criteria
- We should handle errors in the `startWallet` saga, allowing the user to retry the request
- We should fix an error where the wallet would break (and reset) when initializing on the wallet-service
  - The `walletUtils.reloadElectron` method was broken, I removed it and added a mechanism for starting the wallet using the saga again

### Security Checklist

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -12,7 +12,7 @@ coverage:
     patch:
       default:
         # minimum coverage ratio that the commit must meet to be considered a success
-        target: 70%
+        target: 0%
         if_ci_failed: error
         only_pulls: true
 
--- a/src/sagas/helpers.js
+++ b/src/sagas/helpers.js
@@ -53,7 +53,7 @@ export function* dispatchAndWait(action, successAction, failureAction) {
  * Handles errors thrown from the main saga (started with call) by yielding
  * an action passed as a parameter
  *
- * @param action - The action to call
+ * @param saga - The saga to call (synchronously)
  * @param failureAction - Yields this action (with put) if the main action throws
  */
 export function errorHandler(saga, failureAction) {
--- a/src/sagas/helpers.js
+++ b/src/sagas/helpers.js
@@ -1,5 +1,5 @@
 import { get } from 'lodash';
-import { put, race, take } from 'redux-saga/effects';
+import { put, call, race, take } from 'redux-saga/effects';
 
 /**
  * Helper method to be used on take saga effect, will wait until an action
@@ -47,4 +47,22 @@ export function* dispatchAndWait(action, successAction, failureAction) {
     success: take(successAction),
     falure: take(failureAction),
   });
+}
+
+/**
+ * Handles errors thrown from the main saga (started with call) by yielding
+ * an action passed as a parameter
+ *
+ * @param action - The action to call
+ * @param failureAction - Yields this action (with put) if the main action throws
+ */
+export function errorHandler(saga, failureAction) {
+  return function* handleAction(action) {
+    try {
+      yield call(saga, action);
+    } catch (e) {
+      console.error(`Captured error handling ${action.type}`, e);
+      yield put(failureAction);
+    }
+  };
 }
--- a/src/sagas/wallet.js
+++ b/src/sagas/wallet.js
@@ -107,8 +107,12 @@ export function* startWallet(action) {
     dispatch = _dispatch;
   });
 
-  let wallet, connection;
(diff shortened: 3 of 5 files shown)


codecov.yml
```diff
@@ -12,7 +12,7 @@ coverage:
     patch:
       default:
         # minimum coverage ratio that the commit must meet to be considered a success
-        target: 70%
+        target: 0%
```
r4mmer, line 15: Should we remove the patch coverage target?
If we want to exclude some method from testing there are comments we can use so the coverage tool does not count those ([reference](https://github.com/gotwarlost/istanbul/blob/master/ignoring-code-for-coverage.md))

andreabadesso: We are still working on the design for the tests implementation on the wallet-desktop, we will eventually need the patch coverage target, so I think it's OK to leave it at 0 for now (since we really don't have any tests)

r4mmer (approved):

src/sagas/helpers.js
```diff
@@ -48,3 +48,21 @@ export function* dispatchAndWait(action, successAction, failureAction) {
     falure: take(failureAction),
   });
 }
+
+/**
+ * Handles errors thrown from the main saga (started with call) by yielding
+ * an action passed as a parameter
+ *
+ * @param action - The action to call
```
pedroferreira1, line 56: Is it missing the `saga` parameter?

src/sagas/wallet.js
```diff
@@ -188,6 +189,12 @@ export function* startWallet(action) {
   // Thread to listen for feature flags from Unleash
   const featureFlagsThread = yield fork(listenForFeatureFlags, featureFlags);
 
+  const threads = [
```
pedroferreira1, line 192: Maybe add some comments here?

src/sagas/wallet.js
```diff
@@ -215,20 +221,17 @@ export function* startWallet(action) {
       // the feature flag
       yield call(featureFlags.ignoreWalletServiceFlag.bind(featureFlags));
 
-      // Restart the whole bundle to make sure we clear all events
-      walletHelpers.reloadElectron();
+      // Cleanup all listeners
+      yield cancel(threads);
+
+      // Yield the same action so it will now load on the old facade
+      yield put(action);
```
pedroferreira1, line 228: Should we `return` here, so it can start this action again with the old facade?

src/sagas/wallet.js
```diff
@@ -261,10 +264,21 @@ export function* startWallet(action) {
...
+    start: take(types.START_WALLET_REQUESTED),
+    reload: take(types.RELOAD_WALLET_REQUESTED),
+  });
+
+  // We need to cancel threads on both reload and start
   yield cancel(threads);
+
+  if (reload) {
+    // Yield the same action again to reload the wallet
+    yield put(action);
```
pedroferreira1, line 280: Why do we need to reload again here? Haven't we done that above in line 228?

src/sagas/wallet.js
```diff
@@ -595,7 +609,7 @@ export function* walletReloading() {
 
 export function* saga() {
   yield all([
-    takeLatest(types.START_WALLET_REQUESTED, startWallet),
+    takeLatest(types.START_WALLET_REQUESTED, errorHandler(startWallet, startWalletFailed())),
```
pedroferreira1, line 612: Why the second parameter is already a called method and not like the first one?

pedroferreira1 (approved):

src/sagas/wallet.js
andreabadesso: We don't, since this is a generator, it will end after the last yield

src/sagas/wallet.js
andreabadesso: Done in 0b85cd1261820046ed98aac488e2326fe2fd611d

src/sagas/wallet.js
andreabadesso: Yes, but the idea here is to capture the `RELOAD_WALLET_REQUESTED` (that might be dispatched from anywhere in the code) and yield the same action that started the current `startWallet` saga

An example is on line 396

src/sagas/helpers.js
andreabadesso: No, but the parameter was wrong, fixed in 82c136916b8e9e120a079bfa83a532815dc05c4d

PR #379, opened 2023-01-05
fix: issues found during the QA on 05/01/2023

andreabadesso:
### Acceptance Criteria
* [x] Tx details don't work on `testnet` while using the wallet-service facade
	* The wallet-desktop was requesting the full node for the transaction details and it's always requesting the `mainnet` fullnode
* [x] The wallet-desktop ignores the inputted websocket URL on the `Server` component
	* It was rolling back the server and not calling `changeWsServer` on `confirmTestnet`
* [x] There was a race condition where the address was being set to undefined after a tx was sent
* [x] The wallet screen was requesting the `tokenDetails` endpoint for the hathor token
* [X] Added a retry mechanism for graphviz queries in TxData

Here is a video of the retry mechanism for the graphviz queries in the `TxData` screen:

https://user-images.githubusercontent.com/3586068/212172920-3881644b-dab7-4b18-b842-4a43c9bfb705.mov

Here is a video of the retry mechanism for the transaction query:

https://user-images.githubusercontent.com/3586068/212348289-b1c5e80e-9536-4d20-91a6-4fa61f5ae775.mov

Here is a video of the retry mechanism in the transaction weight query:

https://user-images.githubusercontent.com/3586068/212718491-c09c5433-3c07-4288-bec4-20054007cf79.mov

### Security Checklist

### DIFF
--- a/src/api/axiosInstance.js
+++ b/src/api/axiosInstance.js
@@ -9,6 +9,11 @@ import store from '../store/index';
 import { lastFailedRequest, updateRequestErrorStatusCode } from '../actions/index';
 import $ from 'jquery';
 import hathorLib from '@hathor/wallet-lib';
+import url from 'url';
+
+const URL_WHITELIST = [
+  '/v1a/graphviz/neighbours.dot',
+];
 
 /**
  * Create axios instance settings base URL and content type  
@@ -29,6 +34,13 @@ const createRequestInstance = (resolve, timeout) => {
   instance.interceptors.response.use((response) => {
     return response;
   }, (error) => {
+    const parsedUrl = url.parse(error.config.url);
+    // URLs in this whitelist won't be handled by the
+    // request error interceptor.
+    if (URL_WHITELIST.indexOf(parsedUrl.pathname) > -1) {
+      return;
+    }
+
     // Update status code of the last failed request in redux
     // Adding conditional because if the server forgets to send back the CORS
     // headers, error.response will be undefined
--- a/src/api/axiosInstance.js
+++ b/src/api/axiosInstance.js
@@ -14,6 +14,7 @@ import url from 'url';
 const URL_WHITELIST = [
   'graphviz/neighbours.dot',
   'transaction',
+  'transaction_acc_weight',
 ];
 
 /**
--- a/src/api/axiosInstance.js
+++ b/src/api/axiosInstance.js
@@ -12,7 +12,8 @@ import hathorLib from '@hathor/wallet-lib';
 import url from 'url';
 
 const URL_WHITELIST = [
-  '/v1a/graphviz/neighbours.dot',
+  'graphviz/neighbours.dot',
+  'transaction',
 ];
 
 /**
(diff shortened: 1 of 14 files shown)


src/screens/Server.js
```diff
@@ -203,6 +203,7 @@ class Server extends React.Component {
   confirmTestnetServer = () => {
     this.props.wallet.changeServer(this.state.selectedServer);
     if (this.props.useWalletService) {
+      this.props.wallet.changeWsServer(this.state.selectedWsServer);
```
pedroferreira1, line 206: This method here `changeWsServer` already sets the ws URL in the config, right? So we can remove the line below?

andreabadesso: Just tested and yes, thanks

Done in 6a459cb8abaf7245c0e76d816f84cd96b080e76c

pedroferreira1 (approved):

tuliomir (approved):

andreabadesso:
@pedroferreira1 @tuliomir 

I added a new retry mechanism for the graphviz query, and to do that, I had to whitelist the URL on the axios interceptor otherwise it would be captured there and show the global retry modal, which does not work with async/await requests.

Here are the commits:
a73ec5521f82de14ebac03d630108a85c2ee641e...f19fa5c55d7b2733e2e076a8d590f38d10b6787f

src/screens/TransactionDetail.js
```diff
@@ -49,45 +55,71 @@ class TransactionDetail extends React.Component {
...
+
+    try {
+      const data = await this.props.wallet.getFullTxById(this.props.match.params.id);
+
+      if (!hathorLib.helpers.isBlock(data.tx)) {
         this.getConfirmationData();
       }
-    }, (e) => {
+
+      this.setState({
```
pedroferreira1, line 100: If you use an id that does not exist in the full node, `data.success` will be `false`. In that case, we should show "Transaction not found" on the screen. Otherwise, in case of an API error, we can show "Error retrieving transaction, try again".

What do you think?

src/api/axiosInstance.js
```diff
@@ -9,6 +9,12 @@ import store from '../store/index';
 import { lastFailedRequest, updateRequestErrorStatusCode } from '../actions/index';
 import $ from 'jquery';
 import hathorLib from '@hathor/wallet-lib';
+import url from 'url';
+
+const URL_WHITELIST = [
+  'graphviz/neighbours.dot',
+  'transaction',
```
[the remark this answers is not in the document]
andreabadesso: Yes we should, thanks!

Fixed in fa6ea72e7a9c9878e5e177f912d09fc0d944f681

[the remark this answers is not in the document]
andreabadesso: Also added a video on the PR description

src/screens/TransactionDetail.js
andreabadesso: The problem here is that the lib handles the `success: false` by throwing an error, that is not `Transaction not found`

To do this, we need to refactor the lib to return a standard message on both facades so we can handle in the clients

Do you think we should do this now?

tuliomir (approved):

src/screens/TransactionDetail.js
andreabadesso: @pedroferreira1 Done in f56a6131d919ccc9fed577841a0798bdfe4effdf, which depends on the [new lib PR](https://github.com/HathorNetwork/hathor-wallet-lib/pull/456)

pedroferreira1 (approved):

PR #384, opened 2023-01-20
chore: Tools for Windows release with EV Code Signature from YubiKey

msbrogli:
### Acceptance Criteria
- Sign Windows executable using an EV Code Signing Certificate in a YubiKey.

### Security Checklist

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+#
+# Script to release new versions for Windows, macOS, and Linux.
+#
+# This script expects a file `./env` with exported envvars to sign executables
+# for Windows and macOS.
+#
+# The commented cmdlines were used in the past to fix different kind of issues.
+# They were left here because they might be useful for troubleshooting in the future.
+
+set -e  # Exit on any command failure.
+set -u  # Exit on unset variables.
+
+step1() {
+	echo step1: cleaning and installing dependencies
+	rm -rf node_modules/
+	npm install
+}
+
+step2() {
+	echo step2: transaction checks and building
+	make check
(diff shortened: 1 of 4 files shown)


release.sh
```diff
@@ -0,0 +1,46 @@
...
+	source ./env
+	npm run electron-pack-linux || exit -1
+	npm run electron-pack-win || exit -1
+	npm run electron-pack-mac || exit -1
+}
+
+step1
+step2
+step3
+step4
```
andreabadesso, line 46:
```suggestion
set -e # Exit on any command failure
set -u # Exit on unset variables

echo "Step 1: Cleaning and installing dependencies"
rm -rf node_modules/
npm install

echo "Step 2: Running checks and building"
make check
make i18n
rm -rf build/
npm run build

echo "Step 3: Packaging for different platforms"
source ./env
npm run electron-pack-linux
npm run electron-pack-win
npm run electron-pack-mac
```

andreabadesso (changes requested):

release.sh
pedroferreira1: I like the new format but as in the header

> The commented cmdlines were used in the past to fix different kind of issues.
> They were left here because they might be useful for troubleshooting in the future.

We need to understand if it's important to keep the commented lines (I think all the issues solved by them are already in the past)

pedroferreira1 (approved):

release.sh
msbrogli: I kept the methods so one can easily pick which parts to run when debugging an issue.

andreabadesso (approved):

PR #390, opened 2023-02-15
refactor: using new feature flag schema

andreabadesso:
This is the new configured feature flag in unleash: 

image

### Acceptance Criteria
- We should pass the network as the `stage` property when updating unleash context to make use of the [new wallet-desktop feature flag](https://unleash.internal.hathor.network/projects/default/features/wallet-service-desktop.rollout)

### Security Checklist

### DIFF
--- a/src/featureFlags.js
+++ b/src/featureFlags.js
@@ -4,6 +4,7 @@ import {
   UNLEASH_URL,
   UNLEASH_CLIENT_KEY,
   UNLEASH_POLLING_INTERVAL,
+  WALLET_SERVICE_FEATURE_TOGGLE,
 } from './constants';
 
 const IGNORE_WALLET_SERVICE_FLAG = 'featureFlags:ignoreWalletServiceFlag';
@@ -18,7 +19,7 @@ export class FeatureFlags extends events.EventEmitter {
 
     this.userId = userId;
     this.network = network;
-    this.walletServiceFlag = `wallet-service-wallet-desktop-${this.network}.rollout`;
+    this.walletServiceFlag = WALLET_SERVICE_FEATURE_TOGGLE;
     this.walletServiceEnabled = null;
     this.client = new UnleashClient({
       url: UNLEASH_URL,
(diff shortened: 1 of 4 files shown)


alexruzenhack (approved):

src/featureFlags.js
```diff
@@ -57,7 +58,10 @@ export class FeatureFlags extends events.EventEmitter {
       if (shouldIgnore) {
         return false;
       }
-      this.client.updateContext({ userId: this.userId });
+      this.client.updateContext({
+        userId: this.userId,
+        stage: this.network,
```
alexruzenhack, line 63: 💬 There is no actual stage in this project, right? Does it make sense to add?

pedroferreira1, line 63: Shouldn't is be:

```
        properties: {
          stage: this.network,
        },
```
?

And do we have different feature flags for mac/windows/linux?

andreabadesso: You are absolutely correct.

Done in 521280f49f007a9aff049ebcef0019ea9b43300d

Also sending the current OS

andreabadesso: Hmmm, I don't think so as it would have to be manually inserted

I refactored `stage` to `network` though as it makes more sense: c7376a5834224194e61c83db9bfffd19bc4b045e

src/utils/helpers.js
```diff
@@ -200,7 +200,46 @@ const helpers = {
...
+     * "Linux x86_64" - indicates the browser is running on a 64-bit version of Linux.
+     * "Android" - indicates the browser is running on an Android device.
+     * "iPhone" - indicates the browser is running on an iPhone.
+     * "iPad" - indicates the browser is running on an iPad.
+     * "iPod" - indicates the browser is running on an iPod.
+     *
+     * Since the wallet-desktop only runs in desktops, we can return 'other' for any other
+     * platform that is not windows, mac or linux
+     */
+    const platform = window.clientInformation['platform'].toLowerCase();
```
pedroferreira1, line 231: Does it make sense to use the `lodash` `get` here to protect from any future bug?

pedroferreira1 (approved):

alexruzenhack (approved):

src/utils/helpers.js
andreabadesso: Yes it does, done in 9682a49e9818aab589f304d3e09a5f39f664c282

Thanks

alexruzenhack (approved):

pedroferreira1 (approved):

PR #400, opened 2023-04-12
feat: Atomic Swap Create with backend

tuliomir:
### Acceptance Criteria
- Create atomic swaps using the service as a mediator
- Import atomic swap proposal data from the service
- Fetch all listened proposals at wallet loading time

_Note_: The `update` interactions with the backend require the new major version of the wallet lib, and will be implemented when it's available on the desktop wallet.

### Security Checklist

### Demonstrations
Navigation through proposals list, creation of new proposals and import of existing proposals

https://user-images.githubusercontent.com/1299409/214679752-bf5bc9de-0e79-49a3-be4b-3a64a752b15e.mp4


src/reducers/index.js
```diff
@@ -834,6 +887,55 @@ export const onNewProposalGeneration = (state, action) => {
...
+
+/**
+ * @param {string} action.proposalId
+ */
+export const onProposalCreateSuccess = (state, action) => {
+  const { proposalId } = action;
+
+  return {
+    ...state,
+    newProposal: {
```
pedroferreira1, line 915: When it's successful, shouldn't it be added to the `proposals` map and the `newProposal` cleaned?

src/screens/atomic-swap/ProposalList.js
```diff
@@ -63,7 +67,9 @@ export default function ProposalList (props) {
         for (const [proposalId, proposal] of Object.entries(proposals)) {
             const pId = proposalId;
             const password = proposal.password;
-            const pAmountTokens = proposal.data?.amountTokens; // Get this from proposal
+            const pAmountTokens = proposal.data
```
pedroferreira1, line 70: Why this is a string? And why don't you use lodash `get` to handle the default values?

src/screens/atomic-swap/ProposalList.js
```diff
@@ -12,14 +12,18 @@ import { useDispatch, useSelector } from "react-redux";
...
+    updatePersistentStorage
+} from "../../utils/atomicSwap";
 import walletUtil from "../../utils/wallet";
 import { GlobalModalContext, MODAL_TYPES } from '../../components/GlobalModal';
 
 export default function ProposalList (props) {
     const history = useHistory();
     const dispatch = useDispatch();
     const modalContext = useContext(GlobalModalContext);
+    const wallet = useSelector(state => state.wallet);
```
pedroferreira1, line 26: Why do you need it? I didn't see you using it.

src/screens/atomic-swap/ImportExisting.js
```diff
@@ -10,15 +10,20 @@ import Loading from "../../components/Loading";
...
+    updatePersistentStorage
+} from "../../utils/atomicSwap";
 
 export default function ImportExisting(props) {
     // Internal state
     const [isLoading, setIsLoading] = useState(false);
     const [proposalId, setProposalId] = useState('');
     const [password, setPassword] = useState('');
     const [errorMessage, setErrorMessage] = useState('');
+    const wallet = useSelector(state => state.wallet);
```
pedroferreira1, line 26: Why do you need it?

src/screens/atomic-swap/ProposalList.js
tuliomir: If the proposal does not have a `data` property it means it is still being fetched from the backend, so the _amount of tokens_ column shows nothing.

If the `data` property is available we can decide to show the amount of tokens. The `'0'` placeholder is for the remote possibility that the current state of this proposal is between the `proposalFetchSuccess` and the `proposalUpdated` with its calculated metadata.

It seemed unnecessary to use `get` in this circumstance. Do you agree?

src/reducers/index.js
```diff
@@ -111,6 +111,18 @@ const initialState = {
    * @type {Record<string,ReduxProposalData>}
    */
   proposals: {},
+  /**
+   * A temporary object containing the new proposal data while waiting for the unique identifier
```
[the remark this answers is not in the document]
tuliomir: The only place where the proposal identifier is generated is on the backend. Until then, we have only the `partialTx` and `password`, and thus cannot add them to the `proposals` map.

As soon as the `proposalId` is available it is added there, on the `NewSwap` screen.
https://github.com/HathorNetwork/hathor-wallet/blob/e576038725422d5790010b01d618b6328274f45c/src/screens/atomic-swap/NewSwap.js#L71-L78

src/reducers/index.js
tuliomir: #### About the cleanup:
I understand that leaving the `newProposal` data there is not a problem, since it is not persisted. A new `action` could be implemented for this cleanup, should we decide to do it.

#### About using the `proposals` map:
The full redux object has some calculations done on the `utils/atomicSwap:generateReduxObjFromProposal` method. We could implement all this business logic on the reducer, but I feel that this kind of code is better left to the action caller instead.

The best place to do this I found to be the `NewSwap` screen itself. That way it has the common _requested -> success_ actions, makes its processing there and updates the `allProposals` map there.

Another possibility I see is to store only the password on the `NewSwap` screen state and as soon as it receives the `proposalId`, call `importProposal(proposalId, password)` exactly as it happens on the `ImportExisting` screen. That makes the most reuse of code at the cost of an additional `GET` request on the backend.

Do you think this other solution would be better?

src/screens/atomic-swap/ImportExisting.js
tuliomir: I was not needed anymore, removed on [081005e](https://github.com/HathorNetwork/hathor-wallet/pull/400/commits/081005edfb48b8746d565a0361bb41382b098567)

src/screens/atomic-swap/ProposalList.js
tuliomir: I was not needed anymore, removed on [081005e](https://github.com/HathorNetwork/hathor-wallet/pull/400/commits/081005edfb48b8746d565a0361bb41382b098567)

src/reducers/index.js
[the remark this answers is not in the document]
pedroferreira1: With this you just can't have more than one proposal being created in the same time, but I guess it's fine for now.

src/reducers/index.js
pedroferreira1: About the cleanup:

Leaving the data there seems it was not completed yet, I feel we should do it.

Now the map:

I agree to continue what you are doing but having `allProposals[proposalId] = reduxObj;` is not the best way to update the redux state. You should create an action for that, don't you have one already?

src/screens/atomic-swap/ProposalList.js
tuliomir: If this value was kept as a number, there would be no way to differentiate a `0` as in "zero tokens for this proposal" from an `undefined` as in that there is no proposal data in memory.

The cleanest way to do this in code was to convert it to an empty string _or_ zero, when necessary.

Added an explicit conversion to string on [a8a7f29](https://github.com/HathorNetwork/hathor-wallet/pull/400/commits/a8a7f299285c617a572374c0b33734a1ec540f36) to make this clearer.

src/reducers/index.js
tuliomir: Added the `proposalCreateCleanup` action for cleaning the temporary object and updating the listened proposals map on new proposal creation.

pedroferreira1: The name `proposalCreateCleanup` is misleading in my opinion. It seems you are only cleaning something but you are actually adding to the proposals object as well. I would use something like `onProposalsCreateSuccess`, what do you think?

pedroferreira1: Actually I just saw you have an action similar to this one. Checking your available actions, you have `onNewProposalGeneration`, which adds a proposal to the proposals object. (I don't really like this name because it's also misleading but it's actually doing it).

Why don't you leave the cleanup only to clean the new proposal and you use this action that you already have to add the full proposal to the redux object?

src/screens/atomic-swap/NewSwap.js
```diff
@@ -30,23 +39,68 @@ export default function NewSwap (props) {
...
+
+        // Adding the new redux object to the listened proposals map
+        dispatch(proposalCreateCleanup(reduxObj));
+        setNewProposalId(proposalId);
+    }, [newProposal])
+
+    /**
+     * Listening to the proposals map when the cleanup was requested
+     */
     useEffect(() => {
```
pedroferreira1, line 88: Why do you need this second `useEffect`?

Can't you call 

```
            updatePersistentStorage(allProposals);
            navigateToProposal(proposalId);
```

In the first one after updating the proposals map in redux?

src/screens/atomic-swap/ProposalList.js
pedroferreira1: I don't understand why there would be no way to differentiate `0` from `undefined`. They are clearly different and easy to differentiate in the code, isn't it?

src/reducers/index.js
tuliomir: Changed the logic to import the newly created proposal on [905f382](https://github.com/HathorNetwork/hathor-wallet/pull/400/commits/905f382a83ed225689a1da88fee8bd015318a468). Aside from the disadvantage of having to do an extra request, this approach has some benefits:
- Single point of adding to the listened proposals: only through importing
- Single point of processing and calculation of redux object metadata: now only when importing
- Synchronized `timestamp` with the backend since the first iteration

This has left the `cleanup` with the scope of just cleaning the temporary object. The `onNewProposalGeneration` action was actually deleted, since it was not being used anymore.

src/screens/atomic-swap/NewSwap.js
tuliomir: There are two saga flows that happen for the creation of a swap:
- The first generates the identifier on the backend
- The second adds the proposal data to the listened proposals map

Added some more comments to this code for clarity.

src/reducers/index.js
```diff
@@ -6,14 +6,13 @@
...
-import { VERSION } from '../constants';
+import { FEATURE_TOGGLE_DEFAULTS, VERSION } from '../constants';
 import { types } from '../actions';
 import { get } from 'lodash';
 import { TOKEN_DOWNLOAD_STATUS } from '../sagas/tokens';
 import { WALLET_STATUS } from '../sagas/wallet';
-import { PROPOSAL_DOWNLOAD_STATUS } from '../utils/atomicSwap';
+import { PROPOSAL_CREATION_STATUS, PROPOSAL_DOWNLOAD_STATUS } from '../utils/atomicSwap';
 import { HATHOR_TOKEN_CONFIG } from "@hathor/wallet-lib/lib/constants";
-import { FEATURE_TOGGLE_DEFAULTS } from '../constants';
```
tuliomir, line 16: Unified the imports from the `constants` file on line 9

src/sagas/atomicSwap.js
```diff
@@ -76,14 +87,91 @@ function* fetchProposalData(action) {
...
+            { proposalId: proposalId })
+          ,
+          specificTypeAndPayload(
+            [ types.PROPOSAL_FETCH_FAILED ],
+            { proposalId: proposalId })
+          ,
+        );
+
+        // Error handling
+        if (fetchProposalResponse.falure) {
```
tuliomir, line 164: This is a typo on an existing code, will be fixed application-wide on the next PR

r4mmer (approved):

src/screens/atomic-swap/NewSwap.js
```diff
@@ -6,47 +6,43 @@
...
+        dispatch(proposalCreateRequested(newPartialTx, password));
     }
 
     useEffect(() => {
-        // If this proposal exists, navigate to it immediately
-        if (allProposals[proposalId]) {
-            updatePersistentStorage(allProposals);
-            navigateToProposal(proposalId);
+        // Shows the error message if it happens
+        if (lastFailedRequest && lastFailedRequest.message) {
```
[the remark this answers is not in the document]
tuliomir: Looking only at this screen, it would make sense to do so. But when looking at [the saga that is called](https://github.com/HathorNetwork/hathor-wallet/blob/7c750e74d7e2922b56f7d00f8e9fee4ff92c3e15/src/sagas/atomicSwap.js#L128) on click, it either:
- Throws and fills this `lastFailedRequest.message` field or;
- Navigates to the `EditSwap` screen

So, if there is no automatic navigation then we're sure to have an error message. Not having one _and_ staying in this screen would be a strange behavior anyway, and I think it wouldn't require a "This should never happen" kind of branch here, as it would also need to be unit tested later.

src/screens/atomic-swap/NewSwap.js
```diff
@@ -6,47 +6,43 @@
...
     const dispatch = useDispatch();
 
     const [errorMessage, setErrorMessage] = useState('');
 
-    const navigateToProposal = (pId) => {
-        history.replace(`/wallet/atomic_swap/proposal/${pId}`);
-    }
-
     const createClickHandler = () => {
         if (password.length < 3) {
```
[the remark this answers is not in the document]
tuliomir: Good catch! Added on [9478338](https://github.com/HathorNetwork/hathor-wallet/pull/400/commits/9478338140added925a6d733e36e1ba5ca81e2e4).

src/sagas/atomicSwap.js
```diff
@@ -76,14 +87,91 @@ function* fetchProposalData(action) {
...
+                message: t`An error occurred while creating this proposal.`
+            }))
+            return;
+        }
+
+        // Generate a minimal redux object to allow for the fetch request
+        yield(put(importProposal(proposalId, password)));
+
+        // Fetch the full proposal data from the backend
+        const fetchProposalResponse = yield call(
```
pedroferreira1, line 150: Why do you need to fetch data of a proposal that you've just created?

src/utils/atomicSwap.js
```diff
@@ -340,3 +333,69 @@ export const updatePersistentStorage = (proposalList) => {
...
+        sigStatus = PROPOSAL_SIGNATURE_STATUS.SIGNED;
+    }
+    rObj.data.signatureStatus = sigStatus;
+
+    return rObj;
+}
+
+/**
+ * Returns the Atomic Swap Service base URL,
+ * @param {string} network Network name for fetching the default base server url
```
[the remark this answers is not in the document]
tuliomir: Added on [7b3d362](https://github.com/HathorNetwork/hathor-wallet/pull/400/commits/7b3d36219cb382d33b17ac5fd8883cbb1a91e731)

src/sagas/atomicSwap.js
tuliomir: Refactored the saga in a way that the `createProposalOnBackend` method generates the full Redux object locally after receiving the newly generated `proposalId`.

src/utils/atomicSwap.js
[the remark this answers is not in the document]
pedroferreira1: Is the docstring correct? I thought we were returning the URL, that's why I asked for the return type.

src/utils/atomicSwap.js
```diff
@@ -340,3 +333,78 @@ export const updatePersistentStorage = (proposalList) => {
...
+
+/**
+ * Calculates the object that will be stored in Redux, containing all the helper values for a more
+ * enriched exhibition on the Atomic Swap screens
+ * @param {string} proposalId
+ * @param {string} password
+ * @param {string} partialTx
+ * @param {HathorWallet} wallet
+ * @param [options]
+ * @param {boolean} [options.signatures] Optional signatures data for the proposal
```
pedroferreira1, line 345: This is a string, right?

pedroferreira1 (approved):

src/utils/atomicSwap.js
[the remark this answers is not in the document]
tuliomir: Yes, the method retrieves the URL internally and uses it to set the lib base url. But never returns it to the caller, as it would have no use in that context.

src/utils/atomicSwap.js
tuliomir: You're right! Fixed on [f2b09c4](https://github.com/HathorNetwork/hathor-wallet/pull/400/commits/f2b09c49a2fbc9427b8adffcd7652d8d226269e4)

PR #407, opened 2023-08-08
feat: register HTR on start saga

r4mmer:
### Acceptance Criteria
- Include HTR in the registered tokens
- Update registered tokens when starting a wallet
- When first starting the wallet we should migrate the localStorage to the new keys, so we can start from the same point that the wallet would start on the previous version./
- mark the wallet as opened when we open the wallet
- Pass the storage as prop to ModalUnregisteredTokenInfo, so it can unregister the token.
- bump the version to 0.27.0-rc1

### Security Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   "productName": "Hathor Wallet",
   "description": "Light wallet for Hathor Network",
   "author": "Hathor Labs <contact@hathor.network> (https://hathor.network/)",
-  "version": "0.26.0",
+  "version": "0.27.0-rc1",
   "private": true,
   "dependencies": {
     "@hathor/wallet-lib": "^1.0.0-rc8",
--- a/src/App.js
+++ b/src/App.js
@@ -160,13 +160,14 @@ class Root extends React.Component {
  * Validate if version is allowed for the loaded wallet
  */
 const returnLoadedWalletComponent = (Component, props) => {
+  // For server screen we don't need to check version
+  const isServerScreen = props.match.path === '/server';
+
   // If was closed and is loaded we need to redirect to locked screen
-  if (LOCAL_STORE.wasClosed() || LOCAL_STORE.isLocked()) {
+  if ((!isServerScreen) && (LOCAL_STORE.wasClosed() || LOCAL_STORE.isLocked())) {
     return <Redirect to={{ pathname: '/locked/' }} />;
   }
 
-  // For server screen we don't need to check version
-  const isServerScreen = props.match.path === '/server';
   const reduxState = store.getState();
 
   // Check version
@@ -228,13 +229,14 @@ const returnStartedRoute = (Component, props, rest) => {
   // The wallet is already loaded
   const routeRequiresWalletToBeLoaded = rest.loaded;
   if (LOCAL_STORE.getWalletId()) {
+    const isServerScreen = props.match.path === '/server';
     // Wallet is locked, go to locked screen
-    if (LOCAL_STORE.isLocked()) {
+    if (LOCAL_STORE.isLocked() && !isServerScreen) {
       return <Redirect to={{pathname: '/locked/'}}/>;
     }
 
     // Route requires the wallet to be loaded, render it
-    if (routeRequiresWalletToBeLoaded) {
+    if (routeRequiresWalletToBeLoaded || isServerScreen) {
       return returnLoadedWalletComponent(Component, props);
     }
 
--- a/src/App.js
+++ b/src/App.js
@@ -161,6 +161,8 @@ class Root extends React.Component {
  */
 const returnLoadedWalletComponent = (Component, props) => {
   // For server screen we don't need to check version
+  // We also allow the server screen to be reached from the locked screen
+  // In the case of an unresponsive fullnode, which would block the wallet start
   const isServerScreen = props.match.path === '/server';
 
   // If was closed and is loaded we need to redirect to locked screen
@@ -229,6 +231,9 @@ const returnStartedRoute = (Component, props, rest) => {
   // The wallet is already loaded
   const routeRequiresWalletToBeLoaded = rest.loaded;
   if (LOCAL_STORE.getWalletId()) {
+    // The server screen is a special case since we allow the user to change the
+    // connected server in case of unresponsiveness, this should be allowed from
+    // the locked screen since the wallet would not be able to be started otherwise
     const isServerScreen = props.match.path === '/server';
     // Wallet is locked, go to locked screen
     if (LOCAL_STORE.isLocked() && !isServerScreen) {
--- a/src/App.js
+++ b/src/App.js
@@ -34,6 +34,7 @@ import WalletVersionError from './screens/WalletVersionError';
 import LoadWalletFailed from './screens/LoadWalletFailed';
 import version from './utils/version';
 import tokens from './utils/tokens';
+import storageUtils from './utils/storage';
 import { connect } from 'react-redux';
 import RequestErrorModal from './components/RequestError';
 import store from './store/index';
@@ -215,6 +216,9 @@ const returnStartedRoute = (Component, props, rest) => {
       return <Redirect to={{pathname: pathname.slice(3)}} />;
     }
   }
+
+  // Detect a previous instalation and migrate before continuing
+  storageUtils.migratePreviousLocalStorage();
 
   // The wallet was not yet started, go to Welcome
   if (!LOCAL_STORE.wasStarted()) {
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -51,6 +51,7 @@ export const types = {
   FEATURE_TOGGLE_INITIALIZED: 'FEATURE_TOGGLE_INITIALIZED',
   SET_FEATURE_TOGGLES: 'SET_FEATURE_TOGGLES',
   SET_UNLEASH_CLIENT: 'SET_UNLEASH_CLIENT',
+  MARK_TX_RECEIVED: 'MARK_TX_RECEIVED',
 };
 
 /**
@@ -154,7 +155,7 @@ export const resetWallet = () => ({ type: 'reset_wallet' });
  * tokens {Array} array of token uids the the wallet has
  * currentAddress {Object} The current unused address
  */
-export const loadWalletSuccess = (tokens, currentAddress) => ({ type: 'load_wallet_success', payload: { tokens, currentAddress } });
+export const loadWalletSuccess = (tokens, registeredTokens, currentAddress) => ({ type: 'load_wallet_success', payload: { tokens, registeredTokens, currentAddress } });
 
 /**
  * tx {Object} the new transaction
@@ -481,4 +482,9 @@ export const walletResetSuccess = () => ({
 
 export const walletReset = () => ({
   type: types.WALLET_RESET,
+});
+
+export const markTxReceived = (txId) => ({
+  type: types.MARK_TX_RECEIVED,
+  payload: txId,
 });
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -42,6 +42,7 @@ export const types = {
   WALLET_STATE_READY: 'WALLET_STATE_READY',
   WALLET_STATE_ERROR: 'WALLET_STATE_ERROR',
   WALLET_RELOAD_DATA: 'WALLET_RELOAD_DATA',
+  WALLET_CHANGE_STATE: 'WALLET_CHANGE_STATE',
   WALLET_RESET: 'WALLET_RESET',
   WALLET_RESET_SUCCESS: 'WALLET_RESET_SUCCESS',
   WALLET_REFRESH_SHARED_ADDRESS: 'WALLET_REFRESH_SHARED_ADDRESS',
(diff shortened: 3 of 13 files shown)


package.json
```diff
@@ -22,7 +22,7 @@
   "productName": "Hathor Wallet",
   "description": "Light wallet for Hathor Network",
   "author": "Hathor Labs <contact@hathor.network> (https://hathor.network/)",
-  "version": "0.26.0",
+  "version": "0.27.0-rc1",
```
r4mmer, line 25: Needed to bump so I could test the migration from a previous version.

src/App.js
```diff
@@ -224,13 +229,14 @@ const returnStartedRoute = (Component, props, rest) => {
   // The wallet is already loaded
   const routeRequiresWalletToBeLoaded = rest.loaded;
   if (LOCAL_STORE.getWalletId()) {
+    const isServerScreen = props.match.path === '/server';
     // Wallet is locked, go to locked screen
-    if (LOCAL_STORE.isLocked()) {
+    if (LOCAL_STORE.isLocked() && !isServerScreen) {
```
r4mmer, line 234: Even a locked wallet can access the server screen if the server is unresponsive and the user wants to change the server.

src/components/ModalUnregisteredTokenInfo.js
```diff
@@ -62,7 +67,7 @@ class ModalUnregisteredTokenInfo extends React.Component {
       );
 
       try {
-        const tokenData = await hathorLib.tokensUtils.validateTokenToAddByConfigurationString(configurationString, this.props.wallet.storage);
```
tuliomir, line 65: It seems this line was actually broken before your change :wrench:

src/sagas/wallet.js
```diff
@@ -456,11 +459,21 @@ export function* handleTx(action) {
   const tx = action.payload;
   const wallet = yield select((state) => state.wallet);
   const routerHistory = yield select((state) => state.routerHistory);
+  const receivedTxs = yield select((state) => state.receivedTxs);
 
   if (!wallet.isReady()) {
     return;
   }
 
+  if (receivedTxs.has(tx.tx_id)) {
```
tuliomir, line 468: Should we have a special treatment if the first event was a `NEW_TX` and the second one an `UPDATE_TX`? Or this would not be necessary here?

src/utils/storage.js
```diff
@@ -0,0 +1,56 @@
...
+
+const storageUtils = {
+  migratePreviousLocalStorage() {
+    const accessData = LOCAL_STORE.getItem('wallet:accessData');
+
+    if (!accessData) {
+      return;
+    }
+
+    if (LOCAL_STORE.getItem('wallet:started')) {
```
tuliomir, line 18: I feel it would be easier to read if we had a map with the old key name instead of the string literals here.

For example:
```js
if (LOCAL_STORE.getItem(OLD_STARTED_KEY) {
  LOCAL_STORE.setItem(
    STARTED_KEY,
    LOCAL_STORE.getItem(OLD_STARTED_KEY),
  );
  LOCAL_STORE.removeItem(OLD_STARTED_KEY);
}
```
The map could even be within the `migratePreviousLocalStorage` function, not to pollute the file scope with unnecessary data.

tuliomir (approved):

src/components/ModalUnregisteredTokenInfo.js
r4mmer: Yes, the wallet was not being passed as a prop

src/sagas/wallet.js
r4mmer: We can but I do not think we should right now.
The reason for this `receivedTxs` was because when we send a transaction we receive the same event multiple times for different reasons (once for every address in the tx, once when a block confirms the transaction, etc.)

These events would trigger multiple notifications for a each transaction and would request multiple history updates, which would make the UI of the wallet page "blink" while updating.
The main issue is that even though the event was being received as duplicate, our wallet-lib interprets it as 1 `new-tx` and multiple `update-tx` but this is not the case.

These issues will be resolved with the reliable integration eventually, but we currently should resolve the negative impact on the user experience (e.g. blinking pages, multiple notifications).

src/utils/storage.js
r4mmer: Done

src/sagas/wallet.js
```diff
@@ -346,13 +349,13 @@ export function* loadTokens() {
 
   // Fetch all tokens, including the ones that are not registered yet
   const allTokens = yield call([wallet, wallet.getTokens]);
-  const registeredTokens = yield getRegisteredTokensUids(wallet, true);
+  const registeredTokens = yield call(tokensUtils.getRegisteredTokens, wallet);
```
pedroferreira1, line 352: You were using the parameter `true` to exclude the default token but now you are using `false` (the default value), is this a bug or it was a bug before?

src/sagas/wallet.js
```diff
@@ -456,11 +459,21 @@ export function* handleTx(action) {
...
+  const receivedTxs = yield select((state) => state.receivedTxs);
 
   if (!wallet.isReady()) {
     return;
   }
 
+  if (receivedTxs.has(tx.tx_id)) {
+    // This transaction is being processed for a second time.
+    // This may happen due to an event being received twice.
+    // We will skip to avoid multiple notifications and UI changes.
```
pedroferreira1, line 471: When do you clean this `receivedTxs` redux state? Because if you receive an update in 2 minutes, it should update this tx, right? It was not a duplicated event.

src/utils/storage.js
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import LOCAL_STORE, { STARTED_KEY, LOCKED_KEY, CLOSED_KEY, IS_BACKUP_DONE_KEY, IS_HARDWARE_KEY } from '../storage';
+
+const storageUtils = {
+  /**
+   * Migrate the localStorage from a previous version to the current version.
```
pedroferreira1, line 12: Was it a bug before? It's not in the acceptance criteria

r4mmer: I tested the migration method, which migrates the access data to the current version, but during the QA the first screen to show when upgrading the version was the welcome screen since the `wallet:started` key was set to true but the new key was not.

So the issue was that we didn't even get to the locked screen to migrate the data.
Knowing this, I created this method to run during startup, a one time migration of keys which will take the user to the locked screen.

src/sagas/wallet.js
r4mmer: The event will still be treated by the wallet-lib, so we will not ignore it (even with the actual duplicate ones) this will just avoid sendind a "You've received a new transaction" notification which (even with updates) should be avoided.

The other effect would be that we do not request the history update, this request is just to update the redux state history with the history on the wallet-lib.
The only real effect I can think of are voided transactions, but we can add a special case for those and keep the skip on the notification, what do you think°

src/sagas/wallet.js
r4mmer: Before we were translating the behavior of the old wallet, which only requested the custom tokens since the first screen would load the history of the default token (or HTR).

We do not need to skip the default token (or HTR) here since we started using this list to set the registered tokens and we do not have any issue loading the HTR history before we load the first wallet screen.

src/components/ModalUnregisteredTokenInfo.js
```diff
@@ -62,7 +67,7 @@ class ModalUnregisteredTokenInfo extends React.Component {
...
+        const tokenData = await hathorLib.tokensUtils.validateTokenToAddByConfigurationString(configurationString, this.props.storage);
```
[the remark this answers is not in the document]
r4mmer: Done

src/App.js
```diff
@@ -216,6 +218,9 @@ const returnStartedRoute = (Component, props, rest) => {
     }
   }
 
+  // Detect a previous instalation and migrate before continuing
+  storageUtils.migratePreviousLocalStorage();
```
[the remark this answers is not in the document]
r4mmer: I explained this on this [comment](https://github.com/HathorNetwork/hathor-wallet/pull/407#discussion_r1291924877).

src/actions/index.js
```diff
@@ -482,3 +483,8 @@ export const walletResetSuccess = () => ({
 export const walletReset = () => ({
   type: types.WALLET_RESET,
 });
+
+export const markTxReceived = (txId) => ({
```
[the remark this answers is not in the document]
r4mmer: Done

src/screens/Wallet.js
```diff
@@ -459,8 +461,13 @@ class Wallet extends React.Component {
...
     const renderUnlockedWallet = () => {
       let template;
-      if (tokenHistory.status === TOKEN_DOWNLOAD_STATUS.LOADING
-          || tokenBalance.status === TOKEN_DOWNLOAD_STATUS.LOADING) {
+      /**
+       * If an oldStatus of either balance or history should be undefined this
+       * means that this is the first time we are loading it, so it should show
+       * the loading message.
+       */
+      if (tokenHistory.oldStatus === undefined
```
pedroferreira1, line 469: If the wallet loses connection with the full node and then reconnects. Will it reload the wallet and show the loading message? I think we should

src/screens/LockedWallet.js
```diff
@@ -80,6 +80,7 @@ class LockedWallet extends React.Component {
       }
 
       await LOCAL_STORE.handleDataMigration(pin);
+      LOCAL_STORE.open();
```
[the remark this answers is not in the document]
pedroferreira1: We should add a comment about this in the code, it's not clear

src/screens/Wallet.js
```diff
@@ -459,8 +460,12 @@ class Wallet extends React.Component {
...
     const renderUnlockedWallet = () => {
       let template;
-      if (tokenHistory.status === TOKEN_DOWNLOAD_STATUS.LOADING
-          || tokenBalance.status === TOKEN_DOWNLOAD_STATUS.LOADING) {
+      /**
+       * We only show the loading message if we are syncing the entire history
+       * This will happen on the first history load and if we lose connection
+       * to the fullnode.
+       */
+      if (this.props.walletState === hathorLib.HathorWallet.SYNCING) {
```
pedroferreira1, line 468: What are we doing if a new tx arrives in the wallet? I thought we were loading the history from the wallet lib. In that case, we should show the loading message as well, shouldn't we?

src/sagas/wallet.js
```diff
@@ -493,26 +528,33 @@ export function* handleTx(action) {
...
+export function* handleUpdateTx(action) {
+  const tx = action.payload;
+  const wallet = yield select((state) => state.wallet);
+
+  if (!wallet.isReady()) {
+    return;
+  }
+
+  // reload token history of affected tokens
+  // We always reload the history and balance
```
pedroferreira1, line 542: This is not true anymore, right?

src/sagas/wallet.js
```diff
@@ -493,26 +528,33 @@ export function* handleTx(action) {
...
+  const affectedTokens = yield call(handleTx, wallet, tx);
+  const stateTokens = yield select((state) => state.tokens);
+  const registeredTokens = stateTokens.map((token) => token.uid);
+  const txbalance = yield call([wallet, wallet.getTxBalance], tx);
 
-  yield put(sharedAddressUpdate({
-    lastSharedAddress: newAddress.address,
-    lastSharedIndex: newAddress.index,
-  }));
   // We should download the **balance** and **history** for every token involved
```
pedroferreira1, line 548: This is not true anymore, right?

src/reducers/index.js
```diff
@@ -1014,5 +1028,42 @@ const onWalletResetSuccess = (state) => ({
...
+  const { tx, tokenId, balance } = action.payload;
+  const tokenHistory = state.tokensHistory[tokenId];
+
+  if (!tokenHistory) {
+    return state;
+  }
+
+  for (const [index, histTx] of tokenHistory.data.entries()) {
+    if (histTx.tx_id === tx.tx_id) {
+      tokenHistory.data[index] = {
```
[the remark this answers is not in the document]
r4mmer: The transaction saved in the redux is mapped with `mapTokenHistory` from `src/utils/helpers.js` which does not have the `isAllAuthority` property, so i did not include it.
Another observation is that while `getTxHistoryFromWSTx` is used on the `onUpdateTx` the event that triggers the `onUpdateTx` is not used anymore.
I'll use `getTxHistoryFromWSTx` but we should remove unused events in the future.

src/screens/Wallet.js
r4mmer: The component used to show the transactions as a table will show a loading message when we send the event to load the history.

src/screens/Wallet.js
r4mmer: Changed to only show the entire wallet screen as loading if the wallet itself is in a loading state.

src/actions/index.js
```diff
@@ -482,3 +484,25 @@ export const walletResetSuccess = () => ({
 export const walletReset = () => ({
   type: types.WALLET_RESET,
 });
+
+/**
+ * Action to update the token history with this tx.
+ * @param {Object} tx New transaction to update in the token history.
+ * @param {string} tokenId token to update the history.
+ * @param {number} balance transaction balance of the token on the wallet.
+ * @returns {Object} action to mark tx as received.
```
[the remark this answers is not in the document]
r4mmer: Fixed

src/sagas/wallet.js
r4mmer: Removed

src/sagas/wallet.js
r4mmer: Removed

pedroferreira1 (approved):

PR #409, opened 2023-08-21
Feat/show authority operations

r4mmer:
### Motivation

Authority operations should show on the history list as "Authority" and not as transactions with zero balance.

The image below was taken from wallet version 0.26.0
before

This is how this PR will show a similar operation
after

### Acceptance Criteria
- Correctly map the history tx to a redux tx
- Show txs with only authority operations as "Authority" on the wallet screen

### Security Checklist

### DIFF
--- a/src/sagas/tokens.js
+++ b/src/sagas/tokens.js
@@ -201,8 +201,8 @@ function* fetchTokenHistory(action) {
       return;
     }
 
-    const response = yield call(wallet.getTxHistory.bind(wallet), { token_id: tokenId });
-    const data = response.map((txHistory) => helpers.mapTokenHistory(txHistory, tokenId));
+    const response = yield call([wallet, wallet.getTxHistory], { token_id: tokenId });
+    const data = yield call([helpers, helpers.mapTokenHistory], wallet, response, tokenId);
 
     yield put(tokenFetchHistorySuccess(tokenId, data));
   } catch (e) {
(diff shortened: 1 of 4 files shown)


src/utils/helpers.js
```diff
@@ -187,12 +187,62 @@ const helpers = {
...
+   * Check if the tx has only inputs and outputs that are authorities
+   *
+   * @param {Object} tx Transaction data
    *
-   * tx {Object} history data element
-   * tokenUid {String} token uid
+   * @return {boolean} If the tx has only authority
    */
-  mapTokenHistory(tx, tokenUid) {
+  isAllAuthority(tx) {
```
tuliomir, line 216: Could this method become `static`? It doesn't interact with the instance.

r4mmer: Well, this is not a class, so there is no `static` methods

src/sagas/tokens.js
```diff
@@ -201,8 +201,8 @@ function* fetchTokenHistory(action) {
       return;
     }
 
-    const response = yield call(wallet.getTxHistory.bind(wallet), { token_id: tokenId });
-    const data = response.map((txHistory) => helpers.mapTokenHistory(txHistory, tokenId));
+    const response = yield call([wallet, wallet.getTxHistory], { token_id: tokenId });
+    const data = yield call([helpers, helpers.mapTokenHistory], wallet, response, tokenId);
```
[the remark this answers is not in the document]
r4mmer: I also think it is a bit weird, but namespacing the utilities with an object made this a requirement.

tuliomir (approved):

andreabadesso (approved):

PR #411, opened 2023-08-31
fix: change server

r4mmer:
### Acceptance Criteria
- Use the correct methods to change the server and ws server
- When changing networks, clean the registered tokens as well.

### Security Checklist

### DIFF
--- a/src/reducers/index.js
+++ b/src/reducers/index.js
@@ -972,13 +972,16 @@ export const onStoreRouterHistory = (state, action) => {
   };
 };
 
-const onSetServerInfo = (state, action) => ({
-  ...state,
-  serverInfo: {
-    network: action.payload.network,
-    version: action.payload.version,
-  },
-});
+const onSetServerInfo = (state, action) => {
+  const network = action.payload.network.startsWith('testnet') ? 'testnet' : action.payload.network;
+  return {
+    ...state,
+    serverInfo: {
+      network,
+      version: action.payload.version,
+    },
+  }
+};
 
 const onFeatureToggleInitialized = (state) => ({
   ...state,
--- a/src/reducers/index.js
+++ b/src/reducers/index.js
@@ -973,11 +973,10 @@ export const onStoreRouterHistory = (state, action) => {
 };
 
 const onSetServerInfo = (state, action) => {
-  const network = action.payload.network.startsWith('testnet') ? 'testnet' : action.payload.network;
   return {
     ...state,
     serverInfo: {
-      network,
+      network: action.payload.network,
       version: action.payload.version,
     },
   }
--- a/src/sagas/wallet.js
+++ b/src/sagas/wallet.js
@@ -151,6 +151,8 @@ export function* startWallet(action) {
   if (useWalletService) {
     // Set urls for wallet service. If we have it on storage, use it, otherwise use defaults
     try {
+      // getWsServer can return null if not previously set
+      config.setWalletServiceBaseUrl(LOCAL_STORE.getWsServer());
       config.getWalletServiceBaseUrl();
     } catch(err) {
       if (err instanceof hathorErrors.GetWalletServiceUrlError) {
--- a/src/sagas/wallet.js
+++ b/src/sagas/wallet.js
@@ -268,6 +268,11 @@ export function* startWallet(action) {
       // takeLatest will stop running the generator if a new START_WALLET_REQUESTED
       // action is dispatched, but returning so the code is clearer
       return;
+    } else {
+      console.error(e);
+      // Return to locked screen when the wallet fails to start
+      LOCAL_STORE.lock();
+      routerHistory.push('/locked/');
     }
   }
 
--- a/src/sagas/wallet.js
+++ b/src/sagas/wallet.js
@@ -272,7 +272,8 @@ export function* startWallet(action) {
       console.error(e);
       // Return to locked screen when the wallet fails to start
       LOCAL_STORE.lock();
-      routerHistory.push('/locked/');
+      routerHistory.push('/');
+      return
     }
   }
 
--- a/src/sagas/wallet.js
+++ b/src/sagas/wallet.js
@@ -248,7 +248,7 @@ export function* startWallet(action) {
 
     if (serverInfo) {
       version = serverInfo.version;
-      networkName = serverInfo.network;
+      networkName = serverInfo.network && serverInfo.network.split('-')[0];
     }
 
     yield put(setServerInfo({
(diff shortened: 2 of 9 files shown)


src/utils/wallet.js
```diff
@@ -353,8 +353,14 @@ const wallet = {
...
    */
-  async changeServer(wallet, pin, routerHistory) {
-    await wallet.stop({ cleanStorage: false });
+  async changeServer(wallet, pin, routerHistory, networkChanged) {
+    // We only clean the storage if the network has changed
+    await wallet.stop({ cleanStorage: true, cleanAddresses: true });
+
+    if (networkChanged) {
+      // need to clean the storage, including registered tokens.
+      await wallet.storage.cleanStorage(true, true, true);
```
luislhl, line 362: Why don't we make the `wallet.stop`method above accept a new `cleanTokens` param instead of calling the `wallet.storage` directly?

alexruzenhack (approved):

src/sagas/wallet.js
```diff
@@ -266,6 +268,11 @@ export function* startWallet(action) {
       // takeLatest will stop running the generator if a new START_WALLET_REQUESTED
       // action is dispatched, but returning so the code is clearer
       return;
+    } else {
+      console.error(e);
+      // Return to locked screen when the wallet fails to start
+      LOCAL_STORE.lock();
+      routerHistory.push('/locked/');
```
r4mmer, line 275: Sometimes the wallet would be stuck on loading addresses, the error I got was that the configured network and server network did not match, it was raised during the start, but we already have a `try..catch` it just didn't handle the error, so the sagas would be waiting on a wallet that was not started.

The solution here is just to go back in the locked screen, but we can also add an error message in the future.
The reason I was having this error was because the change server error this PR solves would change the server but did not change the network so I do not expect to have this error again but i'm leaving this to make debugging future errors of this nature easier.

src/screens/TransactionDetail.js
```diff
@@ -94,6 +94,25 @@ class TransactionDetail extends React.Component {
...
+
+      for (const input of data.tx.inputs) {
+        if (!input.token) {
+          if (input.token_data === 0) {
+            input.token = hathorLib.constants.HATHOR_TOKEN_CONFIG.uid;
+          } else {
+            input.token = data.tx.tokens[(input.token_data & hathorLib.constants.TOKEN_INDEX_MASK) -1].uid;
+          }
+        }
+      }
```
r4mmer, line 115: `getFullTxById` fetches the transaction from the api, which uses has a slightly different structure from the websocket transactions and transactions received via history sync.
These lines just add the `token` uid where the helper will need it.

The issue here was that not having the `token` would make the transaction balance think that there was only a single token of uid `undefined`.

src/utils/wallet.js
r4mmer: Stopping a wallet should not remove registered tokens, this is necessary here because we will be moving to another network where these tokens do not exist and leaving the tokens on the storage would make the wallet screen show tokens that do not exist on the connected network.

src/reducers/index.js
```diff
@@ -972,13 +972,16 @@ export const onStoreRouterHistory = (state, action) => {
...
 
-const onSetServerInfo = (state, action) => ({
-  ...state,
-  serverInfo: {
-    network: action.payload.network,
-    version: action.payload.version,
-  },
-});
+const onSetServerInfo = (state, action) => {
+  const network = action.payload.network.startsWith('testnet') ? 'testnet' : action.payload.network;
```
luislhl, line 976: In which cases does the network come with a suffix? And why does it happen only on testnet?

Maybe a comment explaining this is a good idea to help future readers.

src/screens/TransactionDetail.js
luislhl: I'm not sure I get this: would we have some problem if we implemented this in the lib itself, so the `getFullTxById` method always returns with the right token?

I mean, when using this method in other wallets, won't they need to perform a similar treatment to this response?

src/screens/TransactionDetail.js
```diff
@@ -94,6 +94,25 @@ class TransactionDetail extends React.Component {
 
     try {
       const data = await this.props.wallet.getFullTxById(this.props.match.params.id);
+      for (const output of data.tx.outputs) {
```
luislhl, line 97: Should we have unit tests for this?

As a suggestion, it might be a good idea to avoid adding too much business logic here so we have a clearer separation between presentation layer components and business logic.

It could also facilitate unit tests for the business logic, since by putting it here you would have to mount the component and all its dependencies in order to test this logic.

Does it make sense?

src/screens/TransactionDetail.js
r4mmer: You are absolutely right, the only other place this is used is on the decode api of the headless and when it is used the token would be missing. I created a PR to fix this on the headless https://github.com/HathorNetwork/hathor-wallet-headless/pull/342

The main problem in my opinion is that the same transaction can have multiple structures depending on which api is used to fetch it from the fullnode or if it comes from the websocket.
Maybe the best solution would be to normalize the transaction model sent by the fullnode and the wallet-lib would be more stable.
Since solving this on the wallets is the simpler approach I think we should stick with it for now.
What do you think?

src/screens/TransactionDetail.js
r4mmer: Maybe, this only affects the UI, but it is a pretty important piece of UI on the transaction data.
But I don't think it's in the scope of this PR, since we should think of how to structure the UI and business logic tests with electron.

src/reducers/index.js
luislhl: Oh, ok! I forgot for a moment that we have this suffix on the testnets.

src/screens/TransactionDetail.js
luislhl: That's ok for me. But if you think this is important, we should create an issue or KTLO, so we don't forget.

src/screens/TransactionDetail.js
luislhl: I think it's ok to stick to the simpler approach. But we should have an issue to implement the best solution in the future.

src/reducers/index.js
r4mmer: I removed this change and made this treatment where we create the network name, since it is how we used to do it.

src/utils/wallet.js
r4mmer: Created this issue to track this change on the wallet-lib https://github.com/HathorNetwork/hathor-wallet-lib/issues/575

luislhl (approved):

PR #413, opened 2023-09-21
fix: wait for new address

r4mmer:
### Acceptance Criteria
Fix issues described [here](https://github.com/HathorNetwork/internal-issues/issues/185#issuecomment-1729906329) 
- await the promise to generate a new address
- Call the version api from the configured server when first starting the wallet.
- Fix the passphrase operations

### Security Checklist

### DIFF
--- a/src/components/WalletAddress.js
+++ b/src/components/WalletAddress.js
@@ -69,9 +69,9 @@ export class WalletAddress extends React.Component {
    *
    * @param {Object} e Event emitted by the link clicked
    */
-  generateNewAddress = (e) => {
+  generateNewAddress = async (e) => {
     e.preventDefault();
-    const address = this.props.wallet.getNextAddress();
+    const address = await this.props.wallet.getNextAddress();
 
     if (address.address === this.props.lastSharedAddress) {
       this.alertErrorRef.current.show(3000);


src/App.js
```diff
@@ -170,31 +171,38 @@ const returnLoadedWalletComponent = (Component, props) => {
...
+
   const reduxState = store.getState();
 
   // Check version
-  if (reduxState.isVersionAllowed === undefined && !isServerScreen) {
+  if (reduxState.isVersionAllowed === undefined) {
     // We already handle all js errors in general and open an error modal to the user
     // so there is no need to catch the promise error below
-    version.checkApiVersion(reduxState.wallet);
+    version.checkApiVersion();
```
r4mmer, line 186: removed the need for a wallet, since we need to call it from the locked screen as well

src/App.js
```diff
@@ -170,31 +171,38 @@ const returnLoadedWalletComponent = (Component, props) => {
...
     return <Redirect to={{
       pathname: '/loading_addresses/',
       state: {path: props.match.url},
       waitVersionCheck: true
     }} />;
-  } else if (reduxState.isVersionAllowed === false && !isServerScreen) {
+  }
+  if (reduxState.isVersionAllowed === false) {
     return <VersionError {...props} />;
-  } else {
```
r4mmer, line 187: Flattening the code for simplicity

src/components/WalletAddress.js
```diff
@@ -69,9 +69,9 @@ export class WalletAddress extends React.Component {
    *
    * @param {Object} e Event emitted by the link clicked
    */
-  generateNewAddress = (e) => {
+  generateNewAddress = async (e) => {
     e.preventDefault();
-    const address = this.props.wallet.getNextAddress();
+    const address = await this.props.wallet.getNextAddress();
```
r4mmer, line 74: This is a promise, so it should be awaited

src/reducers/index.js
```diff
@@ -372,14 +372,13 @@ const onUpdateLoadedData = (state, action) => ({
 });
 
 const onCleanData = (state) => {
-  if (state.wallet) {
-    state.wallet.stop();
```
r4mmer, line 376: Stopping the wallet emits a change of state, which emits an event to update the wallet state on redux.
Since stop is sync, we receive an error that we cannot emit an action from a reducer.
Also the wallet does not need to be stopped here.

src/reducers/index.js
```diff
@@ -372,14 +372,13 @@ const onUpdateLoadedData = (state, action) => ({
...
-  }
-
   return Object.assign({}, initialState, {
     isVersionAllowed: state.isVersionAllowed,
     loadingAddresses: state.loadingAddresses,
     ledgerWasClosed: state.ledgerWasClosed,
+    // Keep the unleashClient as it should continue running
+    unleashClient: state.unleashClient,
+    featureTogglesInitialized: state.featureTogglesInitialized,
```
r4mmer, line 381: If not persisted, the `unleashClient` and `featureTogglesInitialized` may lock the wallet in an infinite wait period on the next start, since we wait for an event to set the featureTogglesInitialized but it will never come.

src/sagas/helpers.js
```diff
@@ -20,6 +20,7 @@ export function* waitForFeatureToggleInitialization() {
   const featureTogglesInitialized = yield select((state) => state.featureTogglesInitialized);
 
   if (!featureTogglesInitialized) {
+    console.log('Feature toggle is not initialized, will wait indefinetely until it is.');
```
r4mmer, line 23: Just to make it easier to debug if the wallet enters this waiting period

src/storage.js
```diff
@@ -89,8 +89,9 @@ export class LocalStorageStore {
   cleanWallet() {
     this.removeItem('wallet:id');
     this.removeItem(IS_HARDWARE_KEY);
-    this.removeItem(STARTED_KEY);
-    this.removeItem(LOCKED_KEY);
+    // this.removeItem(STARTED_KEY);
+    // this.removeItem(LOCKED_KEY);
```
r4mmer, line 93: Removing these will make the wallet start from the welcome screen

src/storage.js
```diff
@@ -104,13 +105,14 @@ export class LocalStorageStore {
     }
   }
 
-  async initStorage(seed, password, pin) {
+  async initStorage(seed, password, pin, passphrase='') {
```
r4mmer, line 108: Allow initializing the storage with passphrase

src/utils/version.js
```diff
@@ -19,6 +19,32 @@ import LOCAL_STORE from '../storage';
  */
 
 const version = {
+
+  /**
+   * Get version data from HathorLib
+   * @returns {Promise} Promise that resolves after getting the version
+   * @memberof Version
+   */
+  async getVersionData() {
```
r4mmer, line 28: Get version data from api instead of wallet facade to allow calling from the locked screen

andreabadesso: Yes, but it will only work in the old facade, so this is breaking the wallet service

src/App.js
```diff
@@ -286,9 +294,12 @@ const returnDefaultComponent = (Component, props) => {
   const reduxState = store.getState();
 
   if (reduxState.isVersionAllowed === undefined) {
+
+    helpers.loadStorageState();
+
     // We already handle all js errors in general and open an error modal to the user
     // so there is no need to catch the promise error below
-    version.checkApiVersion(reduxState.wallet);
```
andreabadesso, line 291: How will this work on the wallet service?

andreabadesso (changes requested):
.

src/App.js
r4mmer: After some though I reversed to how it was originally, the `loadStorageState` is enough to load the server url into the lib config.
And the "check version" will run right before going into loading_addresses.

andreabadesso (changes requested):

src/reducers/index.js
r4mmer: I agree, the reason for this is that some variables in store are loaded by sagas that run only on startup and would not get reloaded after a wallet reset and in this case, adding a passphrase.

I'll create a ktlo for this.

src/utils/version.js
```diff
@@ -22,7 +22,7 @@ const version = {
   /**
    * Checks if the API version of the server the wallet is connected is valid for this wallet version
    *
-   * @param {HathorWallet | HathorWalletServiceWallet} wallet - Current wallet instance
+   * @param {HathorWallet | HathorWalletServiceWallet | undefined} wallet - Current wallet instance
```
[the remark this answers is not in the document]
r4mmer: We always pass the redux wallet even when it is not defined, this is why we need the `if (!wallet)` at the start.

andreabadesso (approved):

src/utils/wallet.js
```diff
@@ -131,6 +131,7 @@ const wallet = {
     try {
       walletUtils.wordsValid(words);
     } catch(e) {
+      console.error(e);
```
tuliomir, line 134: Do we still need this console?
If so, maybe we should move it to the closure on 138/139, to avoid presenting the full error in case of `InvalidWords`.

r4mmer: We don't need this one, but I left it to make debugging easier.
Although having invalid words is expected in one case (importing a seed from user input), I think we should still log this error.

src/utils/wallet.js
```diff
@@ -446,16 +448,20 @@ const wallet = {
   async cleanWallet(wallet) {
     await wallet.storage.cleanStorage(true, true);
```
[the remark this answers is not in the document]
r4mmer: The `cleanData` action was raising an issue, it would stop the redux wallet if it existed and clean the redux data.
By doing so, it would trigger another action and this would be an error since we cannot emit actions from reducers.

The fix for this was just making the wallet.stop call on the method that emitted the `cleanData` action which is the `cleanWalletRedux`.

tuliomir (approved):

PR #418, opened 2023-10-11
feat: changes from v0.27.0-rc3 QA

r4mmer:
### Acceptance Criteria
- When handling migration from a wallet we should load the wallet on the same server and network it was previously configured
- When testing if the ledger version is compatible with custom tokens we should use the correct version

### Security Checklist

### DIFF
--- a/src/storage.js
+++ b/src/storage.js
@@ -82,6 +82,10 @@ export class LocalStorageStore {
     return this.getItem('wallet:id');
   }
 
+  isLoadedSync() {
+    return (!!this.getWalletId()) || (!!this.getItem('wallet:accessData'))
+  }
+
   setWalletId(walletId) {
     this.setItem('wallet:id', walletId);
   }
@@ -244,6 +248,9 @@ export class LocalStorageStore {
    */
   async handleMigrationOldRegisteredTokens(storage) {
     const oldTokens = this.getItem('wallet:tokens');
+    if (!oldTokens) {
+      return;
+    }
     for (const token of oldTokens) {
       await storage.registerToken(token);
     }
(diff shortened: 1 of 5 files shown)


src/utils/storage.js
```diff
@@ -22,6 +24,21 @@ const storageUtils = {
...
+      helpers.updateNetwork(LOCAL_STORE.getItem(oldStorageKeys.OLD_NETWORK_KEY));
+      LOCAL_STORE.removeItem(oldStorageKeys.OLD_NETWORK_KEY);
+    }
+
+    if (LOCAL_STORE.getItem(oldStorageKeys.OLD_SERVER_KEY)) {
+      LOCAL_STORE.setItem(
+        SERVER_KEY,
+        LOCAL_STORE.getItem(oldStorageKeys.OLD_SERVER_KEY),
+      )
+      LOCAL_STORE.removeItem(oldStorageKeys.OLD_SERVER_KEY);
```
r4mmer, line 41: This will load the network and server from the previous version into the new version and update the redux to show the correct network

src/utils/version.js
```diff
@@ -83,7 +83,7 @@ const version = {
    * @inner
    */
   isLedgerCustomTokenAllowed() {
-    const version = LOCAL_STORE.getWalletVersion();
```
r4mmer, line 86: We were using the wallet version instead of the ledger version, which would always fail since the ledger version is ahead of the wallet

src/storage.js
```diff
@@ -244,6 +248,9 @@ export class LocalStorageStore {
    */
   async handleMigrationOldRegisteredTokens(storage) {
     const oldTokens = this.getItem('wallet:tokens');
+    if (!oldTokens) {
+      return;
+    }
     for (const token of oldTokens) {
       await storage.registerToken(token);
     }
```
r4mmer, line 256: When migrating a previous wallet that only had HTR or was empty this would fail since we cannot create a for..of loop for null or undefined

andreabadesso (approved):

src/storage.js
```diff
@@ -82,6 +82,10 @@ export class LocalStorageStore {
     return this.getItem('wallet:id');
   }
 
+  isLoadedSync() {
+    return (!!this.getWalletId()) || (!!this.getItem('wallet:accessData'))
+  }
```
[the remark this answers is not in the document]
pedroferreira1: Can we add this as comment in the code, so we know we can remove `accessData` check later?

pedroferreira1 (approved):

PR #420, opened 2023-10-20
chore: fix ledger issues found during QA v0.27.0-rc4

r4mmer:
[qa issue](https://github.com/HathorNetwork/internal-issues/issues/191)

### Acceptance Criteria

- `getSignatures` util for ledger expects a wallet instance but was receiving a storage instance.
- Clean only needed keys on storage to redirect user to wallet_type instead of welcome when opening the wallet.
- await `preparingTxfrom` so that the transaction is ready when we push it to the network.

### Security Checklist

### DIFF
--- a/src/components/Version.js
+++ b/src/components/Version.js
@@ -8,7 +8,7 @@
 import React from 'react';
 import { t } from 'ttag';
 import $ from 'jquery';
-import { VERSION, LEDGER_ENABLED } from '../constants';
+import { VERSION } from '../constants';
 import { GlobalModalContext, MODAL_TYPES } from './GlobalModal';
 import SoftwareWalletWarningMessage from './SoftwareWalletWarningMessage';
 import LOCAL_STORE from '../storage';
@@ -25,16 +25,14 @@ class Version extends React.Component {
    * If it's software wallet show modal warning
    */
   walletTypeClicked = () => {
-    if (LEDGER_ENABLED) {
-      if (!LOCAL_STORE.isHardwareWallet()) {
-        $('#softwareWalletWarningModal').modal('show');
-        this.context.showModal(MODAL_TYPES.ALERT, {
-          body: <SoftwareWalletWarningMessage />,
-          buttonName: 'Ok',
-          id: 'softwareWalletWarningModal',
-          title: 'Software wallet warning',
-        });
-      }
+    if (!LOCAL_STORE.isHardwareWallet()) {
+      $('#softwareWalletWarningModal').modal('show');
+      this.context.showModal(MODAL_TYPES.ALERT, {
+        body: <SoftwareWalletWarningMessage />,
+        buttonName: 'Ok',
+        id: 'softwareWalletWarningModal',
+        title: 'Software wallet warning',
+      });
     }
   }
 
--- a/src/components/Version.js
+++ b/src/components/Version.js
@@ -26,7 +26,6 @@ class Version extends React.Component {
    */
   walletTypeClicked = () => {
     if (!LOCAL_STORE.isHardwareWallet()) {
-      $('#softwareWalletWarningModal').modal('show');
       this.context.showModal(MODAL_TYPES.ALERT, {
         body: <SoftwareWalletWarningMessage />,
         buttonName: 'Ok',
--- a/src/constants.js
+++ b/src/constants.js
@@ -132,11 +132,6 @@ if (window.require) {
 export const IPC_RENDERER = ipcRenderer;
 
 /**
- * Flag to hide/show elements after ledger integration is done
- */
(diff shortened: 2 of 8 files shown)


src/constants.js
```diff
@@ -131,11 +131,6 @@ if (window.require) {
  */
 export const IPC_RENDERER = ipcRenderer;
 
-/**
- * Flag to hide/show elements after ledger integration is done
- */
-export const LEDGER_ENABLED = true;
-
```
r4mmer, line 138: This feature flag was from before ledger was enabled, just removed it since it does not seem useful anymore.

src/sagas/wallet.js
```diff
@@ -114,12 +114,11 @@ export function* startWallet(action) {
...
-  if (!walletId) {
-    // The wallet has not been initialized yet
-    if (hardware) {
-      yield LOCAL_STORE.initHWStorage(xpub);
-    } else {
+  if (hardware) {
+    yield LOCAL_STORE.initHWStorage(xpub);
+  } else {
+    const walletId = yield LOCAL_STORE.getWalletId();
+    if (!walletId) {
```
r4mmer, line 121: Here, we always init the hardware storage, this will ensure we generate the wallet id from the xpub from scratch to avoid issues when using 2 different ledger devices.

src/screens/SendTokens.js
```diff
@@ -190,14 +190,14 @@ class SendTokens extends React.Component {
...
-  onLedgerSuccess = (signatures) => {
+  onLedgerSuccess = async (signatures) => {
     try {
       // Prepare data and submit job to tx mining API
       const arr = [];
       for (let i=0;i<signatures.length;i++) {
         arr.push(Buffer.from(signatures[i]));
       }
-      this.sendTransaction.prepareTxFrom(arr);
+      await this.sendTransaction.prepareTxFrom(arr);
```
r4mmer, line 200: This was not being awaited and the transactions was being pushed before being ready, which would fail.

src/screens/SendTokens.js
```diff
@@ -219,7 +219,7 @@ class SendTokens extends React.Component {
   getSignatures = () => {
     ledger.getSignatures(
       Object.assign({}, this.data),
-      this.props.wallet.storage,
+      this.props.wallet,
```
r4mmer, line 222: getSignatures expect a wallet instance

src/components/Version.js
```diff
@@ -25,16 +25,14 @@ class Version extends React.Component {
...
-        $('#softwareWalletWarningModal').modal('show');
-        this.context.showModal(MODAL_TYPES.ALERT, {
-          body: <SoftwareWalletWarningMessage />,
-          buttonName: 'Ok',
-          id: 'softwareWalletWarningModal',
-          title: 'Software wallet warning',
-        });
-      }
+    if (!LOCAL_STORE.isHardwareWallet()) {
+      $('#softwareWalletWarningModal').modal('show');
```
andreabadesso, line 29: We should refactor this modal to be a GlobalModal, to prevent the "greyed screen" bug

Should we do it in this PR?

andreabadesso (changes requested):

src/components/Version.js
r4mmer: It already is, this line is doing nothing actually.
I removed it and the warning still works

andreabadesso (approved):

src/components/Version.js
```diff
@@ -25,16 +25,13 @@ class Version extends React.Component {
    * If it's software wallet show modal warning
    */
   walletTypeClicked = () => {
-    if (LEDGER_ENABLED) {
-      if (!LOCAL_STORE.isHardwareWallet()) {
-        $('#softwareWalletWarningModal').modal('show');
```
pedroferreira1, line 30: This is not needed anymore, right?

src/sagas/wallet.js
pedroferreira1: Can you add this comment to the code, please?

pedroferreira1 (approved):

src/sagas/wallet.js
r4mmer: Done

src/components/Version.js
r4mmer: There are 2 changes here.
- Removing the `$('#softwareWalletWarningModal').modal('show');` since the modal uses the global modal component so this is not needed.
- Removing the `LEDGER_ENABLED` branch since it is hardcoded to true.

This `LEDGER_ENABLED` flag was created as a feature flag before we had a proper feature flag solution, but since it serves no actual purpose now and the flag was creating some confusion I decided to remove it.

PR #421, opened 2023-10-24
chore: Update electron-builder

tuliomir:
### Acceptance Criteria
- The `electron-builder` dependency should be on the latest version

### Security Checklist


r4mmer (approved):

package.json
```diff
@@ -101,7 +100,6 @@
   "build": {
     "appId": "network.hathor.macos.wallet",
     "compression": "maximum",
-    "afterSign": "scripts/notarize.js",
```
msbrogli, line 104: I feel we have to add `"notarize": "notarytool"` here. But that's not exactly what I tested.

Link: https://www.electron.build/configuration/mac.html

msbrogli (changes requested):

package.json
tuliomir: Reading about the parameters to be passed to the build options, it seems that we can use all the default settings for notarization on our app.

I updated our pack settings to use them on [ef0cda4](https://github.com/HathorNetwork/hathor-wallet/pull/421/commits/ef0cda42a8662e0d9e98b0f3ada018f4b39f445e).

Reference: https://github.com/electron/notarize#method-notarizeopts-promisevoid

msbrogli: Note: You MUST specify APPLE_ID and APPLE_APP_SPECIFIC_PASSWORD via environment variables to activate notarization step

tuliomir: Added this description to the release docs on [b7f6b74](https://github.com/HathorNetwork/hathor-wallet/pull/421/commits/b7f6b749b640ba6c27993bbef0e21b20a63a6bc2).

msbrogli (approved):

package.json
```diff
@@ -110,6 +108,7 @@
     },
     "mac": {
       "provisioningProfile": "keys/mac_production.provisionprofile",
+      "notarize": { "teamId": "TEAM-ID-HERE" },
```
r4mmer, line 111: non-blocking: Can this be moved to a configuration file or an envvar?

This way the builder always has to edit the package.json file to build, so having it as a separate config file may be the best, unless we can put the team-id here (don't know if it has security issues)

r4mmer (approved):

package.json
tuliomir: I agree it's a nuisance to have this identifier set like this. However, it's hard to fix it in a simple way.

1) It's not possible to interpolate environment variables into the `package.json` file

2) This configuration objects feeds into the `update-builder` implementation of calling the notarization scripts.
If we wanted to call the notarization without this specific script we would need to implement the notarization ourselves. This is precisely what we removed on the deleted `scripts/notarize.js` file.

That way, I think this is a technical debt small enough to be solved later by a more elegant solution.

PR #425, opened 2023-10-25
Chore: Update to Node.js 18

tuliomir:
### Acceptance Criteria
- Should upgrade the application to use Node.js v18
- Should upgrade the lockfile to version 3
- No dependencies should be changed

### Security Checklist


package.json
```diff
@@ -59,9 +63,9 @@
   "scripts": {
     "build-css": "sass --no-source-map src/index.scss src/index.css",
     "watch-css": "npm run build-css && sass --no-source-map -w src/index.scss src/index.css",
-    "start-js": "react-scripts start",
+    "start-js": "react-scripts --openssl-legacy-provider start",
```
msbrogli, line 66: I think we should map which dependencies need this `--openssl-legacy-provider`. So we know when we can remove this flag.

msbrogli (changes requested):

tuliomir:
Superseded by #501

package.json
tuliomir: As described on the updated PR, the most relevant dependency for removing this flag is being discussed on #441

PR #430, opened 2023-11-03
fix: show untrust all tokens option

r4mmer:
### Acceptance Criteria
- To show the 'Untrust All Tokens' button we must have `LOCAL_STORE.isHardwareWallet()` as `true`, because it means that the ledger is in use.

### Security Checklist


tuliomir (approved):

andreabadesso (approved):
Approved, but this is not a `chore` at all, right?

From what I understood, this is a `fix`

PR #446, opened 2023-11-30
Fixes #443 - Automatically removes whitespaces from Addresses

### LINKED ISSUE #443: Remove whitespaces from addresses
When sending a transaction recently, I was getting a message the address was invalid. I later realized there was a whitespace before the address, which was causing the problem.

Maybe we should trim the string on the address field automatically?

ketansahugit:
### Acceptance Criteria
- Resolves the issue #443 
- Automatically removes whitespaces from the entered address to prevent errors during the transactions.

### Security Checklist

### DIFF
--- a/src/components/SendTokensOne.js
+++ b/src/components/SendTokensOne.js
@@ -104,7 +104,7 @@ class SendTokensOne extends React.Component {
   getData = () => {
     let data = {'outputs': [], 'inputs': []};
     for (const output of this.outputs) {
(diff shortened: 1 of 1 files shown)


tuliomir:
Hi @ketansahugit , thanks for the contribution!
Could you please change your base branch to `dev` on your fork? Our release process requires the changes to be made to `dev` and only later we consolidate it into a release to `master`.

PR #450, opened 2023-12-07
feat: Memory HybridStore

r4mmer:
### Acceptance Criteria

- Implement a hybrid store based off `MemoryStore`

### Security Checklist

### DIFF
--- a/public/electron.js
+++ b/public/electron.js
@@ -122,7 +122,7 @@ function createWindow () {
     template.push({
       label: 'Debug',
       submenu: [
-        { label: `Open DevTools`, click: function() { mainWindow.webContents.openDevTools(); }}
+        { label: `Open DevTools`, accelerator: 'CmdOrCtrl+B', click: function() { mainWindow.webContents.openDevTools(); }}
       ]
     });
   };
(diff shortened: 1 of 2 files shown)


public/electron.js
```diff
@@ -122,7 +122,7 @@ function createWindow () {
     template.push({
       label: 'Debug',
       submenu: [
-        { label: `Open DevTools`, click: function() { mainWindow.webContents.openDevTools(); }}
+        { label: `Open DevTools`, accelerator: 'CmdOrCtrl+B', click: function() { mainWindow.webContents.openDevTools(); }}
```
tuliomir, line 125: :star2: A debugging hotkey! Sweet!

src/storage.js
```diff
@@ -471,4 +570,6 @@ export class LocalStorageStore {
   }
 }
 
-export default new LocalStorageStore();
+const STORE = new LocalStorageStore();
+
+export default STORE;
```
tuliomir, line 575: question: Do you think this PR would be the most adequate scope to add unit tests for the Desktop Wallet storage? Or should we add those in the future?

pedroferreira1 (approved):

PR #460, opened 2023-12-15
chore: Upgrades `caniuse-lite`

tuliomir:
This solves a warning raised on server start indicating our `caniuse` database was obsolete.

### Acceptance Criteria
- Updates [`caniuse-lite`](https://github.com/browserslist/caniuse-lite) dependency

### Notes
Currently this breaks the CI because of some issue with the translation files. This is still pending a solution, as the `update_pot` command does not solve the problem.

### Security Checklist


tuliomir:
This dependency was updated through #567

PR #461, opened 2023-12-19
Refactor: Navigation root for Router v5

tuliomir:
In the process to update React Router to v6, it is first necessary to make the current implementation of v5 fully integrated with react hooks. ( see: [official docs](https://reactrouter.com/en/main/upgrading/v5#upgrade-to-react-router-v51) )

### Acceptance Criteria
- Refactor the Router `root` element for React to use React v18
- Refactor the `ErrorWrapper` to be a _functional component_ and use hooks
- Refactor the `ErrorWrapper` to be a direct child of the root router

### Security Checklist

### DIFF
--- a/src/ErrorWrapper.js
+++ b/src/ErrorWrapper.js
@@ -5,9 +5,8 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { Route } from 'react-router-dom';
-import wallet from './utils/wallet';
 import $ from 'jquery';
 import App from './App';
 import ModalUnhandledError from './components/ModalUnhandledError';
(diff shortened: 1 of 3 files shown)


src/ErrorWrapper.js
tuliomir, line 1: Most of the code logic on these components was unreachable and was removed on simplification.

src/index.js
```diff
@@ -21,14 +21,15 @@ import './index.css';
 import store from "./store/index";
 import { Provider } from "react-redux";
 
-ReactDOM.render(
+const container = document.getElementById('root');
+const root = createRoot(container);
+root.render(
```
tuliomir, line 26: This is the React 18 syntax of building the root element. ( [Docs](https://react.dev/blog/2022/03/08/react-18-upgrade-guide#updates-to-client-rendering-apis) )

src/screens/LoadWallet.js
```diff
@@ -95,6 +95,7 @@ class LoadWallet extends React.Component {
     // First we clean what can still be there of a last wallet
     wallet.generateWallet(this.state.words, '', pin, password, this.props.history);
     LOCAL_STORE.markBackupDone();
+    LOCAL_STORE.open(); // Mark this wallet as open, so that is does not appear locked after loading
```
tuliomir, line 98: This was actually a bug discovered during the tests for this refactor: any imported wallet would be locked.

The steps to reproduce this are:
- Reset the wallet
- Close the app and open it again
- Import a wallet
- Bug: the wallet would be locked again immediately after successful history loading

r4mmer (approved):

src/screens/LoadWallet.js
alexruzenhack, line 98:
```suggestion
    LOCAL_STORE.open(); // Mark this wallet as open, so that it does not appear locked after loading
```

alexruzenhack: Do we find this behavior in other wallets, or is it specific for this wallet?

alexruzenhack (approved):

src/screens/LoadWallet.js
tuliomir: It is specific to this wallet, because it has a special logic to display the "Welcome" screen.

PR #468, opened 2023-12-26
Refactor: Route `children` for welcome screens

tuliomir:
Among the [necessary steps](https://reactrouter.com/en/main/upgrading/v5#upgrade-to-react-router-v51) to upgrade to React Router v6 are:
- Using react hooks on every screen
- Using the `children` prop instead of `component` or `render` on the `<Route />` elements

There are many screens to update and the ones selected for the scope of this PR are only the screens that are accessed without any router wrappers on the `App.js` file ( Ex.: `NavigationRoute` or `StartedRoute` ).

Some other components that are easy to refactor were also included, such as the `BackButton` that no longer needs any `props` from its parent.

### Acceptance Criteria
- All routes that are not wrapped in other components should be refactored with these best practices
- All screens that are called by these routes should be refactored to _functional components_
- Components used by those screens that are easily refactored should be turned into _functional components_ as well
- Unreachable component files that used to interact with those screens should be removed

### Notes
As the screens' translatable messages were moved around, the `locale/texts.pot` had to be updated. This file alone has over 1000 lines of changed code that are actually from other contexts too.

### Security Checklist


src/components/RequestError.js
```diff
@@ -141,53 +125,66 @@ class RequestErrorModal extends React.Component {
...
     e.preventDefault();
-    $(this.refs.advancedData).hide(300);
-    $(this.refs.hideAdvancedLink).hide();
-    $(this.refs.showAdvancedLink).show();
+    $(advancedDataRef.current).hide(300);
+    $(hideAdvancedLinkRef.current).hide();
+    $(showAdvancedLinkRef.current).show();
   }
 
-  render() {
```
tuliomir, line 163: tip: To review these changes to the `render()` method I suggest using the "Hide whitespace" feature on GitHub, as it highlights only the words that were changed on this file, and not the indentation adjustments necessary:

Ignore Whitespace Diff

src/components/TxRow.js
```diff
@@ -1,39 +0,0 @@
-/**
```
r4mmer, line 1: _unrelated to line_

question(non-blocking): Were this and `WaitVersion.js` removed because they are not used anymore?

r4mmer (changes requested):
There are some problems to verify some screens since they only come up in special circunstances. For instance failing a request to the fullnode  (or timing out) with the wallet locked should show the `Server` screen so the user can select another server with the wallet still locked, this is very hard to check without manually spoofing a request.

Maybe we should check these unlikely scenarios on this PR since we are changing the component that handles errors.

src/components/TxRow.js
tuliomir: That's correct :+1:

src/App.js
```diff
@@ -153,11 +153,11 @@ class Root extends React.Component {
         <StartedRoute exact path="/signin" component={Signin} loaded={false} />
         <StartedRoute exact path="/hardware_wallet" component={StartHardwareWallet} loaded={false} />
         <NavigationRoute exact path="/locked" component={LockedWallet} />
```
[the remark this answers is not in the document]
tuliomir: Actually, it's a bit more complicated. That's why, on this PR, I decided to implement only on those routes without these wrappers.

When we refactor to use `children` instead of `Component`, we pass the rendered component to the props ( Ex.: `<Signin />` instead of `SignIn` ). This requires some changes on the wrappers that would make this PR much bigger that what it is.

I plan to refactor the `NavigationRoute` and `StartedRoute` each in their own PRs, according to [a suggestion from the official guide](https://reacttraining.com/blog/react-router-v5-1#staying-ahead-of-the-curve) to upgrading Routers:
> You don't have to migrate all of your route components at once! You can do just a few of them now and come back for the rest later.

It will be confusing for only a week or two until all PRs are merged, but they will all be done within a short time window.

What do you think?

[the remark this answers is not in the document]
tuliomir: I mentioned my plans for the `NavigationRoute` and `StartedRoute` on the comment above :+1:

tuliomir:
I had already tested turning off the internet connection and the `RequestError.js` modal had worked fine. However I tried to intercept the requests and test more throroughly as you suggested and found a bug on the retry button.

The solution for this was to remove the `retry` state altogether, as it no longer has a _setState callback_ function as it had before. The logic was redone so that it no longer depends on DOM manipulation and it has worked fine on all my tests so far.

This fix was implemented on [f2cd753](https://github.com/HathorNetwork/hathor-wallet/pull/468/commits/f2cd75399a27c06c774efde0622334053085423b).

src/App.js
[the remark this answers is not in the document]
r4mmer: LGTM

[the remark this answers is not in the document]
r4mmer: Ok

r4mmer (approved):

src/components/RequestError.js
```diff
@@ -5,81 +5,61 @@
...
+  const hideAdvancedLinkRef = useRef();
 
   /**
    * User clicked to change server, then push to choose server screen
    */
-  handleChangeServer = () => {
-    $('#requestErrorModal').modal('hide');
-    this.props.history.push('/server/');
+  const handleChangeServer = () => {
+    $(MODAL_DOM_ID).modal('hide');
```
andreabadesso, line 42: We should create an issue to convert this to the GlobalModal syntax

src/components/BackButton.js
```diff
@@ -7,33 +7,33 @@
 
 import React from 'react';
 import { t } from 'ttag';
+import { useHistory } from "react-router-dom";
```
andreabadesso, line 10:
```suggestion
import { useHistory } from 'react-router-dom';
```

src/components/RequestError.js
```diff
@@ -5,81 +5,61 @@
...
+import React, { useState, useEffect, useRef } from 'react';
 import { t } from 'ttag';
 import $ from 'jquery';
 import createRequestInstance from '../api/axiosInstance';
 import SpanFmt from './SpanFmt';
-import { connect } from 'react-redux';
+import { useSelector } from 'react-redux';
 import hathorLib from '@hathor/wallet-lib';
 import wallet from '../utils/wallet';
+import { useHistory } from "react-router-dom";
```
andreabadesso, line 16:
```suggestion
import { useHistory } from 'react-router-dom';
```

src/screens/SentryPermission.js
```diff
@@ -5,7 +5,8 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React from 'react';
+import React, { useState, useEffect } from 'react';
+import { useHistory } from "react-router-dom";
```
andreabadesso, line 9:
```suggestion
import { useHistory } from 'react-router-dom';
```

andreabadesso (approved):

src/screens/SentryPermission.js
tuliomir: Fixed on [e744e0d](https://github.com/HathorNetwork/hathor-wallet/pull/468/commits/e744e0d84bc13c6af17de4034a9c98cb0aafdda7)

src/components/RequestError.js
tuliomir: Fixed on [e744e0d](https://github.com/HathorNetwork/hathor-wallet/pull/468/commits/e744e0d84bc13c6af17de4034a9c98cb0aafdda7)

src/components/BackButton.js
tuliomir: Fixed on [e744e0d](https://github.com/HathorNetwork/hathor-wallet/pull/468/commits/e744e0d84bc13c6af17de4034a9c98cb0aafdda7)

PR #469, opened 2023-12-29
Refactor: Locked screen functional component

tuliomir:
### Acceptance Criteria
- The Locked screen should be a functional component
- The version error screen, only accessible through the Locked screen, should also be refactored
- The Navigation component can also be refactored, as it is displayed by the Locked screen on the `NavigationRoute` context

### Security Checklist

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -332,9 +332,7 @@ const returnDefaultComponent = (Component, props) => {
  * Return a component with the navigation component
  */
 const NavigationRoute = ({ component: Component, ...rest }) => (
-  <Route {...rest} render={(props) => (
-      returnDefaultComponent(Component, props)
-  )} />
+  returnDefaultComponent(Component, rest)
 )
 
 export default connect(mapStateToProps, mapDispatchToProps)(Root);
--- a/src/App.js
+++ b/src/App.js
@@ -317,7 +317,7 @@ const returnDefaultComponent = (Component, props) => {
     } else {
       return (
         <div className='component-div h-100'>
-          <Navigation {...props}/>
+          <Navigation />
           <Component {...props} />
           <RequestErrorModal />
         </div>
--- a/src/App.js
+++ b/src/App.js
@@ -324,7 +324,7 @@ const returnDefaultComponent = (Component, props) => {
       );
     }
   } else {
-    return <WalletVersionError {...props} />;
+    return <WalletVersionError />;
   }
 }
 
--- a/src/components/Navigation.js
+++ b/src/components/Navigation.js
@@ -12,74 +12,66 @@ import logo from '../assets/images/hathor-white-logo.png';
 import Version from './Version';
 import ServerStatus from './ServerStatus';
 import helpers from '../utils/helpers';
-import { connect } from "react-redux";
-
-
-const mapStateToProps = (state) => {
-  return {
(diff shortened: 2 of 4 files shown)


src/App.js
```diff
@@ -317,24 +317,22 @@ const returnDefaultComponent = (Component, props) => {
...
 }
 
 /*
  * Return a component with the navigation component
  */
 const NavigationRoute = ({ component: Component, ...rest }) => (
-  <Route {...rest} render={(props) => (
-      returnDefaultComponent(Component, props)
-  )} />
+  returnDefaultComponent(Component, rest)
```
tuliomir, line 335: The locked screen no longer needs props, so this is a viable simplification.

src/screens/WalletVersionError.js
```diff
@@ -30,89 +23,85 @@ const mapDispatchToProps = dispatch => {
...
+   * Called when reset wallet succeed, then close modal and go to welcome screen
    */
-  backupSuccess = () => {
-    this.context.hideModal();
-    LOCAL_STORE.markBackupDone();
-    this.props.updateWords(null);
-    this.alertSuccessRef.current.show(3000);
+  const handleReset = () => {
+    context.hideModal();
+    dispatch(walletReset());
```
tuliomir, line 61: Note: a bug was found while testing this. The redux `wallet` object is not available on the store during the context of this screen.

So, when the relevant saga method tries to reset the wallet, it finds nothing to reset and the operation is not executed properly.

Since the scope of this PR is only to refactor this screen to a functional component, this fix will be implemented in a future moment.

src/components/Navigation.js
```diff
@@ -12,74 +12,66 @@ import logo from '../assets/images/hathor-white-logo.png';
...
 import helpers from '../utils/helpers';
-import { connect } from "react-redux";
-
-
-const mapStateToProps = (state) => {
-  return {
-    useAtomicSwap: state.useAtomicSwap,
-  };
-};
+import { useSelector } from "react-redux";
```
andreabadesso, line 15:
```suggestion
import { useSelector } from 'react-redux';
```

src/screens/WalletVersionError.js
```diff
@@ -5,23 +5,16 @@
...
 import LOCAL_STORE from '../storage';
-
-
-const mapDispatchToProps = dispatch => {
-  return {
-    updateWords: (data) => dispatch(updateWords(data)),
-    walletReset: () => dispatch(walletReset()),
-  };
-};
+import { useHistory } from "react-router-dom";
```
andreabadesso, line 17:
```suggestion
import { useHistory } from 'react-router-dom';
```

src/screens/WalletVersionError.js
andreabadesso: Did you create an issue describing this?

andreabadesso (approved):

src/App.js
r4mmer: The `Route` wrapper/component adds no logic to work with the react router?

r4mmer (approved):

src/App.js
tuliomir: Not for a functional component, at least.
The `locked` screen is the only one called by this `NavigationRoute`. I even thought of removing it altogether, but decided to wait until a dedicated PR for the root element.

PR #422, opened 2023-10-24
chore: Pin dependencies

tuliomir:
### Acceptance Criteria
- All dependency versions should be pinned down to the patch number
- Dependency versions on package lock should not be changed

### Motivation
Besides the security improvement, the dependencies of the Desktop Wallet will be updated soon: having the version numbers pinned down will help with the peer reviews of the impacted dependencies.

### Security Checklist
- [X] Make sure you do not include new dependencies in the project unless strictly necessary and do not include dev-dependencies as production ones. More dependencies increase the possibility of one of them being hijacked and affecting us.

PR #474, opened 2024-01-04
fix: unresponsive ledger device

r4mmer:
### Summary

Fix the issues described in #473

### Acceptance Criteria
- Update ledger transport dependency
- Update hathor wallet lib dependency
- Do not redirect to locked screen when starting a hardware wallet
- During preload, clean any previous hardware wallet access data.
- Add better logs to debug issues with ledger communication

### Security Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,8 +25,8 @@
   "version": "0.27.1-rc2",
   "private": true,
   "dependencies": {
-    "@hathor/wallet-lib": "1.0.2",
-    "@ledgerhq/hw-transport-node-hid": "6.27.1",
+    "@hathor/wallet-lib": "1.0.4",
+    "@ledgerhq/hw-transport-node-hid": "6.28.1",
     "@sentry/electron": "3.0.7",
     "babel-polyfill": "6.26.0",
     "bootstrap": "4.6.1",
@@ -108,7 +108,9 @@
     },
     "mac": {
       "provisioningProfile": "keys/mac_production.provisionprofile",
-      "notarize": { "teamId": "TEAM-ID-HERE" },
+      "notarize": {
+        "teamId": "TEAM-ID-HERE"
+      },
       "hardenedRuntime": true,
       "gatekeeperAssess": false,
       "entitlements": "build/entitlements.mac.plist",
--- a/public/ledger.js
+++ b/public/ledger.js
@@ -108,10 +108,12 @@ class Ledger {
     this.subscriptor = Transport.listen({
       next: e => {
         if (e.type === 'add') {
+          console.log(`Device detected: ${e.deviceModel?.id}`);
           // We get this event when any app is opened on Ledger (Bitcoin, Ethereum, Hathor, etc). We have
           // the getVersion call next to make sure we're on the Hathor app.
           this.getVersion();
         } else if (e.type === 'remove') {
+          console.log(`Device removed: ${e.deviceModel?.id}`);
           // This is called in some situations
           // - Usb disconnected
           // - When user quits an app
@@ -134,7 +136,10 @@ class Ledger {
           }
         }
       },
-      error: error => {},
+      error: error => {
+        console.log('Error communicating with device');
+        console.log(error);
+      },
       complete: () => {}
     })
 
@@ -175,11 +180,21 @@ class Ledger {
 
     const { command, p1, p2, data, resolve, reject } = this.sendQueue.shift();
     this.getTransport()
-      .then(transport => transport.send(ledgerCLA, command, p1, p2, data))
-      .then((response) => { resolve(response) }, (error) => { reject(error) })
+      .then(transport => {
+        console.log(`Sending to ledger ${command} ${p1} ${p2}`);
+        return transport.send(ledgerCLA, command, p1, p2, data);
+      })
+      .then((response) => {
+        console.log(`Got response ${response.toString('hex')}`);
+        resolve(response);
+      }, (error) => {
+          console.log(error);
+          reject(error);
+        })
       .finally(() => {
-      // When done, check the queue again
-      this.checkSendQueue();
+        console.log('Finished sending command, will retry queue');
+        // When done, check the queue again
+        this.checkSendQueue();
     });
   }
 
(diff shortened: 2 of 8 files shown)


public/preload.js
```diff
@@ -19,6 +19,17 @@ process.once('loaded', () => {
   // Set closed in localStorage, so user does not open in the wallet page
   localStorage.setItem('localstorage:closed', 'true');
 
+  const accessDataRaw = localStorage.getItem('localstorage:accessdata');
+  if (accessDataRaw) {
+    // check if the access data is from a hardware wallet, if so we need to remove it
+    // This ensures that a previously closed hardware wallet will not be loaded again
+    // unless the device is connected
+    const accessData = JSON.parse(accessDataRaw);
+    if ((accessData?.walletFlags & 0x02) > 0) {
```
r4mmer, line 28: `0x02` is the hardware wallet flag ([ref](https://github.com/HathorNetwork/hathor-wallet-lib/blob/v1.0.4/src/types.ts#L204)).
We can expose `WALLET_FLAGS` from the lib on the next version and import it here but I wish to release this fix asap so I hardcoded the value here for now.

package.json
```diff
@@ -25,8 +25,8 @@
   "version": "0.27.1-rc2",
   "private": true,
   "dependencies": {
-    "@hathor/wallet-lib": "1.0.2",
-    "@ledgerhq/hw-transport-node-hid": "6.27.1",
+    "@hathor/wallet-lib": "1.0.4",
+    "@ledgerhq/hw-transport-node-hid": "^6.28",
```
tuliomir, line 29: question: Could we use a pinned version for this dependency? This has been the default since #422

src/storage.js
```diff
@@ -198,14 +198,17 @@ export class LocalStorageStore {
     }
   }
 
-  cleanWallet() {
-    this.removeItem(IS_HARDWARE_KEY);
-    this.removeItem(CLOSED_KEY);
+  cleanAccessData() {
```
tuliomir, line 201: suggestion: Could you add descriptions for these methods? It's not clear to me when we should use `clearWallet` or `cleanAccessData`.

r4mmer: Done

package.json
r4mmer: Done

public/ledger.js
```diff
@@ -134,7 +136,10 @@ class Ledger {
           }
         }
       },
-      error: error => {},
+      error: error => {
+        console.log('Error communicating with device');
+        console.log(error);
```
tuliomir, line 141: nitpick: Maybe switch these error messages to `console.error` instead?

tuliomir (approved):

src/screens/StartHardwareWallet.js
```diff
@@ -128,9 +129,6 @@ class StartHardwareWallet extends React.Component {
         xpub,
         hardware: true,
       });
-      LOCAL_STORE.markBackupDone();
-
-      this.props.history.push('/wallet/');
```
pedroferreira1, line 133: After starting the wallet we must change to the wallet screen. How is it done now? Through the sagas?

package.json
```diff
@@ -25,8 +25,8 @@
   "version": "0.27.1-rc2",
   "private": true,
   "dependencies": {
-    "@hathor/wallet-lib": "1.0.2",
-    "@ledgerhq/hw-transport-node-hid": "6.27.1",
+    "@hathor/wallet-lib": "1.0.4",
```
pedroferreira1, line 28: Why do we need this upgrade?

public/preload.js
pedroferreira1: Can you create the issue for that, please?

public/ledger.js
```diff
@@ -175,11 +180,21 @@ class Ledger {
...
-      .then((response) => { resolve(response) }, (error) => { reject(error) })
+      .then(transport => {
+        console.log(`Sending to ledger ${command} ${p1} ${p2}`);
+        return transport.send(ledgerCLA, command, p1, p2, data);
+      })
+      .then((response) => {
+        console.log(`Got response ${response.toString('hex')}`);
+        resolve(response);
+      }, (error) => {
+          console.log(error);
```
pedroferreira1, line 191: Is this indentation correct?

public/ledger.js
```diff
@@ -240,7 +255,9 @@ class Ledger {
         this.transport.decorateAppAPIMethods(this, this.methods, "HTR");
         resolve(transport);
       }, (e) => {
-        reject(Ledger.parseLedgerError(e));
+          console.log('Error creating transport');
```
pedroferreira1, line 258: Is this indentation correct?

src/storage.js
```diff
@@ -198,14 +198,26 @@ export class LocalStorageStore {
     }
   }
 
-  cleanWallet() {
-    this.removeItem(IS_HARDWARE_KEY);
-    this.removeItem(CLOSED_KEY);
+  /**
+   * Clean access data from persistent storage, making the wallet 'not loaded'
+   */
+  cleanAccessData() {
```
pedroferreira1, line 204: Why did you need to change this?

src/screens/StartHardwareWallet.js
r4mmer: This was here before to avoid changing the screen too fast (i.e. flashing the scren) before showing the loading addresses screen.
This has been fixed and is no longer needed, the saga will redirect to the loading addresses screen when required.

package.json
r4mmer: It's not necessary, Its just that wallet-lib v1.0.3 fixed an issue with the typing of the wallet-lib, which makes my LSP more useful (autocompletes, documentation, method signatures, etc.).

src/storage.js
r4mmer: When starting a hardware wallet we were having an issue that the acess data was being deleted during the `startWallet` saga, this would fail the process to start a wallet and if we clicked on retry it would work.

The issue was that we were being redirected to the locked screen as a hardware wallet, and we call `cleanWallet` to show the `wallet_type` screen.
This was fixed by not redirecting the hardware wallet to the locked screen during the `startWallet` saga and not cleaning the access data on `cleanWallet`, this is why I created the `cleanAccessData` if we really want to delete the access data from the localStorage.

public/preload.js
r4mmer: Done in https://github.com/HathorNetwork/hathor-wallet-lib/issues/601

public/ledger.js
r4mmer: Fixed

public/ledger.js
r4mmer: Fixed

pedroferreira1 (approved):

PR #475, opened 2024-01-05
Refactor: Admin screens to functional components

tuliomir:
### Acceptance Criteria
- Screens that are from the loaded wallet, but do not deal with transactions, should be refactored to functional components

### Security Checklist

### DIFF
--- a/locale/texts.pot
+++ b/locale/texts.pot
@@ -7,11 +7,11 @@ msgstr ""
 msgid "I want to reset my wallet"
 msgstr ""
 
-#: src/screens/AddressList.js:154
+#: src/screens/AddressList.js:160
 msgid "Search address"
 msgstr ""
 
-#: src/screens/AddressList.js:178
+#: src/screens/AddressList.js:184
 msgid "Addresses"
 msgstr ""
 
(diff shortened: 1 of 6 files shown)


locale/texts.pot
andreabadesso, line 1: You also need to update the other po files, right?

src/screens/AddressList.js
```diff
@@ -72,132 +44,165 @@ class AddressList extends React.Component {
...
+        fetchedAddresses.push(value);
+      }
+
+      setAddresses(fetchedAddresses);
+      setFilteredAddresses(fetchedAddresses);
+      setTotalPages(getTotalPages(fetchedAddresses));
+    }
+
+    fetchAddresses();
+  }, [wallet]);
```
andreabadesso, line 75: Can `wallet` mutate here? 

If it does, should we cancel the async task?

Maybe something like:

```suggestion
  useEffect(() => {
    let isCancelled = false;
    
    /* useEffect callback cannot be async, so we create a function to do the asynchronous work inside it
     * @see https://devtrium.com/posts/async-functions-useeffect#write-the-asynchronous-function-inside-the-useeffect
     */
    const fetchAddresses = async () => {
      const fetchedAddresses = [];
      const iterator = wallet.getAllAddresses();
      for (;;) {
        if (isCancelled) {
          break;
        }
        const addressObj = await iterator.next();
        const { value, done } = addressObj;

        if (done) {
          break;
        }

        fetchedAddresses.push(value);
      }
      if (!cancelled) {
        setAddresses(fetchedAddresses);
        setFilteredAddresses(fetchedAddresses);
        setTotalPages(getTotalPages(fetchedAddresses));
      }
    }

    fetchAddresses();
    
    return () => {
      isCancelled = true;
    };
  }, [wallet]);
```

andreabadesso (changes requested):

src/screens/AddressList.js
tuliomir: Good point!
`wallet` cannot mutate, as that would mean that the user has reset the wallet and reloaded it again without ever leaving the `AddressList` screen ( which is absurd ).

Adding the `wallet` property to the effect dependencies was a way to follow the guideline _[Don't lie to React about dependencies](https://overreacted.io/a-complete-guide-to-useeffect/#dont-lie-to-react-about-dependencies)_, but in this case it's not really needed.

So I believe the best course of action would be to remove the dependency altogether and just ignore this guideline as it doesn't fit this screen's business rules.

locale/texts.pot
tuliomir: At the moment there are no other `po` files to update, even though there is support for them on the `check_pot` script.
For now, no further action is needed here.

src/screens/AddressList.js
tuliomir: Fixed on [b4481d4](https://github.com/HathorNetwork/hathor-wallet/pull/475/commits/b4481d4b75fcc676ec66b76ad5be3d8aaa57c9b0)

alexruzenhack (approved):

andreabadesso (approved):

PR #497, opened 2024-01-26
fix: forceful quit

alexruzenhack:
### Acceptance Criteria
- refactor: extract id selector to ModalUnhandledError component
- refactor: replace Route by useHistory hook to transmit history to App
- fix: navigation to /wallet_type path after remove device

Closes: https://github.com/HathorNetwork/internal-issues/issues/224

### Security Checklist

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -72,7 +72,7 @@ class Root extends React.Component {
     // When Ledger device loses connection or the app is closed
     if (this.props.ledgerClosed && !prevProps.ledgerClosed) {
       LOCAL_STORE.lock();
-      this.props.history.push('/wallet_type/');
+      this.props.history.push('/locked/');
     }
   }
 
--- a/src/ErrorWrapper.js
+++ b/src/ErrorWrapper.js
@@ -9,7 +9,7 @@ import React, { useEffect, useState } from 'react';
 import { Route, withRouter } from 'react-router-dom';
 import $ from 'jquery';
 import App from './App';
-import ModalUnhandledError from './components/ModalUnhandledError';
+import ModalUnhandledError, { UNHANDLED_ERROR_MODAL_ID_SELECTOR } from './components/ModalUnhandledError';
 
 import 'bootstrap';
 import 'bootstrap/dist/css/bootstrap.min.css';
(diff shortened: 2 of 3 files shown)


src/App.js
```diff
@@ -72,7 +72,7 @@ class Root extends React.Component {
     // When Ledger device loses connection or the app is closed
     if (this.props.ledgerClosed && !prevProps.ledgerClosed) {
       LOCAL_STORE.lock();
-      this.props.history.push('/wallet_type/');
+      this.props.history.push('/locked/');
```
alexruzenhack, line 75: Without this change, a complex navigation rendering takes place, ending up to redirect to `/wallet` path and appearing it didn't changed the view.

On other hand, by changing the redirection to `/locked` path, the returnDefaultComponent function executes a wallet cleanup and and redirect to `/wallet_type` path, our desired destination.

I still have to test by I believe this change also fix this issue:
https://github.com/HathorNetwork/internal-issues/issues/225

src/ErrorWrapper.js
```diff
@@ -30,7 +33,7 @@ function ErrorBoundary({ onError }) {
 	}, []);
 
 	return (
-		<Route path="/" children={<App/>}/>
+		<App history={history} />
```
alexruzenhack, line 36: We doesn't need a router here because the `children` API from `Route` renders the component either it matches the path or not. Therefore, we can just render the component directly.

In general, the `children` API is useful for animation or style.

See [documentation](https://v5.reactrouter.com/web/api/Route/children-func).

src/ErrorWrapper.js
```diff
@@ -41,7 +44,7 @@ function ErrorWrapper(props) {
 		// Don't propagate error messages listened more than once
 		if (newError !== error) {
 			setError(error);
-			$('#unhandledErrorModal').modal('show');
+			$(UNHANDLED_ERROR_MODAL_ID_SELECTOR).modal('show');
```
alexruzenhack, line 47: Extract the hardcoded value to a constant in the origin.

tuliomir (approved):
I think it's weird to have the `ErrorWrapper` component render the whole `App` component. It seems to me to be some part of our logic that could be improved, but this is out of the scope of this PR as I understand it.

r4mmer (approved):

PR #502, opened 2024-01-31
refactor: Functional root component

tuliomir:
### Acceptance Criteria
- Refactors the `Root` component inside `App.js` to a functional component
- Refactors the `Version` component, called uniquely by `Root` to a functional component
- Refactors the `VersionError` component, called uniquely by `Root` to a functional component
- Adapts the event handling codes to work with `useEffect()`

### Security Checklist


src/App.js
```diff
@@ -87,252 +96,266 @@ class Root extends React.Component {
...
+    // Fetch the version information before rendering any screens, if possible
+    if (isVersionAllowed === undefined && wallet) {
+      helpersUtils.loadStorageState();
+
+      // We already handle all js errors in general and open an error modal to the user
+      // so there is no need to catch the promise error below
+      versionUtils.checkApiVersion(wallet);
     }
-  }
+  }, [isVersionAllowed, wallet]);
```
tuliomir, line 149: The version validation used to be inside the `returnDefaultComponent` function, but was moved up to the `Root` level to reduce the amount of times it was called during the application runtime.

src/App.js
```diff
@@ -87,252 +96,266 @@ class Root extends React.Component {
...
+    }
+    setVersionIsKnown(true);
+  }, [isVersionAllowed]);
+
+  // Do not any wallet screens if the wallet version is still unknown. Only the Locked Sreen is allowed
+  const isLockedScreen = history.location.pathname === '/locked/';
+  if (!versionIsKnown) {
+    if (!isLockedScreen) {
+      return null;
     }
```
alexruzenhack, line 335: I didn't understand it. Shouldn't it return at least an error page?

tuliomir: This indeed is still confusing. I believe the outer navigation and outer wrappers should be refactored in a future moment for increased legibility.

Added some more documentation on [ae60eeb](https://github.com/HathorNetwork/hathor-wallet/pull/502/commits/ae60eebbab367a2be5c3dba0c28b41c5149bcfd6), but in short: this screen should not render anything while the version is still unknown.

The response from the fullnode informing the version should be received in a fraction of a second, so this will normally not be an issue. Connection failures with the fullnode during this request should be intercepted by the `<ErrorWrapper />` component

alexruzenhack (approved):

r4mmer (changes requested):

src/App.js
```diff
@@ -87,252 +96,270 @@ class Root extends React.Component {
...
+        IPC_RENDERER.removeAllListeners('ledger:closed');
+        IPC_RENDERER.removeAllListeners('ledger:manyTokenSignatureValid');
+      }
+    };
+  }, []);
+
+  useEffect(() => {
+    // Fetch the version information before rendering any screens, if possible
+    if (isVersionAllowed === undefined && wallet) {
+      helpersUtils.loadStorageState();
```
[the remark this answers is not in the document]
tuliomir: This is already as close to the initial startup as I could move it while being in context, because it needs an initialized wallet to check the API version.

If I moved this into the `ErrorWrapper`, `ErrorBoundary` it would seem to be out of place, I think.
Do you agree?

src/App.js
```diff
@@ -87,252 +96,270 @@ class Root extends React.Component {
...
-          // arg.data is a list of uids with invalid signatures
-          arg.data.forEach(uid => {
-            tokens.removeTokenSignature(uid.toString('hex'));
-          });
-        }
-      });
+    // If there is no ledger connected, finish execution here
+    if (!IPC_RENDERER) {
+      return;
     }
```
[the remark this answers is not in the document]
tuliomir: That's a good point. Refactored on [74bab60](https://github.com/HathorNetwork/hathor-wallet/pull/502/commits/74bab601b5510949f1ba894ca71e238b938ed256).

src/App.js
```diff
@@ -87,252 +96,270 @@ class Root extends React.Component {
...
-          await storage.isHardwareWallet()
-        ) {
-          // remove all invalid signatures
-          // arg.data is a list of uids with invalid signatures
-          arg.data.forEach(uid => {
-            tokens.removeTokenSignature(uid.toString('hex'));
-          });
-        }
-      });
+    // If there is no ledger connected, finish execution here
```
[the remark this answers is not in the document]
tuliomir: Improved the description on [74bab60](https://github.com/HathorNetwork/hathor-wallet/pull/502/commits/74bab601b5510949f1ba894ca71e238b938ed256).

src/App.js
```diff
@@ -87,252 +96,270 @@ class Root extends React.Component {
...
+  }, []);
+
+  useEffect(() => {
+    // Fetch the version information before rendering any screens, if possible
+    if (isVersionAllowed === undefined && wallet) {
+      helpersUtils.loadStorageState();
+
+      // We already handle all js errors in general and open an error modal to the user
+      // so there is no need to catch the promise error below
+      versionUtils.checkApiVersion(wallet);
```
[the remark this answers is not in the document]
tuliomir: It makes sense. On [51dc965](https://github.com/HathorNetwork/hathor-wallet/pull/502/commits/51dc9659c0e91479e84d2e08313f85680db9da2c) I modified the `Server.js` logic to reset the `isVersionAllowed` property, triggering a new validation from this effect.

src/screens/Server.js
```diff
@@ -203,6 +205,9 @@ function Server() {
       setLoading(false);
       setErrorMessage(e.message);
     }
+
+    // Forces the validation of the allowed version
+    dispatch(isVersionAllowedUpdate({ allowed: undefined }));
```
r4mmer, line 210: I like this solution but can you double check that the server change flow is not affected?
For instance, we have:
- Change server from mainnet to another in mainnet
- Change server from mainnet to testnet successfully
- Change server from mainnet to testnet and fail the testnet confirm modal
- Change server from testnet to another in testnet
- Change server from testnet to mainnet

Since we are changing a variable that may change the current screen I think this is important.

src/App.js
```diff
@@ -213,12 +212,8 @@ function LoadedWalletComponent({ children }) {
...
   // Check version
   if (isVersionAllowed === undefined) {
-    throw new Error('[LoadedWalletComponent] isVersionAllowed is undefined');
-    // return <Redirect to={{
-    //   pathname: '/loading_addresses/',
-    //   state: {path: match.url},
-    //   waitVersionCheck: true
-    // }} />;
+    // The version information is not yet available, no op
+    return;
```
r4mmer, line 216: Why this change?

src/App.js
```diff
@@ -109,230 +121,240 @@ class Root extends React.Component {
...
+  useEffect(() => {
+    // Fetch the version information before rendering any screens, if possible
+    if (isVersionAllowed === undefined && wallet) {
+      helpersUtils.loadStorageState();
+
+      // We already handle all js errors in general and open an error modal to the user
+      // so there is no need to catch the promise error below
+      versionUtils.checkApiVersion(wallet);
     }
+  }, [isVersionAllowed, wallet]);
```
r4mmer, line 148: `checkApiVersion` is an async method with a state change, I think we should write this `useEffect` more like whats described [here](https://devtrium.com/posts/async-functions-useeffect) what do you think?

r4mmer (approved):

src/App.js
tuliomir: Because when the `ServerChange` sets the `isVersionAllowed` to `undefined` to force a revalidation of the version, the `LoadedWalletComponent` would throw an exception and the whole application would come to a halt.

Instead, it should just wait a few more fractions of a second for this property to be populated and the app can continue working normally.

src/screens/Server.js
tuliomir: All those were working OK, but the dispatch was not being done on an optimal place.
Refactored it to inside the `executeServerChange()` method.

src/App.js
tuliomir: As we discussed, this is an important change and will be made on a dedicaded PR. I'll open an issue for that.

tuliomir: This is being tracked on issue
- #526

r4mmer (approved):

PR #510, opened 2024-02-02
fix: Accessing colors from `scss`

tuliomir:
After the `react-scripts` upgrade on #426 the colors on the loading component were incorrect. This was tracked to [a known issue](https://github.com/facebook/create-react-app/issues/10047) on CRA v4 that is currently not fixed on the stable release and probably will be available only on v5 onwards.

The chosen workaround for this bug was to rename the `index.scss` file to `index.module.scss` so that the current configurations of `react-scripts` can identify it and load its data correctly to the application.

To minimize future impacts related to `scss` interaction, this interface was also centralized and sanitized on the `constants` file.

### Acceptance Criteria
- Exported colors from the `scss` file should be retrieved from javascript context

### Security Checklist


src/constants.js
```diff
@@ -267,3 +268,12 @@ export const FEATURE_TOGGLE_DEFAULTS = {
   [WALLET_SERVICE_FEATURE_TOGGLE]: false,
   [ATOMIC_SWAP_SERVICE_FEATURE_TOGGLE]: false,
 };
+
+/**
+ * This property filters the full exported scss properties and exposes only the ones
```
pedroferreira1, line 273: Is there any downside to exposing all exported variables from css? I'm just wondering about this because in case we decide to add a new variable there, we would need to change it here as well. If we do a direct export, we must change one place only.

tuliomir: In the current webpack loading behavior all elements from the `scss` file, both class definitions and colors in a garbled notation, are exposed to the javascript environment together with the ones we explicitly export through `:export {}`.

So, the downside of leaving this unfiltered is making the `constants.colors` property have hundreds of elements inside it that are not colors, which can confuse developers that try to read all constants values. 

Other than that, removing the filter will have no other downsides on the application.

r4mmer (approved):

pedroferreira1 (approved):

PR #520, opened 2024-02-05
feat: token info update

r4mmer:
### Acceptance Criteria
- On wallet screen update token info on any history change
- Mark wallet as opened when creating a new software wallet

### Security Checklist

### DIFF
--- a/src/screens/NewWallet.js
+++ b/src/screens/NewWallet.js
@@ -104,13 +104,14 @@ function NewWallet() {
    */
   const pinSuccess = () => {
     // Generate addresses and load data
+    LOCAL_STORE.unlock();
     wallet.generateWallet(words, '', pin, password, history);
+    // Mark this wallet as open, so that it does not appear locked after loading
+    LOCAL_STORE.open();
     // Clean pin, password and words from redux
     dispatch(updatePassword(null));
     dispatch(updatePin(null));
     dispatch(updateWords(null));
-    // Go to wallet
-    history.push('/wallet/');
   }
 
   /**
--- a/src/screens/Wallet.js
+++ b/src/screens/Wallet.js
@@ -78,9 +78,6 @@ function Wallet() {
     };
   });
 
-  // Token history timestamp to check if we should fetch the token details again
-  const [tokenHistoryTimestamp, setTokenHistoryTimestamp] = useState(-1);
-
   // Refs
   const alertSuccessRef = useRef(null);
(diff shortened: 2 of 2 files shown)


src/screens/NewWallet.js
```diff
@@ -104,13 +104,14 @@ function NewWallet() {
...
+    LOCAL_STORE.unlock();
     wallet.generateWallet(words, '', pin, password, history);
+    // Mark this wallet as open, so that it does not appear locked after loading
+    LOCAL_STORE.open();
     // Clean pin, password and words from redux
     dispatch(updatePassword(null));
     dispatch(updatePin(null));
     dispatch(updateWords(null));
-    // Go to wallet
-    history.push('/wallet/');
```
pedroferreira1, line 113: How are we changing screens without this now?

src/screens/Wallet.js
```diff
@@ -99,19 +96,10 @@ function Wallet() {
...
-    });
-
-    if (tokenHistoryTimestamp !== selectedTokenHistory.updatedAt) {
-      setTokenHistoryTimestamp(selectedTokenHistory.updatedAt);
-      updateTokenInfo(selectedToken);
-      updateWalletInfo(selectedToken);
-    }
+    updateTokenInfo(selectedToken);
+    updateWalletInfo(selectedToken);
   }, [tokensHistory]);
```
pedroferreira1, line 103: Is it possible to have here the `tokensHistory` of the `selectedToken`? Otherwise we would be updating token info and wallet info every time any token has its history updated.

src/screens/NewWallet.js
r4mmer: The redux has a copy of `reactHistory` object and the wallet start saga will redirect the screen to the `loading_addresses` screen when the wallet is loading.
After that it will naturally go to the loaded wallet screen since the wallet is loaded, unlocked and opened.

src/screens/Wallet.js
r4mmer: Yes, this is the downside of this approach.
I can test with `[selectedToken, tokensHistory[selectedToken]]` but react is not great at detecting changes on arrays/objects, we would need to check something specific like `[selectedToken, tokensHistory[selectedToken].updatedAt]` but this would re-enable the issue this is meant to fix.

alexruzenhack (approved):

pedroferreira1 (approved):

PR #528, opened 2024-02-23
feat: add feature flag for nano in Navigation

pedroferreira1:
### Acceptance Criteria
- Add feature flag for nano navigation with `false` as default

### Security Checklist

### DIFF
--- a/src/components/Navigation.js
+++ b/src/components/Navigation.js
@@ -13,6 +13,8 @@ import Version from './Version';
 import ServerStatus from './ServerStatus';
 import helpers from '../utils/helpers';
 import { useSelector } from 'react-redux';
+import { FEATURE_TOGGLE_DEFAULTS, NANO_CONTRACTS_FEATURE_TOGGLE } from '../constants';
+import { get } from 'lodash';
 
 /**
  * Component that shows a navigation bar with the menu options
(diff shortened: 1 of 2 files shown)


src/components/Navigation.js
```diff
@@ -21,6 +23,9 @@ import { useSelector } from 'react-redux';
  */
 function Navigation() {
   const useAtomicSwap = useSelector(state => state.useAtomicSwap);
+  const featureToggles = useSelector(state => state.featureToggles);
```
tuliomir, line 26: thought(non-blocking): Should we integrate the atomic swap flag into the feature toggles in the future? This is a bit of information that also comes from unleash, so it seems fitting to have them all together.

src/components/Navigation.js
```diff
@@ -32,6 +37,14 @@ function Navigation() {
     helpers.openExternalURL(helpers.getExplorerURL());
   }
 
+  const renderNCNavigation = () => {
+    return (
+      <li className="nav-item">
+        <NavLink to="/nano_contract/" exact className="nav-link" activeClassName="active" activeStyle={{ fontWeight: 'bold' }}>{t`Nano Contract`}</NavLink>
```
tuliomir, line 43: chore: Some of those props are no longer allowed since the upgrade to React Router v6. You can see which ones were removed on [the diff of PR 555](https://github.com/HathorNetwork/hathor-wallet/pull/555/files#diff-4e130ab97c387c481bb920b51e80a88b3148332602174ea833378ad96428d810)

If you just keep the pattern of the other ones in this file, with only `to` and `className`, it should work ok.

Also: the atomic Swap navigation button has a different approach to the conditional rendering, injecting the element directly without using a function. It would be good to keep a consistent pattern of those conditionals.

tuliomir (changes requested):

andreabadesso (approved):

src/components/Navigation.js
pedroferreira1: Issue created https://github.com/HathorNetwork/hathor-wallet/issues/604

src/components/Navigation.js
pedroferreira1: Removed the unecessary attributes and used the same conditional format as in the atomic swap:
https://github.com/HathorNetwork/hathor-wallet/pull/528/commits/46e45f43ca4a980384e084203e0265d1cee2de76

tuliomir (approved):

PR #529, opened 2024-02-23
feat: add nano contract list and modal to register a nano

pedroferreira1:
### Acceptance Criteria
- Add nano contract list screen.
- Add modal to register a nano contract.
- Add support for nano testnet network.
- HTTP 4XX errors shouldn't show error modal.

TODO: Missing wallet lib bump to v1.6.0 with `getRegisteredNanoContracts` method

https://github.com/HathorNetwork/hathor-wallet/assets/3298774/29af27e3-7b44-444a-86fb-549ae77bb805

### Security Checklist

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -60,6 +60,7 @@ export const types = {
   NANOCONTRACT_REGISTER_ERROR: 'NANOCONTRACT_REGISTER_ERROR',
   NANOCONTRACT_REGISTER_SUCCESS: 'NANOCONTRACT_REGISTER_SUCCESS',
   NANOCONTRACT_CLEAN_REGISTER_METADATA: 'NANOCONTRACT_CLEAN_REGISTER_METADATA',
+  BLUEPRINT_ADD_INFORMATION: 'BLUEPRINT_ADD_INFORMATION',
 };
 
 /**
@@ -577,4 +578,14 @@ export const setNativeTokenData = (data) => ({
 export const addRegisteredTokens = (data) => ({
   type: types.ADD_REGISTERED_TOKENS,
   payload: data,
+});
+
+/**
+ * Add blueprint information
+ *
+ * @param {NanoContractBlueprintInformationAPIResponse} blueprintInformation Blueprint information to add to redux
+ */
+export const addBlueprintInformation = (blueprintInformation) => ({
+  type: types.BLUEPRINT_ADD_INFORMATION,
+  payload: { blueprintInformation }
 });
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -60,6 +60,8 @@ export const types = {
   NANOCONTRACT_REGISTER_ERROR: 'NANOCONTRACT_REGISTER_ERROR',
   NANOCONTRACT_REGISTER_SUCCESS: 'NANOCONTRACT_REGISTER_SUCCESS',
   NANOCONTRACT_CLEAN_REGISTER_METADATA: 'NANOCONTRACT_CLEAN_REGISTER_METADATA',
+  NANOCONTRACT_EDIT_ADDRESS: 'NANOCONTRACT_EDIT_ADDRESS',
+  NANOCONTRACT_UNREGISTER: 'NANOCONTRACT_UNREGISTER',
   BLUEPRINT_ADD_INFORMATION: 'BLUEPRINT_ADD_INFORMATION',
 };
 
@@ -589,4 +591,25 @@ export const addRegisteredTokens = (data) => ({
 export const addBlueprintInformation = (blueprintInformation) => ({
   type: types.BLUEPRINT_ADD_INFORMATION,
   payload: { blueprintInformation }
+});
+
+/**
+ * Edit address of a registered nano contract
+ *
+ * @param {string} ncId Nano contract id
+ * @param {string} address New address of the nano contract
+ */
+export const editAddressNC = (ncId, address) => ({
+  type: types.NANOCONTRACT_EDIT_ADDRESS,
+  payload: { ncId, address }
+});
+
+/**
+ * Unregister nano contract
+ *
+ * @param {string} ncId ID of nano contract to unregister
+ */
+export const nanoContractUnregister = (ncId) => ({
+  type: types.NANOCONTRACT_UNREGISTER,
+  payload: ncId,
 });
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -56,6 +56,10 @@ export const types = {
   SET_MINING_SERVER: 'SET_MINING_SERVER',
   SET_NATIVE_TOKEN_DATA: 'SET_NATIVE_TOKEN_DATA',
   ADD_REGISTERED_TOKENS: 'ADD_REGISTERED_TOKENS',
+  NANOCONTRACT_REGISTER_REQUEST: 'NANOCONTRACT_REGISTER_REQUEST',
+  NANOCONTRACT_REGISTER_ERROR: 'NANOCONTRACT_REGISTER_ERROR',
+  NANOCONTRACT_REGISTER_SUCCESS: 'NANOCONTRACT_REGISTER_SUCCESS',
+  NANOCONTRACT_CLEAN_REGISTER_METADATA: 'NANOCONTRACT_CLEAN_REGISTER_METADATA',
 };
 
 /**
@@ -132,9 +136,11 @@ export const updateHeight = (height, htrUpdatedBalance) => ({ type: 'update_heig
 
 /**
  * tokens {Array} array of token uids the the wallet has
+ * registeredTokens {{uid:string, name:string, symbol:string}[]} The registered tokens from the storege
  * currentAddress {Object} The current unused address
+ * registeredNanoContracts {NanoContractData} The registered nano contracts from the storege
  */
-export const loadWalletSuccess = (tokens, registeredTokens, currentAddress) => ({ type: 'load_wallet_success', payload: { tokens, registeredTokens, currentAddress } });
+export const loadWalletSuccess = (tokens, registeredTokens, currentAddress, registeredNanoContracts) => ({ type: 'load_wallet_success', payload: { tokens, registeredTokens, currentAddress, registeredNanoContracts } });
 
 /**
  * tx {Object} the new transaction
@@ -508,6 +514,43 @@ export const updateMiningServer = (url, reset) => ({
 export const setMiningServer = (url) => ({
   type: types.SET_MINING_SERVER,
   payload: url,
+
+/**
+ * Register nano contract in redux
+ *
+ * @param {string} ncId Nano contract id
+ * @param {string} address Address associated with the nano contract
+ */
+export const registerNanoContract = (ncId, address) => ({
+  type: types.NANOCONTRACT_REGISTER_REQUEST,
+  payload: { ncId, address },
+});
+
+/**
+ * Error when registering a nano contract
+ * @param {string} error Error reason
+ */
+export const nanoContractRegisterError = (error) => ({
+  type: types.NANOCONTRACT_REGISTER_ERROR,
+  payload: { error },
+});
+
+/**
+ * Success when registering nano contract
+ *
+ * @param {INcData} ncData nano contract data
+ */
+export const nanoContractRegisterSuccess = (ncData) => ({
+  type: types.NANOCONTRACT_REGISTER_SUCCESS,
+  payload: ncData,
+});
+
+/**
+ * Clean register metadata of nano contracts
+ *
+ */
+export const cleanNanoContractRegisterMetadata = () => ({
+  type: types.NANOCONTRACT_CLEAN_REGISTER_METADATA,
 });
 
 /**
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -137,9 +137,9 @@ export const updateHeight = (height, htrUpdatedBalance) => ({ type: 'update_heig
 
 /**
  * tokens {Array} array of token uids the the wallet has
- * registeredTokens {{uid:string, name:string, symbol:string}[]} The registered tokens from the storege
+ * registeredTokens {{uid:string, name:string, symbol:string}[]} The registered tokens from the storage
  * currentAddress {Object} The current unused address
- * registeredNanoContracts {NanoContractData} The registered nano contracts from the storege
+ * registeredNanoContracts {NanoContractData} The registered nano contracts from the storage
  */
 export const loadWalletSuccess = (tokens, registeredTokens, currentAddress, registeredNanoContracts) => ({ type: 'load_wallet_success', payload: { tokens, registeredTokens, currentAddress, registeredNanoContracts } });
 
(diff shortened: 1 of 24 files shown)


src/utils/nanoContracts.js
```diff
@@ -0,0 +1,49 @@
...
+
+  /**
+   * Check if transaction is a nano contract create tx
+   *
+   * @param {Transaction} tx
+   * @returns {boolean}
+   */
+  isNanoContractCreate(tx) {
+    return tx.version === hathorLib.constants.NANO_CONTRACTS_VERSION && tx.nc_method === hathorLib.constants.NANO_CONTRACTS_INITIALIZE_METHOD;
+  },
```
r4mmer, line 46: This makes more sense as an util from the lib, what do you think?

src/utils/version.js
```diff
@@ -48,7 +48,7 @@ const version = {
     // Set network in lib to use the correct address byte
     let network = data.network;
 
-    if (data.network.startsWith('testnet')) {
+    if (data.network.includes('testnet')) {
```
r4mmer, line 51: This is also happens in other files, like: 

- `src/components/ServerStatus.js#33` 
- `src/screens/Server.js#L240`

src/screens/nano/NanoContractList.js
```diff
@@ -0,0 +1,78 @@
...
+  const context = useContext(GlobalModalContext);
+  const nanoContracts = useSelector(state => state.nanoContracts);
+  const navigate = useNavigate();
+
+  const registerNC = () => {
+    context.showModal(MODAL_TYPES.NANOCONTRACT_REGISTER);
+  }
+
+  const createNC = () => {
+    navigate('/nano_contract/select_blueprint/');
```
r4mmer, line 35: This screen is probably part of another PR, correct?
In any case, what do you think of using `/nano_contract/create` instead?

src/sagas/nanoContract.js
```diff
@@ -0,0 +1,115 @@
...
+  if (!nanoUtils.isNanoContractCreate(tx)) {
+    console.debug('Transaction is not a nano contract creation transaction.');
+    yield put(nanoContractRegisterError(t`Invalid nano contract creation transaction.`));
+    return;
+  }
+
+  let blueprintInformation = blueprintsData[tx.nc_blueprint_id];
+  if (!(blueprintInformation)) {
+    // We store the blueprint information in redux because it's used in some screens
+    // then we don't need to request it every time
```
tuliomir, line 81: suggestion(non-blocking): this comment could also be copied on the `reducers` file, as an explanation on why to store that information there.

src/sagas/nanoContract.js
```diff
@@ -0,0 +1,115 @@
...
+  const nc = {
+    address,
+    ncId,
+    blueprintId: tx.nc_blueprint_id,
+    blueprintName: blueprintInformation.name
+  };
+  yield call([wallet.storage, wallet.storage.registerNanoContract], ncId, nc);
+
+  console.debug('Success registering Nano Contract.');
+  // emit action NANOCONTRACT_REGISTER_SUCCESS
```
tuliomir, line 106: suggestion: remove this comment

src/storage.js
```diff
@@ -134,9 +136,62 @@ class HybridStore extends MemoryStore {
...
+  /**
+   * Register a nano contract.
+   *
+   * @param ncId {string} Nano contract id to register
+   * @param ncValue {INcData} Nano contract data to register
+   * @async
+   * @returns {Promise<void>}
+   */
+  async registerNanoContract(ncId, ncValue) {
+    await super.registerNanoContract(ncValue);
```
tuliomir, line 165: question(blocking): Are we missing the `ncId` parameter here?

```suggestion
    await super.registerNanoContract(ncId, ncValue);
```

src/utils/version.js
tuliomir: thought(non-blocking): I have the impression that this is another util that would be helpful to have centralized on the lib itself. We recently had some issues with this specific code on the mobile wallet as well.

src/reducers/index.js
```diff
@@ -193,6 +312,8 @@ const rootReducer = (state = initialState, action) => {
       return resetSelectedTokenIfNeeded(state, action);
     case 'set_ledger_was_closed':
       return Object.assign({}, state, { ledgerWasClosed: action.payload });
+    case 'save_nano_contract':
```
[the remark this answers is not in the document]
pedroferreira1: This was left from a previous rebase, I've just removed it.

src/components/nano/ModalRegisterNanoContract.js
```diff
@@ -0,0 +1,124 @@
+/**
```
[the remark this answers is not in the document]
pedroferreira1: I agree it's better. I think I prefer doing this change in the last PR of the nano sequence to prevent some conflicts in rebases. Is it ok for you?

src/utils/nanoContracts.js
pedroferreira1: The method was added to the lib: https://github.com/HathorNetwork/hathor-wallet-lib/pull/691

As soon as the version is released I will create a PR to use it here. (if this PR takes more time to be merged, I will use it here already)

src/utils/version.js
pedroferreira1: I created a new PR in the lib for this (https://github.com/HathorNetwork/hathor-wallet-lib/pull/696) because this was also a bug in the mobile.

As soon this PR is merged and the lib version is released, I will create a PR to use this method in both wallets.

src/screens/nano/NanoContractList.js
pedroferreira1: Yes, this is the next PR. I decided to use `select_blueprint` because there's a first screen for the user to select the blueprint before the screen to create the nano contract.

We might use this `select_blueprint` screen for other reasons, so I prefer to keep it more generic like this.

src/api/axiosInstance.js
```diff
@@ -47,6 +47,11 @@ const createRequestInstance = (resolve, timeout) => {
     // Adding conditional because if the server forgets to send back the CORS
     // headers, error.response will be undefined
     const statusCode = error.response ? error.response.status : -1;
+    if (statusCode < 500) {
+      // This request error is to handle problems in the server, so the user
+      // can retry a request. So 404 and 400 errors shouldn't show the error modal
+      return Promise.reject(error);
+    }
```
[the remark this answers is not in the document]
pedroferreira1: The error handling we have nowadays is not good for some reasons and the main one is that it depends on a promise that the resolve method is stored in the redux, so we can't have API calls with async/await. 

We must have a full refactor in the retry mechanism, so it can handle errors better maybe with the `validateStatus` as optional parameter as @tuliomir suggested.

I will change the code to specify 400 and 404 errors only to be excluded from this error handling, as @r4mmer noticed that the current code has a bug for 429 errors, and then we can proper discuss this new retry mechanism in an [issue](https://github.com/HathorNetwork/hathor-wallet/issues/641).

What do you think?

Note: we have explorer service calls (metadataApi) that don't show the error modal for 404 responses but they use the explorerServiceAxios, not the default axios, which is overridden by the desktop wallet.

src/components/nano/ModalRegisterNanoContract.js
```diff
@@ -0,0 +1,124 @@
...
+                <div className="col-12 col-sm-10">
+                    <p className="error-message text-danger">
+                      {ncRegisterStatus === NANOCONTRACT_REGISTER_STATUS.ERROR && ncRegisterMetadata.error}
+                    </p>
+                </div>
+              </div>
+            </form>
+          </div>
+          <div className="modal-footer">
+            {ncRegisterStatus === NANOCONTRACT_REGISTER_STATUS.LOADING && <ReactLoading type='spin' color={colors.purpleHathor} width={24} height={24} delay={200}/>}
```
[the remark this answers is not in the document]
pedroferreira1: I didn't understand. If I use `ncRegisterStatus === NANOCONTRACT_REGISTER_STATUS.LOADING or ncRegisterStatus == null` here I would be showing the loading icon before typing any ID.

src/sagas/nanoContract.js
```diff
@@ -0,0 +1,115 @@
...
+    yield put(addBlueprintInformation(blueprintInformation));
+  }
+
+  const nc = {
+    address,
+    ncId,
+    blueprintId: tx.nc_blueprint_id,
+    blueprintName: blueprintInformation.name
+  };
+  yield call([wallet.storage, wallet.storage.registerNanoContract], ncId, nc);
```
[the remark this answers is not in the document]
pedroferreira1: Yes, it makes sense! Done it

src/actions/index.js
```diff
@@ -128,9 +133,11 @@ export const updateHeight = (height, htrUpdatedBalance) => ({ type: 'update_heig
 
 /**
  * tokens {Array} array of token uids the the wallet has
+ * registeredTokens {{uid:string, name:string, symbol:string}[]} The registered tokens from the storege
  * currentAddress {Object} The current unused address
+ * registeredNanoContracts {NanoContractData} The registered nano contracts from the storege
```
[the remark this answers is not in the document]
pedroferreira1: Thanks! Fixed

src/components/nano/ModalRegisterNanoContract.js
```diff
@@ -0,0 +1,124 @@
...
+  const [ncId, setNcId] = useState('');
+  const ncRegisterMetadata = useSelector(state => state.nanoContractsRegisterMetadata);
+  const ncRegisterStatus = get(ncRegisterMetadata, 'status', null);
+
+  const formRegisterNCRef = useRef(null);
+
+  const idRef = useRef(null);
+
+  useEffect(() => {
+    $('#registerNCModal').modal('show');
```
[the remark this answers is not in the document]
pedroferreira1: Done

src/components/nano/ModalRegisterNanoContract.js
```diff
@@ -0,0 +1,124 @@
...
+            <button type="button" className="btn btn-secondary" data-dismiss="modal">{t`Cancel`}</button>
+            <button onClick={handleRegister} type="button" className="btn btn-hathor">{t`Register`}</button>
+          </div>
+        </div>
+      </div>
+    </div>
+  );
+}
+
+export default ModalRegisterNanoContract;
```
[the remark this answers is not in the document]
pedroferreira1: Done

src/sagas/nanoContract.js
pedroferreira1: Why? This just details that this dispatch emits this action. If any other saga method wants to execute anything after this dispatch, it should take this action.

src/storage.js
pedroferreira1: Definitely. This was a bug, thanks.

src/reducers/index.js
```diff
@@ -130,6 +143,112 @@ const initialState = {
...
+   *     ],
+   *     historyMetadata: {
+   *       isLoading: false,
+   *       error: null,
+   *       after: '000075e15f015dc768065763acd9b563ec002e37182869965ff2c712bed83e1e',
+   *     },
+   *   },
+   * }
+   */
+  nanoContracts: {},
```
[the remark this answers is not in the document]
pedroferreira1: I read the document link, and it seems interesting to do this refactor. However, it's bigger than just moving the reducer part; this also requires changes in the actions.

I prefer keeping the same pattern we've been using for now, and discuss an improvement in the redux in general in the future.

What do you think?

src/sagas/nanoContract.js
pedroferreira1: Done

src/components/nano/ModalRegisterNanoContract.js
```diff
@@ -0,0 +1,124 @@
...
+            <h5 className="modal-title" id="exampleModalLabel">{t`Register a nano contract`}</h5>
+            <button type="button" className="close" data-dismiss="modal" aria-label="Close">
+              <span aria-hidden="true">&times;</span>
+            </button>
+          </div>
+          <div className="modal-body">
+            <p>{t`To register your nano contract, just write down the ID of the contract.`}</p>
+            <form ref={formRegisterNCRef}>
+              <div className="form-group">
+                <input autoFocus required type="text" className="form-control" ref={idRef} placeholder={t`Nano Contract ID`} />
```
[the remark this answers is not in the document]
pedroferreira1: I added the pattern regex to allow only hexadecimal with 64 chars. With this validation we don't allow whitespace, so it doesn't make sense to trim.

What do you think?

src/components/nano/ModalRegisterNanoContract.js
```diff
@@ -0,0 +1,124 @@
...
+    });
+
+    $('#registerNCModal').on('shown.bs.modal', (e) => {
+      idRef.current.focus();
+    });
+
+    return () => {
+      $('#registerNCModal').modal('hide');
+      // Removing all event listeners
+      $('#registerNCModal').off();
```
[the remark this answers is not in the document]
pedroferreira1: Ah, that's nice to have! Just refactored it, thanks

src/api/axiosInstance.js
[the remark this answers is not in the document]
pedroferreira1: Code updated: https://github.com/HathorNetwork/hathor-wallet/pull/529/commits/dfdad87afe39d40c5b464282fe16928de0d893e6

src/components/nano/ModalRegisterNanoContract.js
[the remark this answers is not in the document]
r4mmer: Makes sense

src/components/nano/ModalRegisterNanoContract.js
```diff
@@ -0,0 +1,119 @@
...
+
+  const idRef = useRef(null);
+
+  useEffect(() => {
+    $(`#${registerNCModalID}`).modal('show');
+    $(`#${registerNCModalID}`).on('hidden.bs.modal', (e) => {
+      dispatch(cleanNanoContractRegisterMetadata());
+      idRef.current.value = '';
+      // We always need to call on close when using global context modal
+      onClose(registerNCModalID);
```
tuliomir, line 45: question: Did you test this? I'm afraid this method requires the jQuery selector `#${registerNCModalID}`, not only the identifier.

It's actually an improvement point for the Global Modal, but currently I think this would not work properly.

src/reducers/index.js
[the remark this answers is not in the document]
tuliomir: Sure! I just wanted to raise the flag that this could be extracted, but it's a lot of effort to do so for the first time.

src/components/nano/ModalRegisterNanoContract.js
pedroferreira1: You are correct, fixed https://github.com/HathorNetwork/hathor-wallet/pull/529/commits/9c950018367442d1c13b45552ee6a685bb256b6b

tuliomir (approved):

PR #533, opened 2024-02-27
feat: start wallet with wallet-service

r4mmer:
### Acceptance Criteria
- Fix start wallet saga to start with wallet-service

### Security Checklist


src/sagas/featureToggle.js
```diff
@@ -196,6 +196,7 @@ export function* setupUnleashListeners(unleashClient) {
 }
 
 function mapFeatureToggles(toggles) {
+  console.log('feature toggles', toggles);
```
r4mmer, line 199: note: This is just to make debugging unleash related feature toggles easier

src/sagas/wallet.js
```diff
@@ -172,13 +173,14 @@ export function* startWallet(action) {
 
     if (!(words || xpub)) {
       const accessData = yield storage.getAccessData();
-      xpriv = cryptoUtils.decryptData(accessData.mainKey, pin);
+      xpriv = cryptoUtils.decryptData(accessData.acctPathKey, pin);
+      authxpriv = cryptoUtils.decryptData(accessData.authKey, pin);
```
r4mmer, line 177: note: To start a wallet-service you can use the seed or the account path xpriv along with the auth path key.

src/sagas/wallet.js
```diff
@@ -188,6 +190,7 @@ export function* startWallet(action) {
       }),
       passphrase,
       storage,
+      network,
```
r4mmer, line 193: note: network is required on the wallet-service facade constructor

src/utils/helpers.js
```diff
@@ -269,10 +269,21 @@ const helpers = {
...
-    // tx comes from getTxHistory and does not have token_data
-    // We need the actual history tx to access if it is an authority tx
-    const histTx = await wallet.getTx(tx.txId);
-    const isAllAuthority = this.isAllAuthority(histTx);
+
+    let isAllAuthority = false;
+    try {
+      // tx comes from getTxHistory and does not have token_data
+      // We need the actual history tx to access if it is an authority tx
+      const histTx = await wallet.getTx(tx.txId);
```
r4mmer, line 277: note: wallet-service facade does not implement `getTx`, it throws when called.

This was wrapped in a try..catch and `isAllAuthority` will default to false in this case.

pedroferreira1 (approved):

src/utils/helpers.js
r4mmer: I agree, and I believe that we should check this and other methods to have parity on the facades.

For instance, just implementing `getTx` would be bad since it uses an api call for each tx and the wallet facade gets it from memory (O(1) operation) since we use the `getTx` sequentially for each tx we want to show on screen it would severely impact the refresh rate.

andreabadesso (approved):

PR #537, opened 2024-03-01
Fix: Server change issues

tuliomir:
Three user behaviors related to network changes were triggering crashes on the application and were fixed.

## Acceptance Criteria
- Selecting a custom token and changing the network should not crash with an "Invalid Token" error ( [1e91a0a](https://github.com/HathorNetwork/hathor-wallet/pull/537/commits/1e91a0af1536f3d4086524187d4d1404aaec0e1e) )
- When recovering from being offline, the application tried to call a discontinued lib method and crashed ( [1d854c4](https://github.com/HathorNetwork/hathor-wallet/pull/537/commits/1d854c4c2c42c30b4455955f7264624db4478a91) )
- Unused methods cleanup ( also on [1d854c4](https://github.com/HathorNetwork/hathor-wallet/pull/537/commits/1d854c4c2c42c30b4455955f7264624db4478a91) )
- When resetting the wallet while on testnet, the lib configuration object became inconsistent and caused crashes ( [9de36b4](https://github.com/HathorNetwork/hathor-wallet/pull/537/commits/9de36b47f0b03f11c98c8faff3cc9cb347cfff7e) )

### Security Checklist

### DIFF
--- a/src/sagas/wallet.js
+++ b/src/sagas/wallet.js
@@ -195,7 +195,9 @@ export function* startWallet(action) {
   } else {
     // Set the server and network as saved on localStorage
     // If the localStorage is empty, fetch data directly from the lib config
-    config.setServerUrl(LOCAL_STORE.getServer() || config.getServerUrl());
+    const serverUrl = LOCAL_STORE.getServer() || config.getServerUrl();
+    config.setServerUrl(serverUrl);
+    LOCAL_STORE.setServers(serverUrl);
 
     connection = new Connection({
       network: network.name,
--- a/src/sagas/wallet.js
+++ b/src/sagas/wallet.js
@@ -194,7 +194,8 @@ export function* startWallet(action) {
     connection = wallet.conn;
   } else {
     // Set the server and network as saved on localStorage
-    config.setServerUrl(LOCAL_STORE.getServer());
+    // If the localStorage is empty, fetch data directly from the lib config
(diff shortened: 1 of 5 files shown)


src/sagas/wallet.js
```diff
@@ -194,7 +194,8 @@ export function* startWallet(action) {
     connection = wallet.conn;
   } else {
     // Set the server and network as saved on localStorage
-    config.setServerUrl(LOCAL_STORE.getServer());
+    // If the localStorage is empty, fetch data directly from the lib config
+    config.setServerUrl(LOCAL_STORE.getServer() || config.getServerUrl());
```
r4mmer, line 198: issue(blocking): If thats the case we should update the storage as well.

Since closing the app will load from localStorage we need to keep the localStorage updated with the current network.

r4mmer (changes requested):

src/sagas/wallet.js
tuliomir: I'm not sure this would be necessary, as the `server_url` configurations were correct right after wallet load. However, I think it's a good practice to update both the local store and the lib configuration together at all times until we unify them somehow.

Done on [2634e07](https://github.com/HathorNetwork/hathor-wallet/pull/537/commits/2634e070cf9675ec78a06528e36a11fa1d31b920)

r4mmer (approved):

pedroferreira1 (approved):

PR #546, opened 2024-03-05
fix: Ledger modal not updating on SendTokens

tuliomir:
The ledger modal was deciding its rendering state based on a component state that was not being updated correctly after the refactor on #483 .

The code was changed to decide the rendering based on function parameters instead of component states, so that the correct context is always provided directly. Also, the `txData` local variable was completely removed and re-calculated whenever necessary to reduce component complexity.

Because this transaction data parameter needed to be handled in the Ledger event handlers, a workflow that is different from the one of the component rendering, the `sendTransaction` variable was moved outside the component. 

### Acceptance Criteria
- The Ledger modal should work on the full transaction send flow
- Minor function name refactorings should make their responsibilities clearer

### Alternate solution: declarative modal exhibition
Another solution for this event handling would be to make the Ledger modal more declarative: its exhibition would be requested by `setState()` calls and `useEffect()` functions would monitor those states to decide when to exhibit the modal and which parameters to pass to it.

This would increase the amount of code changed, modify the business logic of the component and would not improve readability too much. So this solution was discarded for now.

### Security Checklist

### DIFF
--- a/src/screens/SendTokens.js
+++ b/src/screens/SendTokens.js
@@ -22,6 +22,12 @@ import { GlobalModalContext, MODAL_TYPES } from '../components/GlobalModal';
 import LOCAL_STORE from '../storage';
 import { useHistory } from 'react-router-dom';
 
+/** @typedef {0|1} LEDGER_MODAL_STATE */
+const LEDGER_MODAL_STATE = {
+  WAITING_APPROVAL: 0,
+  LEDGER_APPROVED: 1,
+}
+
 /**
  * Screen used to send tokens to another wallet.
  * Can send more than one token in the same transaction.
@@ -195,7 +201,7 @@ function SendTokens() {
       globalModalContext.showModal(MODAL_TYPES.ALERT, {
         id: 'ledgerAlertModal',
         title: t`Validate outputs on Ledger`,
-        body: renderAlertBody(sendTransaction),
+        body: renderAlertBody(LEDGER_MODAL_STATE.LEDGER_APPROVED, sendTransaction),
         showFooter: false,
       });
     } catch(e) {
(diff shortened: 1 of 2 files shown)


andreabadesso (approved):

src/screens/Wallet.js
```diff
@@ -269,6 +269,7 @@ function Wallet() {
     context.showModal(MODAL_TYPES.LEDGER_SIGN_TOKEN, {
       token,
       modalId: 'signTokenDataModal',
+      cb: (newValue) => console.log(`New token signature state: ${newValue}`)
```
tuliomir, line 272: The `cb` parameter is mandatory for this modal, and was causing a crash because it wasn't informed. However, it has no application for this screen, so it's only used as a dummy output.

andreabadesso: Ok, but do we really need this console.log here?

andreabadesso (approved):

src/screens/Wallet.js
tuliomir: We don't! Replaced it with a dummy function and an explanatory comment on https://github.com/HathorNetwork/hathor-wallet/pull/546/commits/a1f9556eaa907d5895647f4c4f2ca0476836224b

src/screens/SendTokens.js
```diff
@@ -492,8 +497,14 @@ function SendTokens() {
...
-  const renderAlertBody = () => {
-    if (ledgerStep === 0) {
+  /**
+   * Renders the body for the ledger alert modal, with instructions on how to approve the tx.
+   * When called with the `sendTransaction` parameter, it renders a transaction success monitoring
+   * component instead, using this parameter as a reference for which transaction to listen to.
+   * @param {SendTransaction} [sendTransactionObj] Reference to tx already signed by Ledger
+   */
+  const renderAlertBody = (sendTransactionObj) => {
+    if (!sendTransactionObj) {
```
r4mmer, line 507: suggestion: Use a number/enum/boolean argument instead of a `SendTransaction` instance.

I understand why using `sendTransactionObj` is nice, since the `else` block requires it, but the object is created before we send the data to the Ledger device it is just not passed to the method.
We could have both as arguments `(ledgerStep, sendTransactionObj)`, this way the caller would be in control of the rendered content.

tuliomir: Done on https://github.com/HathorNetwork/hathor-wallet/pull/546/commits/5fa10de002266ca6a137076c4f49c7df69b0fc81

r4mmer (approved):

PR #547, opened 2024-03-05
fix: Network name inconsistency

tuliomir:
This is a complement of #537 : , it fixed all tested cases for software wallets. However when testing hardware wallets on `testnet` some new errors appeared. This PR fixes them.

The cause is that the _network name_ was never updated on local store for a hardware wallet.

### Acceptance Criteria
- The application should run a hardware wallet successfully on `testnet`
- After closing and opening the application again, the application should not throw any errors and continue on `testnet`

### Security Checklist

### DIFF
--- a/src/sagas/wallet.js
+++ b/src/sagas/wallet.js
@@ -124,7 +124,6 @@ export function* startWallet(action) {
     }
   }
 
-  const network = config.getNetwork();
   const storage = LOCAL_STORE.getStorage();
 
   // We are offline, the connection object is yet to be created
@@ -146,6 +145,13 @@ export function* startWallet(action) {
   // If we've lost redux data, we could not properly stop the wallet object
   // then we don't know if we've cleaned up the wallet data in the storage
   yield storage.cleanStorage(true, true);
+
+  // Set the server and network as saved on localStorage
+  // If the localStorage is empty, fetch data directly from the lib config
+  const networkName = LOCAL_STORE.getNetwork() || config.getNetwork().name;
+  const serverUrl = LOCAL_STORE.getServer() || config.getServerUrl();
+  config.setServerUrl(serverUrl);
+  LOCAL_STORE.setServers(serverUrl);
 
   let wallet, connection;
   if (useWalletService) {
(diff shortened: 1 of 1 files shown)


r4mmer (approved):

src/sagas/wallet.js
```diff
@@ -147,6 +146,13 @@ export function* startWallet(action) {
   // then we don't know if we've cleaned up the wallet data in the storage
   yield storage.cleanStorage(true, true);
 
+  // Set the server and network as saved on localStorage
+  // If the localStorage is empty, fetch data directly from the lib config
+  const networkName = LOCAL_STORE.getNetwork() || config.getNetwork().name;
+  const serverUrl = LOCAL_STORE.getServer() || config.getServerUrl();
+  config.setServerUrl(serverUrl);
```
pedroferreira1, line 153: Does it make sense to set network in the lib config?

src/sagas/wallet.js
```diff
@@ -247,16 +247,16 @@ export function* startWallet(action) {
...
-    let networkName = network.name;
+    let serverNetworkName = networkName;
 
     if (serverInfo) {
       version = serverInfo.version;
-      networkName = serverInfo.network && serverInfo.network.split('-')[0];
+      serverNetworkName = serverInfo.network && serverInfo.network.split('-')[0];
     }
 
     yield put(setServerInfo({
```
pedroferreira1, line 257: Why is this information used? The `networkName` from the lib might be different from the server

tuliomir: This information is later used by the `fetchTokenMetadata()` saga:
https://github.com/HathorNetwork/hathor-wallet/blob/9a477a368bc10415325b29d2e52dc9ed566731fc/src/sagas/tokens.js#L161-L167

The network name obtained directly from the server may have a suffix, and that's why we treat it with a `.split('-')[0]`, but it should not be different from the network name.

Do you see any cases where the names may not match and cause issues?

src/sagas/wallet.js
tuliomir: Good point! Setting the retrieved network name on both the lib and local storage to keep them synchronized. ( https://github.com/HathorNetwork/hathor-wallet/pull/547/commits/03e741100bed24f05664708f85c4b7074b7216dc )

src/sagas/wallet.js
pedroferreira1: Yes, the nano network in the full node is `nano-testnet-alpha`. This will add conflict to my wallet desktop code for nano but I can handle it there, no problem.

pedroferreira1 (approved):

PR #554, opened 2024-03-22
fix: hardware wallet migration

r4mmer:
### Acceptance Criteria

- Clean localStorage when migrating a wallet that was initialized from a hardware wallet.
- Migrate important keys

### Security Checklist

### DIFF
--- a/public/preload.js
+++ b/public/preload.js
@@ -18,19 +18,14 @@ Sentry.init({
 process.once('loaded', () => {
   const oldAccessDataRaw = localStorage.getItem('wallet:accessData');
   if (oldAccessDataRaw) {
-    // When migrating old wallets initialized with hardware wallets we need to clean the localStorage
+    // When migrating from wallets with version v0.26.0 and older initialized with hardware devices
+    // we need to clean the localStorage since some state left over can cause issues
     const accessData = JSON.parse(oldAccessDataRaw);
     const isHardware = localStorage.getItem('wallet:type') === 'hardware';
     if (accessData.from_xpub || isHardware) {
       const uniqueId = localStorage.getItem('app:uniqueId');
-      const backup = localStorage.getItem('wallet:backup');
-      const network = localStorage.getItem('wallet:network');
-      const server = localStorage.getItem('wallet:server');
       localStorage.clear();
       localStorage.setItem('localstorage:started', 'true');
-      localStorage.setItem('localstorage:backup', backup);
-      localStorage.setItem('localstorage:network', network);
-      localStorage.setItem('localstorage:server', server);
       localStorage.setItem('app:uniqueId', uniqueId);
     }
   }
--- a/public/preload.js
+++ b/public/preload.js
@@ -16,6 +16,24 @@ Sentry.init({
 })
 
 process.once('loaded', () => {
+  const oldAccessDataRaw = localStorage.getItem('wallet:accessData');
+  if (oldAccessDataRaw) {
+    // When migrating old wallets initialized with hardware wallets we need to clean the localStorage
+    const accessData = JSON.parse(oldAccessDataRaw);
+    const isHardware = localStorage.getItem('wallet:type') === 'hardware';
+    if (accessData.from_xpub || isHardware) {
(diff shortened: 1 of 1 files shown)


public/preload.js
```diff
@@ -16,6 +16,26 @@ Sentry.init({
...
 
 process.once('loaded', () => {
+  const oldAccessDataRaw = localStorage.getItem('wallet:accessData');
+  if (oldAccessDataRaw) {
+    // When migrating old wallets initialized with hardware wallets we need to clean the localStorage
+    const accessData = JSON.parse(oldAccessDataRaw);
+    const isHardware = localStorage.getItem('wallet:type') === 'hardware';
+    if (accessData.from_xpub || isHardware) {
+      const uniqueId = localStorage.getItem('app:uniqueId');
+      const backup = localStorage.getItem('wallet:backup');
```
pedroferreira1, line 26: Why do you need this attribute? This marks if the software wallet backup has been done, isn't it?

public/preload.js
```diff
@@ -16,6 +16,26 @@ Sentry.init({
...
+      const network = localStorage.getItem('wallet:network');
+      const server = localStorage.getItem('wallet:server');
+      localStorage.clear();
+      localStorage.setItem('localstorage:started', 'true');
+      localStorage.setItem('localstorage:backup', backup);
+      localStorage.setItem('localstorage:network', network);
+      localStorage.setItem('localstorage:server', server);
+      localStorage.setItem('app:uniqueId', uniqueId);
+    }
```
pedroferreira1, line 35: Don't you need to also save the wallet type as hardware?

public/preload.js
```diff
@@ -16,6 +16,26 @@ Sentry.init({
 })
 
 process.once('loaded', () => {
+  const oldAccessDataRaw = localStorage.getItem('wallet:accessData');
+  if (oldAccessDataRaw) {
+    // When migrating old wallets initialized with hardware wallets we need to clean the localStorage
```
pedroferreira1, line 21: Can you add here the version that you are migrating this data from?

public/preload.js
pedroferreira1, line 35: I was thinking about moving this to a separate function just to organize the code. Then we can have `migrate_from_vX.Y()` and below we can have a different migration function in the future, so they are not mixed when we need to add more storage migrations.

What do you think?

r4mmer: No, the user will choose hardware wallet and start with the device, then it will be set as well.

r4mmer: Maybe, but this is not a migration since we are not porting the access data.
We are just clearing the localStorage and keeping some basic flags which could also be removed.

public/preload.js
```diff
@@ -16,6 +16,26 @@ Sentry.init({
...
+    if (accessData.from_xpub || isHardware) {
+      const uniqueId = localStorage.getItem('app:uniqueId');
+      const backup = localStorage.getItem('wallet:backup');
+      const network = localStorage.getItem('wallet:network');
+      const server = localStorage.getItem('wallet:server');
+      localStorage.clear();
+      localStorage.setItem('localstorage:started', 'true');
+      localStorage.setItem('localstorage:backup', backup);
+      localStorage.setItem('localstorage:network', network);
+      localStorage.setItem('localstorage:server', server);
```
tuliomir, line 33: question: Based on our experience with the recent #537 and #547 bugfixes, should we also update the wallet lib's `server` and `networkName` configurations here?

r4mmer: This has since been removed, the only flags being kept are the started and uniqueId.
All others will be set during the normal wallet process.

public/preload.js
r4mmer: Done

tuliomir (approved):

public/preload.js
pedroferreira1: The `uniqueId` is good to have but the backup I don't see the need. I think we can keep some, the `started` attribute moves the startup screen to be the wallet type screen, for example.

pedroferreira1 (approved):

PR #560, opened 2024-03-25
feat: reset data menu

r4mmer:
### Acceptance Criteria
- Create a menu that offers the possibility to reset all wallet data for the user

### Media

> Normal call to reset

https://github.com/HathorNetwork/hathor-wallet/assets/15909384/ddf53d0f-8e00-4c81-9fc9-87cac17131c0

> Calling to reset during a bug

https://github.com/HathorNetwork/hathor-wallet/assets/15909384/ffeaffff-c88d-441d-bfaf-cea835ae4233

PR #562, opened 2024-03-27
docs: hardware wallet migration qa

r4mmer:
### Acceptance Criteria
- Hardware wallet migration QA steps

### Security Checklist

### DIFF
--- a/QA.md
+++ b/QA.md
@@ -222,7 +222,7 @@ You can connect your wallet to the testnet (https://node1.foxtrot.testnet.hathor
     1. Deny the authorization request on Ledger (scroll with left or right and click both buttons on "Reject" step). It should show an error on the wallet.
     1. Click 'Try again'. It goes through both steps and asks for authorization again.
     1. Grant authorization request (click both buttons on the "Approve" step). It will proceed to 'Loading transactions' screen.
-    1. Once the wallet is loaded check that the current address match the copied address during [**Hardware wallet migration**](#hardware-wallet-migration)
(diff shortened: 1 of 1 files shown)


pedroferreira1 (approved):

QA.md
```diff
@@ -205,8 +205,14 @@ You can connect your wallet to the testnet (https://node1.foxtrot.testnet.hathor
     1. Close and open the wallet again and start a new wallet one without doing backup. It must show a yellow warning saying a backup must be done.
     1. Do the backup (following procedures in the 'Initialization' tests). The backup message has to disappear.
 
+1. **Hardware wallet migration**
+    1. Reset the wallet one more time with `localStorage.clear()`.
```
tuliomir, line 209: suggestion: We could already mention the reset wallet menu from #560 instead of directing the user to the console

The only QA this step would fail is this very one we're doing right now. All future versions will have this easier path to test.

QA.md
```diff
@@ -205,8 +205,14 @@ You can connect your wallet to the testnet (https://node1.foxtrot.testnet.hathor
...
+    1. Uninstall the wallet and install the last version.
```
tuliomir, line 210:
```suggestion
    1. Uninstall the wallet and install the previous version.
```
suggestion: The word `last` is correct here, but could cause confusion when compared with `latest`. Maybe `previous` would be clearer

r4mmer: Done

tuliomir (approved):

Issue #441, opened 2023-11-22
Upgrading to `react-scripts@5`


tuliomir:
## Summary
`react-scripts` is one of the dependencies contained with the `create-react-app` bundle that was used to create this application.

According to the [`create-react-app@5.0.0` release announcement](https://github.com/facebook/create-react-app/releases/tag/v5.0.0), a breaking change was added in that the WebPack version being used is now 5.0. This means that the NodeJs globals such as `fs` and `crypto` are now inacessible.

So, as the current state of the code, it is unable to run after an upgrade to `react-scripts` v5.

The `react-scripts` are being updated to version 4 as of now, and future dependency upgrades are discussed below. The suggested upgrade path would be:
- Staying in v4 for the short term
- Upgrading to v5 on the mid term ( solution 3 )
- Migrating away from CRA on the long term ( solution 4 ), if the detailed conditions are maintained

## Solution 1: Polyfill and keep code
It would be possible to upgrade to WebPack v5 and add polyfills if we used it directly.

However, our configuration of WebPack is entirely done through the react scripts, and the config file is not accessible to the developer directly. To fix this, a solution would be to add each polyfill as a separate dependency and integrate them through another dependency such as [`react-app-rewired`](https://www.npmjs.com/package/react-app-rewired).

This adds multiple dependencies to the application and requires the entire `react-scripts` to be executed in the context of the _rewired_ configuration, which adds security concerns to the application at runtime. See below examples of this implementation:
- The [`web3auth.io` troubleshooting guide](https://web3auth.io/docs/troubleshooting/webpack-issues#react-create-react-app) for WebPack issues
- An [example article](https://alchemy.com/blog/how-to-polyfill-node-core-modules-in-webpack-5) explaining the situation and solution steps
- Issue [11756 on `create-react-app`](https://github.com/facebook/create-react-app/issues/11756) where solutions are being discussed

## Solution 2: Fork `react-scripts`
This is actually a solution common enough to have its own entry on the official docs ( see [Alternatives to Ejecting](https://create-react-app.dev/docs/alternatives-to-ejecting) ).

This, however, would require our own maintenance of a complex third-party library and does not seem reasonable for our usage.
## Solution 3: Refactor to a web-based application
This solution is the most recommended given enough development time: adapt the application to avoid using all NodeJS modules and move entirely to a browser-based environment.

[...]

PR #567, opened 2024-03-28
chore: Upgrades lockfile to version 3

tuliomir:
### Summary
This PR continues the work made on #501 and upgrades the application lockfile, dealing with incompatibilities that arise from the upgrade.

Prior to this PR, the `npm install` command could only be run on Node 14. If a developer tried to tun `npm install` on Node 20 the whole lockfile would be rebuilt and compatibility errors would occur if any other dependency was changed.

### Steps description
We were not able to upgrade this file previous to #555 because of an incompatibility in one of the subdependencies of the React Router.

The first step was to only upgrade the lockfile version without changing any of its dependencies, which was done using `npm install --lockfile-version 3` command. ( https://github.com/HathorNetwork/hathor-wallet/pull/567/commits/e603860f75a61ca1b4e0289b87727e484b8785ca )

The next step was to actually run `npm install` on Node 20, which prompted many changes to the lockfile to respect the new environment. This command was previously run only on Node 14. ( https://github.com/HathorNetwork/hathor-wallet/pull/567/commits/813bf7f167319a31521ce0dc4b77d348e6b592da ).

As a consequence, the `pot` translation files failed and had to be updated using `make update_pot` ( https://github.com/HathorNetwork/hathor-wallet/pull/567/commits/5e362c42b3c2c704cbd1beea562aaa9ebf3ac126 )

Lastly, `react-scripts` began to fail all jest tests, because react-scripts `v4` requires Jest `v26` and our React `18` tests require Jest `v29`. We tried using the `SKIP_PREFLIGHT_CHECK=true` flag to allow the version mismatch not to break the CI, but it utimately broke the test run itself. Another approach would be to rewrite the tests using an older version of the react testing library.

Since the tests were still on a stub implementation that do not improve our quality control, they were disabled until we can upgrade `react-scripts`, as described on #441. ( https://github.com/HathorNetwork/hathor-wallet/pull/567/commits/704148fbd9fd6582c3b432e42d46328c34318cb8 )

### Acceptance Criteria
- `package-lock.json` should be upgraded to v3
- Running `npm install` on a development environment should not change the contents of `package-lock.json`

### Security Checklist


locale/texts.pot
```diff
@@ -3,7 +3,7 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
```
andreabadesso, line 3: > The next step was to actually run npm install on Node 20, which prompted many changes to the lockfile to respect the new environment. This command was previously run only on Node 14. ( https://github.com/HathorNetwork/hathor-wallet/commit/813bf7f167319a31521ce0dc4b77d348e6b592da ).

> As a consequence, the pot translation files failed and had to be updated using make update_pot ( https://github.com/HathorNetwork/hathor-wallet/commit/5e362c42b3c2c704cbd1beea562aaa9ebf3ac126 )


I don't understand how one thing can be a consequence of the other, why would the line numbers change here?

andreabadesso (changes requested):

locale/texts.pot
pedroferreira1: What was the error?

tuliomir: Opened #577 to follow this issue.

pedroferreira1 (approved):

andreabadesso (approved):

PR #573, opened 2024-04-03
Refactor: Removes `routerHistory` from Redux store

tuliomir:

The `routerHistory` property stored is currently used for navigating inside our Sagas. The proposed solution is to:
- Create a `navigateTo` property on the store
- Create a `useEffect()` inside the `App.js` screen listening to changes on this property
- `App.js` becomes responsible for navigating whenever a background task requests it to

### Acceptance Criteria
- The `routerHistory` should not be present on the Redux Store

### Security Checklist

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -44,7 +44,7 @@ import hathorLib from '@hathor/wallet-lib';
 import { IPC_RENDERER } from './constants';
 import AddressList from './screens/AddressList';
 import NFTList from './screens/NFTList';
-import { setNavigateTo, updateLedgerClosed } from './actions/index';
+import { resetNavigateTo, updateLedgerClosed } from './actions/index';
 import { WALLET_STATUS } from './sagas/wallet';
 import ProposalList from './screens/atomic-swap/ProposalList';
 import EditSwap from './screens/atomic-swap/EditSwap';
@@ -173,7 +173,7 @@ function Root() {
 
     // Navigate to the informed route and reset the navigateTo state property
     navigate(navigateTo.route, { replace: navigateTo.replace });
-    dispatch(setNavigateTo(''));
+    dispatch(resetNavigateTo());
   }, [navigateTo])
 
   // Handles failed wallet states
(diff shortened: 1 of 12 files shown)


src/App.js
```diff
@@ -160,6 +162,20 @@ function Root() {
...
+   */
+  useEffect(() => {
+    // Ignore this effect if there is no route to navigateTo
+    if (!navigateTo.route) {
+      return;
+    }
+
+    // Navigate to the informed route and reset the navigateTo state property
+    navigate(navigateTo.route, { replace: navigateTo.replace });
+    dispatch(setNavigateTo(''));
```
andreabadesso, line 176: Why an empty string? Why not`{ route: '', replace: false }`;

Or dispatch an action "resetNavigateTo" that sets `navigateTo` to `initialState.navigateTo`

tuliomir: Good point. Implemented `resetNavigateTo()` on https://github.com/HathorNetwork/hathor-wallet/pull/573/commits/0ded8d6511bfa75f55ae9e29fdaa063427f80492

src/sagas/wallet.js
```diff
@@ -537,11 +534,7 @@ export function* handleNewTx(action) {
   // Set the notification click, in case we have sent one
   if (notification !== undefined) {
     notification.onclick = () => {
-      if (!routerHistory) {
-        return;
-      }
-
-      routerHistory(`/transaction/${tx.tx_id}/`);
+      put(setNavigateTo(`/transaction/${ tx.tx_id }/`))
```
pedroferreira1, line 537: Does it work without `yield`?

pedroferreira1 (approved):

src/sagas/wallet.js
tuliomir: It does because this is actually executed in a callback for a `onClick` event, not part of the `handleNewTx()` generator function.

It only benefits from being in the same scope as the generator to be able to use the `put` function.

PR #574, opened 2024-04-05
feat: filter HTR out when fetching token metadata

r4mmer:
### Acceptance Criteria

- Do not fetch HTR token metadata from the explorer-service.

### Security Checklist

### DIFF
--- a/src/sagas/wallet.js
+++ b/src/sagas/wallet.js
@@ -373,7 +373,7 @@ export function* loadTokens() {
   // spawn a new "thread" to handle it.
   //
   // `spawn` is similar to `fork`, but it creates a `detached` fork
-  yield spawn(fetchTokensMetadata, registeredTokens.map(token => token.uid));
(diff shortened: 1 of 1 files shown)


src/sagas/wallet.js
```diff
@@ -373,7 +373,7 @@ export function* loadTokens() {
   // spawn a new "thread" to handle it.
   //
   // `spawn` is similar to `fork`, but it creates a `detached` fork
-  yield spawn(fetchTokensMetadata, registeredTokens.map(token => token.uid));
+  yield spawn(fetchTokensMetadata, registeredTokens.map(token => token.uid).filter(token => token !== htrUid));
```
luislhl, line 376: Is it possible to have a unit test for this?

pedroferreira1 (approved):

src/sagas/wallet.js
r4mmer: I dont think the desktop wallet has unit tests yet, only e2e with cypress

PR #587, opened 2024-04-22
Refactor: State tokens to a serializable array

tuliomir:

### Acceptance Criteria
- Tokens state on Redux must be serializable

### Security Checklist

### DIFF
--- a/src/reducers/index.js
+++ b/src/reducers/index.js
@@ -72,7 +72,11 @@ const initialState = {
   tokens: [hathorLib.constants.HATHOR_TOKEN_CONFIG],
   // Token selected (by default is HATHOR)
   selectedToken: hathorLib.constants.HATHOR_TOKEN_CONFIG.uid,
-  // List of all tokens seen in transactions
+  /**
+   * List of all tokens seen in transactions
+   * @type {Record<string, string>}
+   * @example { 00: "00", abc123: "abc123" }
+   */
   allTokens: {},
   // If is in the proccess of loading addresses transactions from the full node
   // When the request to load addresses fails this variable can continue true
@@ -287,13 +291,20 @@ const getTxHistoryFromWSTx = (tx, tokenUid, tokenTxBalance) => {
 
 /**
  * Got wallet history. Update wallet data on redux
+ * @param {string[]} action.payload.tokens
+ * @param {string} action.payload.currentAddress
+ * @param {{uid:string, name:string, symbol:string}[]} action.payload.registeredTokens
  */
 const onLoadWalletSuccess = (state, action) => {
   // Update the version of the wallet that the data was loaded
   LOCAL_STORE.setWalletVersion(VERSION);
   const { tokens, registeredTokens, currentAddress } = action.payload;
-  const allTokens = new Set(tokens);
 
+  // Convert the tokens list to a new Set to ensure there are no duplicated tokens
+  const allTokens = new Set();
+  for (const key in tokens) {
+    allTokens.add(tokens[key]);
+  }
   return {
     ...state,
     loadingAddresses: false,
@@ -511,12 +522,18 @@ export const resetSelectedTokenIfNeeded = (state, action) => {
   return state;
 };
 
-/*
+/**
  * Used when registering or creating tokens to update the wallet token list.
-*/
+ * @param {Record<string,string>} state.allTokens - Current list of allTokens
+ * @param {string} action.payload.uid - UID of token to add
+ */
 export const onNewTokens = (state, action) => {
+  // Convert `allTokens` to a Set to prevent duplicates
+  const allTokensSet = new Set();
+  for (const token of Object.keys(state.allTokens)) {
+    allTokensSet.add(token);
+  }
   // Add new created token to the all tokens set
-  const allTokensSet = new Set(state.allTokens);
   allTokensSet.add(action.payload.uid);
 
   return {
--- a/src/reducers/index.js
+++ b/src/reducers/index.js
@@ -73,7 +73,7 @@ const initialState = {
   // Token selected (by default is HATHOR)
   selectedToken: hathorLib.constants.HATHOR_TOKEN_CONFIG.uid,
   // List of all tokens seen in transactions
-  allTokens: new Set(),
+  allTokens: {},
   // If is in the proccess of loading addresses transactions from the full node
   // When the request to load addresses fails this variable can continue true
   loadingAddresses: false,
@@ -299,7 +299,7 @@ const onLoadWalletSuccess = (state, action) => {
     loadingAddresses: false,
     lastSharedAddress: currentAddress.address,
     lastSharedIndex: currentAddress.index,
-    allTokens,
+    allTokens: helpersUtils.convertSetToObject(allTokens),
     tokens: registeredTokens,
   };
 };
@@ -516,12 +516,14 @@ export const resetSelectedTokenIfNeeded = (state, action) => {
 */
 export const onNewTokens = (state, action) => {
   // Add new created token to the all tokens set
-  state.allTokens.add(action.payload.uid);
+  const allTokensSet = new Set(state.allTokens);
+  allTokensSet.add(action.payload.uid);
 
   return {
     ...state,
     selectedToken: action.payload.uid,
     tokens: action.payload.tokens,
+    allTokensSet: helpersUtils.convertSetToObject(allTokensSet),
   };
 };
 
--- a/src/reducers/index.js
+++ b/src/reducers/index.js
@@ -1016,20 +1016,23 @@ export const onUpdateTxHistory = (state, action) => {
     return state;
   }
 
+  const newTokenHistoryData = [...tokenHistory.data];
   for (const [index, histTx] of tokenHistory.data.entries()) {
     if (histTx.tx_id === tx.tx_id) {
-      tokenHistory.data[index] = getTxHistoryFromWSTx(tx, tokenId, balance);
+      newTokenHistoryData[index] = getTxHistoryFromWSTx(tx, tokenId, balance);
       break;
     }
   }
+  const newTokenHistory = {
+    ...tokenHistory,
+    data: newTokenHistoryData,
+  };
 
   return {
     ...state,
     tokensHistory: {
       ...state.tokensHistory,
-      [tokenId]: {
-        ...tokenHistory,
-      }
+      [tokenId]: newTokenHistory,
     },
   };
 };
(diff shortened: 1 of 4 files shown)


src/reducers/index.js
```diff
@@ -287,19 +291,26 @@ const getTxHistoryFromWSTx = (tx, tokenUid, tokenTxBalance) => {
...
+  for (const key in tokens) {
+    allTokens.add(tokens[key]);
+  }
   return {
     ...state,
     loadingAddresses: false,
     lastSharedAddress: currentAddress.address,
     lastSharedIndex: currentAddress.index,
-    allTokens,
+    allTokens: helpersUtils.convertSetToObject(allTokens),
```
r4mmer, line 313: Do we need this?

If the `allTokens` object will have key and value as the uid, it means that it already will filter out possible duplicates.
We could just have:

```javascript
const allTokens = {};
for (const key in tokens) {
  allTokens[key] = key;
}
```

And the allTokens below would work the same way, right?

src/reducers/index.js
```diff
@@ -511,17 +522,25 @@ export const resetSelectedTokenIfNeeded = (state, action) => {
...
+ */
 export const onNewTokens = (state, action) => {
+  // Convert `allTokens` to a Set to prevent duplicates
+  const allTokensSet = new Set();
+  for (const token of Object.keys(state.allTokens)) {
+    allTokensSet.add(token);
+  }
   // Add new created token to the all tokens set
-  state.allTokens.add(action.payload.uid);
+  allTokensSet.add(action.payload.uid);
```
r4mmer, line 537: I think this is better:
```js
  const allTokensSet = new Set(Object.keys(state.allTokens));
```

But if this is just to convert to set and remove possible duplicates maybe we could just do:

```js
state.allTokens[action.payload.uid] = action.payload.uid;
```

This way we don't need to convert to set and have the same effect.

There is also:

```js
 const newAllTokens = {...state.allTokens, [action.payload.uid]: action.payload.uid};
```

Now we create a new object, so redux can detect the change.

src/reducers/index.js
```diff
@@ -511,17 +522,25 @@ export const resetSelectedTokenIfNeeded = (state, action) => {
...
 
   return {
     ...state,
     selectedToken: action.payload.uid,
     tokens: action.payload.tokens,
+    allTokensSet: helpersUtils.convertSetToObject(allTokensSet),
```
r4mmer, line 543: Should the key name be `allTokensSet`?

src/sagas/wallet.js
```diff
@@ -380,7 +380,7 @@ export function* loadTokens() {
   //
   // Note: We need to download the balance of all the tokens from the wallet so we can
   // hide zero-balance tokens
-  for (const token of allTokens) {
+  for (const token of Object.values(allTokens)) {
```
r4mmer, line 383: I know allTokens key and value are the same, but I think we should have a consistent usage, maybe change this to `Object.keys`

src/sagas/wallet.js
```diff
@@ -674,7 +674,7 @@ export function* walletReloading() {
 
     // We might have lost transactions during the reload, so we must invalidate the
     // token histories:
-    for (const tokenUid of allTokens) {
+    for (const tokenUid of Object.values(allTokens)) {
```
r4mmer, line 677: Same here

src/utils/helpers.js
```diff
@@ -375,6 +375,21 @@ const helpers = {
...
+   * @returns {Record<string, unknown>}
+   */
+  convertSetToObject(set) {
+    let obj = {};
+
+    for (const item of set) {
+      obj[item] = item;
+    }
+
+    return obj;
```
r4mmer, line 392: Do we need this?

Since a set will be converted to an object with the same key and value maybe we could just do `Object.fromEntries(set.entries())` and achieve the same result as this function.

The reverse is even simpler, since `new Set(obj.keys())` will create the set.

src/utils/wallet.js
```diff
@@ -229,7 +229,7 @@ const wallet = {
     const unknownTokens = [];
 
     // Iterating tokens to filter unregistered ones
-    for (const tokenUid of allTokens) {
+    for (const tokenUid of Object.values(allTokens)) {
```
r4mmer, line 232: We should use the `Object.keys`

tuliomir: Done on https://github.com/HathorNetwork/hathor-wallet/pull/587/commits/3ee5ac5cfba3a1e7dbe5387109159d06d4a34485

src/sagas/wallet.js
tuliomir: Done on https://github.com/HathorNetwork/hathor-wallet/pull/587/commits/3ee5ac5cfba3a1e7dbe5387109159d06d4a34485

src/reducers/index.js
tuliomir: No, it shouldn't! It was certainly a leftover from some experimentation and was broken as it was.
Fixed on https://github.com/HathorNetwork/hathor-wallet/pull/587/commits/3ee5ac5cfba3a1e7dbe5387109159d06d4a34485

src/reducers/index.js
tuliomir: That's a good approach. I implemented it on https://github.com/HathorNetwork/hathor-wallet/pull/587/commits/3ee5ac5cfba3a1e7dbe5387109159d06d4a34485

src/reducers/index.js
tuliomir: My implementation on https://github.com/HathorNetwork/hathor-wallet/pull/587/commits/3ee5ac5cfba3a1e7dbe5387109159d06d4a34485 was close to this one, but with more steps for clarity.

What do you think?

src/utils/helpers.js
tuliomir: Removed this function on https://github.com/HathorNetwork/hathor-wallet/pull/587/commits/3ee5ac5cfba3a1e7dbe5387109159d06d4a34485 , as the new approach doesn't need it.

src/reducers/index.js
```diff
@@ -287,19 +291,26 @@ const getTxHistoryFromWSTx = (tx, tokenUid, tokenTxBalance) => {
...
   // Update the version of the wallet that the data was loaded
   LOCAL_STORE.setWalletVersion(VERSION);
   const { tokens, registeredTokens, currentAddress } = action.payload;
-  const allTokens = new Set(tokens);
 
+  // Convert the tokens list to a new object to ensure there are no duplicated tokens
+  const allTokens = {};
+  for (const uid of tokens) {
+    allTokens[uid] = uid;
+  }
```
andreabadesso, line 307: I don't like this in the reducer, I think that the `tokens` should already be in the converted format when they reach this point

Can you move it to the saga?

It would also be nice to move this `LOCAL_STORE.setWalletVersion(VERSION);`, which is a side-effect to the saga, it shouldn't be here

src/reducers/index.js
```diff
@@ -511,17 +522,23 @@ export const resetSelectedTokenIfNeeded = (state, action) => {
...
+ * @param {string} action.payload.uid - UID of token to add
+ */
 export const onNewTokens = (state, action) => {
-  // Add new created token to the all tokens set
-  state.allTokens.add(action.payload.uid);
+  // Convert `allTokens` to a Set to prevent duplicates
+  const allTokens = {...state.allTokens};
+  // Add new created token to the all tokens object
+  const newTokenUid = action.payload.uid;
+  allTokens[newTokenUid] = newTokenUid;
```
andreabadesso, line 535: Same here... I don't think reducers should de-duplicate anything

I understand that they're still pure functions, but makes it harder to maintain

andreabadesso (changes requested):

src/reducers/index.js
r4mmer: The spread is not converting to Set, it is actually an object and adding the token to the object.
The de-duplication is done implicitly if the key already exists.

src/reducers/index.js
```diff
@@ -511,17 +522,23 @@ export const resetSelectedTokenIfNeeded = (state, action) => {
...
+/**
  * Used when registering or creating tokens to update the wallet token list.
-*/
+ * @param {Record<string,string>} state.allTokens - Current list of allTokens
+ * @param {string} action.payload.uid - UID of token to add
+ */
 export const onNewTokens = (state, action) => {
-  // Add new created token to the all tokens set
-  state.allTokens.add(action.payload.uid);
+  // Convert `allTokens` to a Set to prevent duplicates
```
r4mmer, line 531:
```suggestion
  // Create a new object to prevent false negative on object comparisons
```

React will not trigger updates if the "dependency" (of hooks) do not change and object comparisons do not check contents, only that it is the same instance.
This is why the best approach is to create a new object instead of updating the existing one.

r4mmer (approved):

src/reducers/index.js
r4mmer: I usually prefer oneliners, but splitting for better readability is always nice!

src/reducers/index.js
tuliomir: I agree the reducer is not the place to transform an array into an object map, so I extracted the logic for the deduplication of allTokens on https://github.com/HathorNetwork/hathor-wallet/pull/587/commits/6b87012bfa353140c7a864a832f81fdc097a799d .

However this comes with a tradeoff: It is now the responsibility of the action caller to pass a new object as a parameter.

As it was mentioned on [a comment below](https://github.com/HathorNetwork/hathor-wallet/pull/587#discussion_r1576573372), if the object is not completely new ( for example, the caller function mutates the object and passes it again as a parameter ), weird behaviors may happen on the application because of a lack of re-renders.

This should be improved as soon as we have the full Redux tools available, after the major upgrade of the Redux dependency.

The `LOCAL_STORE.setWalletVersion` alteration should be made in a future PR, as it is beyond the scope of this PR.

src/reducers/index.js
tuliomir: That's correct: this code is actually the expected behavior of a reducer.

Since it's very similar to the other one that was inadequate, it raised this confusion.

src/reducers/index.js
tuliomir: This code was refactored to the wallet saga with a more appropriate explanation on https://github.com/HathorNetwork/hathor-wallet/pull/587/commits/6b87012bfa353140c7a864a832f81fdc097a799d

andreabadesso (approved):

src/reducers/index.js
tuliomir: Opened issue #588 to track reducers with side effects such as this one.

PR #580, opened 2024-04-09
Refactor: Removes `HathorWallet` from Redux store

tuliomir:
The next version of Redux throws an error by default whenever there are non-serializable objects in its Store ( more about this on [this article](https://www.bam.tech/article/the-redux-best-practice-do-not-put-non-serializable-values-in-state-or-actions-explained) ). This PR is one in a series of PRs that aim to refactor all of those instances before the upgrade.

The `wallet` Redux property stored is currently used for many operations throughout the whole application. The proposed solution is to:
- Create a `wallet singleton` javascript module, apart from the Redux store
- Refactor all the calls from this store property to use the singleton directly

It is worth noting that we plan on implementing multiple simultaneous wallets in the future ( see #210 ). This refactor also aims to facilitate this feature, while not directly implementing it.

### Acceptance Criteria
- The `routerHistory` should not be present on the Redux Store

### Security Checklist
- [X] Make sure you do not include new dependencies in the project unless strictly necessary and do not include dev-dependencies as production ones. More dependencies increase the possibility of one of them being hijacked and affecting us.

PR #589, opened 2024-04-25
Refactor: Removes `unleash` from Redux store

tuliomir:

The `unleash` Redux property is currently initialized at the start of the application and used as a monitoring tool for dynamic feature activation. The proposed _singleton_ solution is similar to the one on #580 , which was recently merged.

### Acceptance Criteria
- Create an `unleash` singleton javascript module, apart from the Redux store
- Refactor all the calls from this store property to use the singleton directly
- Remove unused code after this change

### Security Checklist

### DIFF
--- a/src/modules/unleash.js
+++ b/src/modules/unleash.js
@@ -1,0 +1,16 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+let unleashClient = null;
+
+export function setUnleashClient(_unleashClient) {
+	unleashClient = _unleashClient;
+}
+
+export function getUnleashClient() {
+	return unleashClient;
+}
--- a/src/reducers/index.js
+++ b/src/reducers/index.js
@@ -132,7 +132,6 @@ const initialState = {
       status: TOKEN_DOWNLOAD_STATUS.READY
     }
   },
-  unleashClient: null,
   featureTogglesInitialized: false,
   featureToggles: {
     ...FEATURE_TOGGLE_DEFAULTS,
@@ -349,12 +348,11 @@ const onUpdateLoadedData = (state, action) => ({
 });
 
 const onCleanData = (state) => {
+  // Keep the unleashClient as it should continue running
   return Object.assign({}, initialState, {
     isVersionAllowed: state.isVersionAllowed,
     loadingAddresses: state.loadingAddresses,
     ledgerWasClosed: state.ledgerWasClosed,
-    // Keep the unleashClient as it should continue running
-    unleashClient: state.unleashClient,
     featureTogglesInitialized: state.featureTogglesInitialized,
   });
 };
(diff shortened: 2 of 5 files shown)


src/reducers/index.js
```diff
@@ -987,20 +983,6 @@ const onSetFeatureToggles = (state, { payload }) => ({
...
-const onSetUnleashClient = (state, { payload }) => ({
-  ...state,
-  unleashClient: payload,
-});
-
-const onWalletResetSuccess = (state) => ({
-  ...state,
-  // Keep the unleashClient as it should continue running
-  unleashClient: state.unleashClient,
-});
```
tuliomir, line 1002: By deleting the `unleashClient`,  this function would only return the `state` unmodified. So, it was completely removed along with its related actions.

src/modules/unleash.js
```diff
@@ -0,0 +1,16 @@
+/**
```
andreabadesso, line 1: Please fix the identing in this file

src/modules/unleash.js
```diff
@@ -0,0 +1,16 @@
...
+
+let unleashClient = null;
+
+export function setUnleashClient(_unleashClient) {
+	unleashClient = _unleashClient;
+}
+
+export function getUnleashClient() {
+	return unleashClient;
+}
```
andreabadesso, line 16: I know we did this in the wallet singleton, but I think we should refactor this singleton (and the wallet singleton) so we don't pollute the global context... Something like:

```javascript
const unleashModule = (function() {
    let unleashClient = null;

    function setUnleashClient(_unleashClient) {
        unleashClient = _unleashClient;
    }

    function getUnleashClient() {
        return unleashClient;
    }

    return {
        setUnleashClient,
        getUnleashClient
    };
})();

export default unleashModule;
```

What do you think?

src/reducers/index.js
```diff
@@ -349,12 +346,11 @@ const onUpdateLoadedData = (state, action) => ({
 });
 
 const onCleanData = (state) => {
+  // Keep the unleashClient as it should continue running
```
andreabadesso, line 349: Should this comment be removed?

src/sagas/featureToggle.js
```diff
@@ -39,18 +38,19 @@ import {
   UNLEASH_POLLING_INTERVAL,
   FEATURE_TOGGLE_DEFAULTS,
 } from '../constants';
+import { getUnleashClient, setUnleashClient } from "../modules/unleash";
```
andreabadesso, line 41: Use single quotes

andreabadesso (changes requested):

src/sagas/featureToggle.js
tuliomir: Done on https://github.com/HathorNetwork/hathor-wallet/pull/589/commits/b08a92e96d9e300b35a288f5cdf5a129095651dd

src/modules/unleash.js
tuliomir: Fixed on https://github.com/HathorNetwork/hathor-wallet/pull/589/commits/b08a92e96d9e300b35a288f5cdf5a129095651dd

src/modules/unleash.js
tuliomir: As we've talked, this application already uses ES6 modules through NodeJs 20. The variables and properties do not pollute the global environment when they're declared as they are.

See [Mdn Web Docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#other_differences_between_modules_and_standard_scripts):
> ...module features are imported into the scope of a single script — they aren't available in the global scope. Therefore, you will only be able to access imported features in the script they are imported into...

Considering this, I believe it's better to keep the code as it is for legibility.
Do you agree?

andreabadesso (approved):

alexruzenhack (approved):

PR #594, opened 2024-04-29
Fix: Missing navigation calls

tuliomir:
Two instances of navigation calls were missed during the refactor of #555 . Those should be corrected so that all navigation calls work accordingly.

### Acceptance Criteria
- All navigation requests within the application should use the `useNavigate` hook

### Notes
Since the `ModalUnhandledError` component is quite simple, I took the opportunity to also refactor it to a functional component. The `TxData` component, however, is complex enough to deserve its own PR in a future moment, so I only made the smallest adjustment to fix the issue.

### Security Checklist

### DIFF
--- a/src/components/ModalUnhandledError.js
+++ b/src/components/ModalUnhandledError.js
@@ -5,7 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React from 'react';
+import React, { useState } from 'react';
 import { t } from 'ttag';
 import walletUtils from '../utils/wallet';
 import { CopyToClipboard } from 'react-copy-to-clipboard';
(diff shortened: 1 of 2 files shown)


src/components/ModalUnhandledError.js
```diff
@@ -7,106 +7,103 @@
...
  */
-class ModalUnhandledError extends React.Component {
+function ModalUnhandledError({ resetError, error, renderError, info }) {
+  const navigate = useNavigate();
+  const dispatch = useDispatch();
   /**
    * successMessage {string} Message to be shown to the user in the modal
    */
-  state = { successMessage: '' };
+  const [successMessage, setSuccessMessage] = React.useState('');
```
alexruzenhack, line 31: [suggestion] Why not import directly the function `useState` from read?

alexruzenhack (approved):
Approved with a suggestion.

PR #608, opened 2024-05-08
refactor: Remove seed words from Redux

tuliomir:
This PR removes the Seed Words from Redux, and instead keeps it in the smallest scope possible within the component state of the `NewWallet` and `ModalBackupWords` screens.

This avoids any future bugs where the Redux store cleanup fails or even security issues should the Redux store happen to leak.

### Acceptance Criteria
- The seed words should never be stored on Redux

### Notes
The translations file also had to be updated, as there were changes to the screen component files.

### Security Checklist

### DIFF
--- a/src/components/ModalBackupWords.js
+++ b/src/components/ModalBackupWords.js
@@ -10,25 +10,9 @@ import { t } from 'ttag';
 
 import $ from 'jquery';
 import _ from 'lodash';
-import { updateWords } from '../actions/index';
-import { connect } from 'react-redux';
 import hathorLib from '@hathor/wallet-lib';
 import { WORDS_VALIDATION } from '../constants';
 import { getGlobalWallet } from '../modules/wallet';
-
-
-const mapDispatchToProps = dispatch => {
-  return {
-    updateWords: data => dispatch(updateWords(data)),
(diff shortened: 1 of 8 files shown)


src/components/ModalBackupWords.js
```diff
@@ -111,14 +95,15 @@ class ModalBackupWords extends React.Component {
         const { storage } = getGlobalWallet();
         const accessData = await storage.getAccessData();
         const words = hathorLib.cryptoUtils.decryptData(accessData.words, password);
-        this.props.updateWords(words);
+        this.state.words = words;
```
alexruzenhack, line 98: [not blocking] Why not use the `this.setState` here? I feel the it would be better.

tuliomir: You actually caught a bug waiting to happen: I was updating the state incorrectly and it could not trigger the re-rendering of the component.

Fixed on https://github.com/HathorNetwork/hathor-wallet/pull/608/commits/afb58ccc5cc1ed5358377df85ff0be2cb267e2da.

alexruzenhack (approved):

r4mmer (approved):

PR #623, opened 2024-06-03
feat: show nano contract detail

pedroferreira1:
### Acceptance Criteria
- Add Nano Contract Detail screen with attributes and available public methods to execute.
- Add Modal to change nano contract registered address.
- Refactor All Addresses screen to share a unique component with change address modal.

https://github.com/HathorNetwork/hathor-wallet/assets/3298774/a50997a6-e1be-437f-96bc-949d2fdccb29

### Security Checklist

### DIFF
--- a/src/components/AddressList.js
+++ b/src/components/AddressList.js
@@ -20,7 +20,7 @@ import { getGlobalWallet } from '../modules/wallet';
  *
  * @memberof Screens
  */
-function AddressList(props) {
+function AddressList({ count, onAddressClick, showNumberOfTransaction }) {
   const wallet = getGlobalWallet();
 
   const alertErrorRef = useRef(null);
@@ -45,7 +45,7 @@ function AddressList(props) {
    * @return {Number} Total number of pages of the list
    */
   const getTotalPages = (array) => {
-    return Math.ceil(array.length / props.count);
+    return Math.ceil(array.length / count);
   }
 
   useEffect(() => {
@@ -139,7 +139,7 @@ function AddressList(props) {
    */
   const goToAddressSearch = (e, address) => {
     e.preventDefault();
-    props.onAddressClick(address);
+    onAddressClick(address);
   }
 
   const loadPagination = () => {
@@ -163,14 +163,14 @@ function AddressList(props) {
   }
 
   const renderData = () => {
-    const startIndex = (page - 1) * props.count;
-    const endIndex = startIndex + props.count;
+    const startIndex = (page - 1) * count;
+    const endIndex = startIndex + count;
     return filteredAddresses.slice(startIndex, endIndex).map((addressObj) => {
       return (
         <tr key={addressObj.address}>
           <td><a href="true" onClick={(e) => goToAddressSearch(e, addressObj.address)}>{addressObj.address}</a></td>
           <td>{addressObj.index}</td>
-          {props.showNumberOfTransaction && <td className="number">{addressObj.transactions}</td>}
+          {showNumberOfTransaction && <td className="number">{addressObj.transactions}</td>}
         </tr>
       )
     });
@@ -187,7 +187,7 @@ function AddressList(props) {
             <tr>
               <th>{t`Address`}</th>
               <th>{t`Index`}</th>
-              {props.showNumberOfTransaction && <th className="number">{t`Number of transactions`}</th>}
+              {showNumberOfTransaction && <th className="number">{t`Number of transactions`}</th>}
             </tr>
           </thead>
           <tbody>
--- a/src/components/AddressList.js
+++ b/src/components/AddressList.js
@@ -13,6 +13,7 @@ import { WALLET_HISTORY_COUNT } from '../constants';
 import walletUtils from '../utils/wallet';
 import { useSelector } from 'react-redux';
 import PropTypes from 'prop-types';
+import { getGlobalWallet } from '../modules/wallet';
 
 /**
  * Screen that has a list of addresses of the wallet
@@ -20,7 +21,7 @@ import PropTypes from 'prop-types';
  * @memberof Screens
  */
 function AddressList(props) {
-  const wallet = useSelector(state => state.wallet);
+  const wallet = getGlobalWallet();
 
   const alertErrorRef = useRef(null);
   const txSearchRef = useRef(null);
--- a/src/components/AddressList.js
+++ b/src/components/AddressList.js
@@ -10,18 +10,17 @@ import { t } from 'ttag'
 import HathorPaginate from '../components/HathorPaginate';
 import HathorAlert from '../components/HathorAlert';
 import { WALLET_HISTORY_COUNT } from '../constants';
-import helpers from '../utils/helpers';
 import walletUtils from '../utils/wallet';
-import path from 'path';
-import { getGlobalWallet } from '../modules/wallet';
+import { useSelector } from 'react-redux';
+import PropTypes from 'prop-types';
 
 /**
  * Screen that has a list of addresses of the wallet
  *
  * @memberof Screens
  */
-function AddressList() {
-  const wallet = getGlobalWallet();
+function AddressList(props) {
+  const wallet = useSelector(state => state.wallet);
 
   const alertErrorRef = useRef(null);
   const txSearchRef = useRef(null);
@@ -45,7 +44,7 @@ function AddressList() {
    * @return {Number} Total number of pages of the list
    */
   const getTotalPages = (array) => {
-    return Math.ceil(array.length / WALLET_HISTORY_COUNT);
+    return Math.ceil(array.length / props.count);
   }
 
   useEffect(() => {
@@ -139,8 +138,7 @@ function AddressList() {
    */
   const goToAddressSearch = (e, address) => {
     e.preventDefault();
-    const url = path.join(helpers.getExplorerURL(), `address/${address}`);
-    helpers.openExternalURL(url);
+    props.onAddressClick(address);
   }
 
   const loadPagination = () => {
@@ -164,45 +162,53 @@ function AddressList() {
   }
 
   const renderData = () => {
-    const startIndex = (page - 1) * WALLET_HISTORY_COUNT;
-    const endIndex = startIndex + WALLET_HISTORY_COUNT;
+    const startIndex = (page - 1) * props.count;
+    const endIndex = startIndex + props.count;
     return filteredAddresses.slice(startIndex, endIndex).map((addressObj) => {
       return (
         <tr key={addressObj.address}>
           <td><a href="true" onClick={(e) => goToAddressSearch(e, addressObj.address)}>{addressObj.address}</a></td>
           <td>{addressObj.index}</td>
-          <td className="number">{addressObj.transactions}</td>
+          {props.showNumberOfTransaction && <td className="number">{addressObj.transactions}</td>}
         </tr>
       )
     });
   }
 
   return (
-    <div className="content-wrapper">
-      <div className="d-flex flex-column">
-        <div className="d-flex flex-row justify-content-between">
-          <h2>{t`Addresses`}</h2>
-          {renderSearch()}
-        </div>
-        <div className="table-responsive">
-          <table className="mt-3 table table-striped" id="address-list">
-            <thead>
-              <tr>
-                <th>{t`Address`}</th>
-                <th>{t`Index`}</th>
-                <th className="number">{t`Number of transactions`}</th>
-              </tr>
-            </thead>
-            <tbody>
-              {renderData()}
-            </tbody>
-          </table>
-        </div>
-        {loadPagination()}
+    <div className="d-flex flex-column">
+      <div className="d-flex flex-row justify-content-end">
+        {renderSearch()}
       </div>
+      <div className="table-responsive">
+        <table className="mt-3 table table-striped" id="address-list">
+          <thead>
+            <tr>
+              <th>{t`Address`}</th>
+              <th>{t`Index`}</th>
+              {props.showNumberOfTransaction && <th className="number">{t`Number of transactions`}</th>}
+            </tr>
+          </thead>
+          <tbody>
+            {renderData()}
+          </tbody>
+        </table>
+      </div>
+      {loadPagination()}
       <HathorAlert ref={alertErrorRef} text="Invalid address" type="danger" />
     </div>
-  );
+  )
+}
+
+/*
+ * showNumberOfTransaction: If should show the column with number of transactions
+ * onAddressClick: Method called when user clicks on the address
+ * count: Quantity of elements in the list
+ */
+AddressList.propTypes = {
+  showNumberOfTransaction: PropTypes.bool.isRequired,
+  onAddressClick: PropTypes.func.isRequired,
+  count: PropTypes.number.isRequired,
 };
 
 export default AddressList;
--- a/src/components/nano/ModalChangeAddress.js
+++ b/src/components/nano/ModalChangeAddress.js
@@ -79,9 +79,10 @@ function ModalChangeAddress(props) {
    *
    * @param {Object} e Event emitted on click
    */
-  const executeChange = (e) => {
+  const executeChange = async (e) => {
     e.preventDefault();
     dispatch(editAddressNC(props.nanoContractID, newAddress.address));
+    await wallet.storage.updateNanoContractRegisteredAddress(props.nanoContractID, newAddress.address);
     $('#changeAddressModal').modal('hide');
   }
 
--- a/src/components/nano/ModalChangeAddress.js
+++ b/src/components/nano/ModalChangeAddress.js
@@ -1,0 +1,166 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React, { useEffect, useRef, useState } from 'react';
+import { t } from 'ttag';
+import $ from 'jquery';
+import PropTypes from 'prop-types';
+import AddressList from '../../components/AddressList';
+import { CHANGE_ADDRESS_LIST_COUNT } from '../../constants';
+import { connect } from 'react-redux';
+import { useDispatch, useSelector } from 'react-redux';
+import { editAddressNC } from '../../actions/index';
+import { getGlobalWallet } from '../../modules/wallet';
+
+/**
+ * Component that shows a modal to change the address of a Nano Contract
+ *
+ * @memberof Components
+ */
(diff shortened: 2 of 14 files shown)


src/screens/nano/NanoContractDetail.js
```diff
@@ -0,0 +1,203 @@
...
+    setData(null);
+    try {
+      const state = await hathorLib.ncApi.getNanoContractState(ncId, Object.keys(blueprintInformationAux.attributes), [], []);
+      setLoading(false);
+      setData(state);
+    } catch(e) {
+      // Error in request
+      setLoading(false);
+      setErrorMessage(t`Error getting nano contract state.`);
+    };
```
r4mmer, line 124:
```suggestion
    try {
      const state = await hathorLib.ncApi.getNanoContractState(ncId, Object.keys(blueprintInformationAux.attributes), [], []);
      setData(state);
    } catch(e) {
      // Error in request
      setErrorMessage(t`Error getting nano contract state.`);
    } finally {
      setLoading(false);
    };
```

src/screens/nano/NanoContractDetail.js
```diff
@@ -0,0 +1,203 @@
...
+      blueprintInformationAux = blueprintInformationResponse;
+      setBlueprintInformation(blueprintInformationResponse);
+      // Store in redux, so it can be reused by other nano contracts
+      dispatch(addBlueprintInformation(blueprintInformationResponse));
+    } catch(e) {
+      // Error in request
+      setLoading(false);
+      setErrorMessage(t`Error getting blueprint details.`);
+    };
+  }
```
r4mmer, line 111: If an error is not thrown, who sets the loading to false?
Is this working because loadNCData is called right after?

If so, why not move the `setLoading` pair to the `loadData` method?

src/screens/nano/NanoContractDetail.js
```diff
@@ -0,0 +1,203 @@
...
+        const type = get(blueprintInformation.attributes, field);
+        return <p key={field}><strong>{field}: </strong>{type}</p>;
+      }
+
+      // Some fields should be better parsed, e.g., timestamp, address
+      // however we don't have a simple and generic way to knowing it
+      // this was discussed and we will have this in the future, so
+      // for now we keep this UI and when we have this feature in the
+      // hathor-core, we can improve the user UI
+      return <p key={field}><strong>{field}: </strong>{value === null ? ' - ' : value}</p>;
```
r4mmer, line 156: This logic is basically `<p key={field}><strong>{field}: </strong>{formattedValue}</p>;` where `formattedValue` is the type when value is undefined.

Why not move this to another method `formatNCField` for instance and move the observations on formatting to this helper?

 ```html
<p key={field}><strong>{field}: </strong>{formatNCField(blueprintInformation, field, value)}</p>
```

File: src/screens/nano/NanoContractDetail.js [hunk 4]
```diff
@@ -0,0 +1,203 @@
...
+    });
+  }
+
+  const renderNanoMethods = () => {
+    return Object.keys(blueprintInformation.public_methods).filter((method) =>
```
r4mmer, line 161: Use lodash get with a default since `public_methods` can be null or undefined, making this throw.

src/components/nano/ModalChangeAddress.js
```diff
@@ -0,0 +1,171 @@
...
+
+  /**
+   * Called when user clicks to change the address in step 2
+   *
+   * @param {Object} e Event emitted on click
+   */
+  const executeChange = async (e) => {
+    e.preventDefault();
+    dispatch(editAddressNC(props.nanoContractID, newAddress.address));
+    await wallet.storage.updateNanoContractRegisteredAddress(props.nanoContractID, newAddress.address);
```
tuliomir, line 89: question: Shouldn't the `editAddressNC` action also update the wallet storage? It seems one line execution wouldn't make much sense without the other.

It looks to me like the kind of work to be done by a saga.

src/components/nano/ModalChangeAddress.js
```diff
@@ -0,0 +1,171 @@
...
+        return null;
+    }
+  }
+
+  const renderWarning = () => {
+    return (
+      <div>
+        <p>{t`You can have only one address associated with a nano contract at a time. If you decide to change your address, any action that you execute in this nano contract will be associated with this new address.`}</p>
+        <p>{t`It's important that you remember which address you are using to execute the nano contract methods.`}</p>
+        <p><strong>{t`Please continue only if you understand your risks and know what you are doing.`}</strong></p>
```
tuliomir, line 111: question(non-blocking): Should we consider adding something like "This could result in the loss of funds" warning, for legal reasons?

src/components/nano/ModalChangeAddress.js
```diff
@@ -0,0 +1,171 @@
...
+import { useDispatch, useSelector } from 'react-redux';
+import { editAddressNC } from '../../actions/index';
+import { getGlobalWallet } from '../../modules/wallet';
+
+/**
+ * Component that shows a modal to change the address of a Nano Contract
+ *
+ * @memberof Components
+ */
+function ModalChangeAddress(props) {
```
tuliomir, line 24: suggestion: Explicit props also on method declaration.

```suggestion
function ModalChangeAddress({ nanoContractId, onClose }) {
```

src/constants.js
```diff
@@ -280,3 +280,9 @@ export const FEATURE_TOGGLE_DEFAULTS = {
 export const colors = {
   purpleHathor: scssColors.purpleHathor,
 }
+
+/**
+ * Number of elements in the list of addresses
+ * when changing the address of a nano contract
+ */
+export const CHANGE_ADDRESS_LIST_COUNT = 5;
```
tuliomir, line 288: suggestion: This name feels confusing to me, especially in the global context as it is. _Change address_ is, until now, a concept related to outputs returning tokens back to the sender after partially consuming an utxo.

What do you think about `NANO_ADDRESS_CHANGE_LIST_COUNT`?

src/screens/AllAddresses.js
```diff
@@ -0,0 +1,43 @@
...
+import helpers from '../utils/helpers';
+import path from 'path';
+import AddressList from '../components/AddressList';
+
+/**
+ * Screen that has a list of addresses of the wallet
+ *
+ * @memberof Screens
+ */
+class AllAddresses extends React.Component {
```
tuliomir, line 20: suggestion(non-blocking): I'd suggest that all new screens and components should already be written in functional component style, to push forward our migration away from class components.

src/screens/nano/NanoContractDetail.js
```diff
@@ -0,0 +1,203 @@
...
+import { get } from 'lodash';
+import { GlobalModalContext, MODAL_TYPES } from '../../components/GlobalModal';
+
+
+/**
+ * Details of a Nano Contract
+ *
+ * @memberof Screens
+ */
+function NanoContractDetail(props) {
```
tuliomir, line 29: chore: `props` is not used here. Removing it completely helps reduce the doubt that some props may have been passed.

```suggestion
function NanoContractDetail() {
```

src/screens/nano/NanoContractDetail.js
```diff
@@ -0,0 +1,203 @@
...
+      blueprintsData: state.blueprintsData,
+      tokenMetadata: state.tokenMetadata
+    }
+  });
+
+  const dispatch = useDispatch();
+  const navigate = useNavigate();
+  const { nc_id: ncId } = useParams();
+  const nc = nanoContracts[ncId];
+  let blueprintInformationAux = blueprintsData[nc.blueprintId];
```
tuliomir, line 44: suggestion: Variables on the higher scope of function components can be misleading about their actual value. If you want predictable values I recommend using states.

For this specific case I'd suggest removing this higher-scope variable and only initializing it right before when it's used, within stricter scopes:
- On the initialization `useEffect()`
- Inside `loadBlueprintInformation()`

src/screens/nano/NanoContractDetail.js
tuliomir: Another alternative would be `Object.keys(blueprintInformation.public_methods || {})`

pedroferreira1: It will never be null/undefined. The API returns an empty object `{}` in case there are no public methods.

src/components/nano/ModalChangeAddress.js
pedroferreira1: But this couldn't result in loss of funds, it's just a matter of seeing the transactions correctly in the expected address.

src/screens/nano/NanoContractDetail.js
pedroferreira1: The reason I need this to be in a variable is because I might get this information from the API response and setting it as a state is async, so I couldn't use in the following API call, then I can't have it inside `loadBlueprintInformation` only. Regarding the `useEffect` scope, it could be possible but the code would be much worse, and I don't see any benefits in doing it, only downsides.

src/screens/nano/NanoContractDetail.js
pedroferreira1: Created a helper method in this file. If I start using it elsewhere I will move to a helper/util file. https://github.com/HathorNetwork/hathor-wallet/pull/623/commits/b3c4a856ffa31e091b7b12213e785de94a669a4e

src/screens/nano/NanoContractDetail.js
pedroferreira1: Much better, thanks! https://github.com/HathorNetwork/hathor-wallet/pull/623/commits/76798852607b4f1d6867f088a306a125049cced1

src/screens/nano/NanoContractDetail.js
pedroferreira1: You are correct regarding `setLoading(false)`.

The method `loadBlueprintInformation` might not be called, in case this blueprint information is already loaded but if it's called, the loading must start in the beginning.

So `setLoading(true)` should exist in both methods but `setLoading(false)` only in the second.

Removed the one from the first method. https://github.com/HathorNetwork/hathor-wallet/pull/623/commits/76798852607b4f1d6867f088a306a125049cced1

src/components/AddressList.js
```diff
@@ -10,17 +10,17 @@ import { t } from 'ttag'
...
+import PropTypes from 'prop-types';
 import { getGlobalWallet } from '../modules/wallet';
 
 /**
  * Screen that has a list of addresses of the wallet
  *
  * @memberof Screens
  */
-function AddressList() {
+function AddressList(props) {
```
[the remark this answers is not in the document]
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet/pull/623/commits/fedee007c98568ba73128512a60a638471711e38

src/components/nano/ModalChangeAddress.js
```diff
@@ -0,0 +1,171 @@
...
+
+  // {number} step The modal has 3 steps: 0 is the warning screen, 1 is the address list and 2 is the confirmation
+  const [step, setStep] = useState(0);
+  // {Object} newAddress New nano contract selected address data {'address': string, 'index': number}
+  const [newAddress, setNewAddress] = useState(null);
+  // {number} oldAddressIndex the index of the old selected address. Need to be as a state because of the async/await call to the wallet method
+  const [oldAddressIndex, setOldAddressIndex] = useState(null);
+
+  useEffect(() => {
+    $('#changeAddressModal').modal('show');
```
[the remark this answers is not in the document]
pedroferreira1: Done, thanks! https://github.com/HathorNetwork/hathor-wallet/pull/623/commits/bf9a293d902c1c9062f3871652e358aa8797ab48

src/components/nano/ModalChangeAddress.js
```diff
@@ -0,0 +1,171 @@
...
+      const oldAddressIndex = await wallet.getAddressIndex(oldAddress);
+      setOldAddressIndex(oldAddressIndex);
+    }
+
+    getAndSetOldAddressIndex();
+
+    return () => {
+      $('#changeAddressModal').modal('hide');
+      // Removing all event listeners
+      $('#changeAddressModal').off();
```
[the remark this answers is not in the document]
pedroferreira1: Done, thanks! https://github.com/HathorNetwork/hathor-wallet/pull/623/commits/bf9a293d902c1c9062f3871652e358aa8797ab48

src/components/nano/ModalChangeAddress.js
pedroferreira1: Done, thanks! https://github.com/HathorNetwork/hathor-wallet/pull/623/commits/bf9a293d902c1c9062f3871652e358aa8797ab48

src/components/nano/ModalChangeAddress.js
pedroferreira1: I don't think it makes sense to have storage update in the redux action but moving to a saga makes sense. Done https://github.com/HathorNetwork/hathor-wallet/pull/623/commits/a68294c5e7e2610ea398580b503812eef17a9590

src/constants.js
pedroferreira1: Removed the `CHANGE` from the variable name. Changed to `NANO_UPDATE_ADDRESS_LIST_COUNT`. What do you think? https://github.com/HathorNetwork/hathor-wallet/pull/623/commits/3a28ae1d2ab7d56576bd140954264140c367f0f7

src/screens/AllAddresses.js
pedroferreira1: This actually should be blocking. Refactored to a functional component. https://github.com/HathorNetwork/hathor-wallet/pull/623/commits/8b18b331c1d863c570559aadf9fbcb1efad03eac

src/screens/nano/NanoContractDetail.js
pedroferreira1: Makes sense, thanks https://github.com/HathorNetwork/hathor-wallet/pull/623/commits/a2e88820abeaa9cc925d036f2cea003a220c13b0

src/reducers/index.js
```diff
@@ -1234,4 +1236,34 @@ export const onBlueprintAddInformation = (state, { payload }) => {
...
 
+/**
+ * @param {Object} state
+ * @param {{
+ *   payload: {
+ *     ncId: string,
+ *     address: string,
+ *   }
+  * }} action
+ */
```
r4mmer, line 1247:
```suggestion
/**
 * @param {Object} state
 * @param {Object} action
 * @param {string} action.payload.ncId
 * @param {string} action.payload.address
 */
```

Also, we should have an explanation of the effects of this reducer method.

src/sagas/nanoContract.js
```diff
@@ -107,9 +107,24 @@ export function* registerNanoContract({ payload }) {
...
 }
 
+/**
+ * Updates nano contract registered address in the wallet storage
+ * @param {{
+ *   payload: {
+ *     address: string,
+ *     ncId: string,
+ *   }
+ * }} action with request payload.
```
r4mmer, line 117:
```suggestion
 * @param {Object} state
 * @param {Object} action with request payload.
 * @param {string} action.payload.ncId
 * @param {string} action.payload.address
```

src/screens/nano/NanoContractDetail.js
```diff
@@ -0,0 +1,205 @@
...
+    try {
+      const state = await hathorLib.ncApi.getNanoContractState(ncId, Object.keys(blueprintInformationAux.attributes), [], []);
+      setData(state);
+    } catch(e) {
+      // Error in request
+      setErrorMessage(t`Error getting nano contract state.`);
+    } finally {
+      setLoading(false);
+    };
+  }
```
r4mmer, line 124:
```suggestion
  const loadData = async () => {
    setLoading(true);
    try {
      await loadBlueprintInformation();
      await loadNCData();
    } finally {
      setLoading(false);
    }
  }

  const loadBlueprintInformation = async () => {
    if (blueprintInformationAux) {
      return;
    }

    try {
      const blueprintInformationResponse = await hathorLib.ncApi.getBlueprintInformation(nc.blueprintId);
      // We need this blueprint information response to call the following get state
      // The set state is not sync, so we need to store it in a common variable to be used in the next call
      blueprintInformationAux = blueprintInformationResponse;
      setBlueprintInformation(blueprintInformationResponse);
      // Store in redux, so it can be reused by other nano contracts
      dispatch(addBlueprintInformation(blueprintInformationResponse));
    } catch(e) {
      // Error in request
      setErrorMessage(t`Error getting blueprint details.`);
    };
  }

  const loadNCData = async () => {
    setData(null);
    try {
      const state = await hathorLib.ncApi.getNanoContractState(ncId, Object.keys(blueprintInformationAux.attributes), [], []);
      setData(state);
    } catch(e) {
      // Error in request
      setErrorMessage(t`Error getting nano contract state.`);
    };
  }
```

src/screens/nano/NanoContractDetail.js
r4mmer: If everything works as intended it should not be null/undefined but we should consider this case and either warn the user with an error or quietly discard it with an empty object default.

src/screens/nano/NanoContractDetail.js
tuliomir: Since this is a read-only screen, I believe we can keep this local variable without any greater worries.

However, it's not always intuitive to understand when a functional component will re-render. And, when it does, what consequence this will have on non-`state` and non-`ref` local variables.

As a reference, we've recently had trouble with local variables that were being mutated in a screen on PR #546 and had to choose another approach to fetching and using that local object.

tuliomir (approved):

src/screens/nano/NanoContractDetail.js
pedroferreira1: Makes sense, added the protection https://github.com/HathorNetwork/hathor-wallet/pull/623/commits/b7a84c1b4326caf4b93daa07aeda4422dda2bc2d

src/reducers/index.js
pedroferreira1: Improved the docstring https://github.com/HathorNetwork/hathor-wallet/pull/623/commits/1ef8ca11cd1729981eff0cdc85d04fc7c74de7bc

src/sagas/nanoContract.js
pedroferreira1: Improved the docstring https://github.com/HathorNetwork/hathor-wallet/pull/623/commits/d9e3f1cb964ed8bb849bf018fc3ab075e235a7ad

src/screens/nano/NanoContractDetail.js
pedroferreira1: Much better, thanks https://github.com/HathorNetwork/hathor-wallet/pull/623/commits/9586f5479bc223a3bcc22f4b17237894162b8759

r4mmer (approved):

PR #625, opened 2024-06-03
feat: unregister nano contract

pedroferreira1:
### Acceptance Criteria
- Add modal to unregister nano contract.
- Clean register nano metadata also on success.

https://github.com/HathorNetwork/hathor-wallet/assets/3298774/e7d16b62-7c1e-409c-9276-38ab6891168d

### Security Checklist

### DIFF
--- a/src/components/nano/ModalConfirmUnregister.js
+++ b/src/components/nano/ModalConfirmUnregister.js
@@ -1,0 +1,55 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React, { useEffect } from 'react';
+import { t } from 'ttag';
+import $ from 'jquery';
+
+export default function ModalConfirmUnregisterNanoContract({ onClose, ncId, success }) {
+  useEffect(() => {
+    $('#modalConfirmUnregister').modal('show');
+    $('#modalConfirmUnregister').on('hidden.bs.modal', onClose);
+
+    return () => {
+      $('#modalConfirmUnregister').modal('hide');
+      $('#modalConfirmUnregister').off();
+    };
+  }, []);
+
+  const confirmUnregister = async () => {
+    onClose();
+    await success();
+  }
+
+  return (
+    <div className="modal fade" id="modalConfirmUnregister" tabIndex="-1" role="dialog" aria-labelledby="alertModal" aria-hidden="true" data-backdrop="static" data-keyboard="false">
+      <div className="modal-dialog" role="document">
+        <div className="modal-content">
+          <div className="modal-header">
+            <h5 className="modal-title">
+              { t`Confirm unregister nano contract?` }
+            </h5>
+          </div>
+          <div className="modal-body">
+            <div>
+              <p>{t`Do you confirm that you want to unregister the nano contract with id`}</p>
+              <p>{t`${ncId}?`}</p>
+            </div>
+          </div>
+          <div className="modal-footer">
+            <button onClick={onClose} type="button" className="btn btn-secondary">
+              { t`Cancel` }
+            </button>
+            <button onClick={confirmUnregister} type="button" className="btn btn-hathor">
+              {t`Unregister`}
+            </button>
+          </div>
+        </div>
+      </div>
+    </div>
+  );
+}
--- a/src/components/nano/ModalConfirmUnregister.js
+++ b/src/components/nano/ModalConfirmUnregister.js
@@ -10,23 +10,22 @@ import { t } from 'ttag';
 import $ from 'jquery';
 
 export default function ModalConfirmUnregisterNanoContract({ onClose, ncId, success }) {
-  useEffect(() => {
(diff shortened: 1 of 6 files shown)


src/screens/nano/NanoContractDetail.js
```diff
@@ -53,7 +55,10 @@ function NanoContractDetail(props) {
   const [errorMessage, setErrorMessage] = useState('');
 
   useEffect(() => {
-    loadData();
+    if (nc) {
+      // The nano contract has just been unregistered
+      loadData();
+    }
   }, []);
```
r4mmer, line 62: The nc exists when the contract is unregistered?
Is it not the opposite?

src/components/nano/ModalConfirmUnregister.js
```diff
@@ -0,0 +1,55 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React, { useEffect } from 'react';
+import { t } from 'ttag';
+import $ from 'jquery';
+
+export default function ModalConfirmUnregisterNanoContract({ onClose, ncId, success }) {
+  useEffect(() => {
+    $('#modalConfirmUnregister').modal('show');
```
tuliomir, line 14: suggestion(non-blocking): Extract `#modalConfirmUnregister` to a constant on the file level. Reference from `ModalUnhandledError`:

https://github.com/HathorNetwork/hathor-wallet/blob/a968e6747afb600bb8bf9ed1b83214ef608cbdaa/src/components/ModalUnhandledError.js#L16-L17

src/components/nano/ModalConfirmUnregister.js
```diff
@@ -0,0 +1,55 @@
...
+    $('#modalConfirmUnregister').on('hidden.bs.modal', onClose);
+
+    return () => {
+      $('#modalConfirmUnregister').modal('hide');
+      $('#modalConfirmUnregister').off();
```
tuliomir, line 19: suggestion(non-blocking): If we pass the modal selector to `onClose(id)` this procedure is done by the GlobalModal manager.

tuliomir (approved):

src/screens/nano/NanoContractDetail.js
pedroferreira1: Actually this comment was not good. I had to add this conditional because this was throwing an error when the nano contract had been unregistered, so the `loadData` method was failing.

The best comment would have been `Load data only if nano contract exists in redux, otherwise it has just been unregistered.`

src/reducers/index.js
```diff
@@ -1260,4 +1262,28 @@ export const onNanoContractEditAddress = (state, { payload }) => {
...
+  if (!(payload in state.nanoContracts)) {
+    // This should never happen, we are trying to unregister
+    // a nano contract that is not registered
+    return state;
+  }
+
+  const newNanoContracts = {
+    ...state.nanoContracts,
+  };
+  delete newNanoContracts[payload];
```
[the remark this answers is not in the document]
pedroferreira1: I had no idea, really interesting, thanks!

src/components/nano/ModalConfirmUnregister.js
pedroferreira1: Done, thanks! https://github.com/HathorNetwork/hathor-wallet/pull/625/commits/1eb97caf38dffe6a3feafdf94db3825369f6e550

src/components/nano/ModalConfirmUnregister.js
pedroferreira1: Much better. Done, thanks https://github.com/HathorNetwork/hathor-wallet/pull/625/commits/1eb97caf38dffe6a3feafdf94db3825369f6e550

src/reducers/index.js
```diff
@@ -1266,4 +1268,25 @@ export const onNanoContractEditAddress = (state, { payload }) => {
...
+  * }} action
+ */
+export const onNanoContractUnregister = (state, { payload }) => {
+  if (!(payload in state.nanoContracts)) {
+    // This should never happen, we are trying to unregister
+    // a nano contract that is not registered
+    return state;
+  }
+
+  const {[payload]: _, ...newNanoContracts} = state.nanoContracts;
```
r4mmer, line 1284: I know I made the suggestion of this syntax but I think we should document this more.

```suggestion
/**
 * Remove the given nano contract from the registered nano contracts state.
 *
 * @param {Object} state
 * @param {Object} action
 * @param {string} action.payload
 * @returns {Object}
 */
export const onNanoContractUnregister = (state, { payload }) => {
  if (!(payload in state.nanoContracts)) {
    // This should never happen, we are trying to unregister
    // a nano contract that is not registered
    return state;
  }

  // Create a new object with all keys from state.nanoContracts except `payload`
  const {[payload]: _, ...newNanoContracts} = state.nanoContracts;
```

tuliomir (approved):

src/reducers/index.js
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet/pull/625/commits/4b3f802c4b1a5ad9de5362b499762673924bb097

r4mmer (approved):

PR #626, opened 2024-06-04
feat: Workflow to automatically merge `release` into `master`

tuliomir:
This is the first version of this automation, with a simple PR creation to ensure the merge is always done for each new public version of the application.

### Acceptance Criteria
- Creates a new workflow to automatically merge `release` into `master` whenever a public version is released

#### 1) Add the PR to Hathor's main Project
This requires specific permissions on the workflow GitHub Token, so will be implemented and tested separately

#### 2) Add the Code Owners as reviewers
This requires parsing specific text syntax from [`CODEOWNERS`](https://github.com/HathorNetwork/hathor-wallet/blob/master/.github/CODEOWNERS). So this will be done after the simplest solution is operational.

#### 3) Send a message to Slack informing about the new release
This workflow will be the best place to implement this solution, but will be done at a later moment to allow for a quick release now

### Security Checklist

### DIFF
--- a/.github/workflows/release-merge.yml
+++ b/.github/workflows/release-merge.yml
@@ -1,0 +1,30 @@
+name: Create PR on Release
+
+# Triggered when a new version is released publicly
+on:
+  release:
+    types: [released]
+
+jobs:
+  create-pull-request:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        id: checkout
+        # https://github.com/actions/checkout/releases/tag/v4.1.6
+        uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29
+
+      - name: Create Pull Request
+        id: create_pull
+        env:
+          GH_TOKEN: ${{ secrets.RELEASE_TOKEN }}
+        run: |
+          # Building the PR Data
+          PR_DATA=$(gh pr create \
+            --title "Merge release into master" \
+            --body "Automated PR to merge `release` branch into `master` based on release event." \
+            --base "master" \
+            --head "release" \
+            --assignee "@me")
+          echo "Merge PR url: $PR_DATA"
--- a/.github/workflows/release-merge.yml
+++ b/.github/workflows/release-merge.yml
@@ -22,7 +22,7 @@ jobs:
         run: |
           # Building the PR Data
           PR_DATA=$(gh pr create \
-            --title "Merge release into master" \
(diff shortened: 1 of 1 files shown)


.github/workflows/release-merge.yml
```diff
@@ -0,0 +1,30 @@
...
+        env:
+          GH_TOKEN: ${{ secrets.RELEASE_TOKEN }}
+        run: |
+          # Building the PR Data
+          PR_DATA=$(gh pr create \
+            --title "Merge release into master" \
+            --body "Automated PR to merge `release` branch into `master` based on release event." \
+            --base "master" \
+            --head "release" \
+            --assignee "@me")
```
r4mmer, line 29: The `@me` runs as as the action actor (who triggered the action)?

.github/workflows/release-merge.yml
```diff
@@ -0,0 +1,30 @@
...
+        uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29
+
+      - name: Create Pull Request
+        id: create_pull
+        env:
+          GH_TOKEN: ${{ secrets.RELEASE_TOKEN }}
+        run: |
+          # Building the PR Data
+          PR_DATA=$(gh pr create \
+            --title "Merge release into master" \
```
r4mmer, line 25: In the [github docs](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) we can see that `github.ref` has this observation:

> [...] For workflows triggered by release, this is the release tag created. [...]

And `github.ref_name` is the shorthand tag name, which means we can use:

```suggestion
            --title "[${{ github.ref_name }}] merge release into master" \
```

So that the name of the PR is `[v1.1.1]: merge release into master` or some other title that includes the release version.

.github/workflows/release-merge.yml
tuliomir: That would be the person who manually set the release version as a public release in the GitHub interface ( or command line ). That person will be assigned for also excuting the merge.

.github/workflows/release-merge.yml
tuliomir: That's a great suggestion! Added on https://github.com/HathorNetwork/hathor-wallet/pull/626/commits/af582a2280245bdecc038c7139bf49546a5f824d

r4mmer (approved):

pedroferreira1 (approved):

PR #630, opened 2024-06-17
feat: add action to add data signed by an address

pedroferreira1:
### Acceptance Criteria
- Add link for every signed data argument to open a modal to sign a string with a specific address from the wallet.

https://github.com/user-attachments/assets/3cc17af9-0ec8-408d-9e2c-2922ce775751

### Security Checklist


src/components/nano/ModalSelectAddressToSignData.js
```diff
@@ -0,0 +1,171 @@
...
+      setError(t`Invalid PIN`);
+      return;
+    }
+
+    const data = dataRef.current.value;
+    const oracleData = hathorLib.nanoUtils.getOracleBuffer(address.address, wallet.getNetworkObject());
+    const nanoSerializer = new hathorLib.NanoContractSerializer();
+    const dataSerialized = nanoSerializer.serializeFromType(data, 'str');
+    // TODO getOracleInputData method should be able to receive the PIN as optional parameter as well
+    wallet.pinCode = pin;
```
tuliomir, line 92: chore: We should clear the `pinCode` from the global wallet as soon as it's used by the `getOracleInputData()`. Am I right?

pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet/pull/630/commits/6bfed308d04fd5cc61fe7e05c7e94fcb0b56bf2b

tuliomir (approved):

src/components/AddressList.js
```diff
@@ -18,9 +18,14 @@ import { getGlobalWallet } from '../modules/wallet';
 /**
  * Screen that has a list of addresses of the wallet
  *
+ * @param props.count Quantity of elements to show in each page of the list
+ * @param props.onAddressClick Callback function to call when address is clicked
+ * @param props.showNumberOfTransactions Boolean to validate if should show the number of txs on the list
+ * @param props.isModal Boolean to set if the component will be used in a modal, to set the width of the search
```
r4mmer, line 24:
```suggestion
 * @param {Object} props
 * @param {number} props.count Quantity of elements to show in each page of the list
 * @param {(address: string) => void} props.onAddressClick Callback function to call when address is clicked
 * @param {bool} props.showNumberOfTransactions Boolean to validate if should show the number of txs on the list
 * @param {bool} props.isModal Boolean to set if the component will be used in a modal, to set the width of the search
```

r4mmer (approved):

src/components/AddressList.js
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet/pull/630/commits/d80238085bc2d364505fe78ab858ea14bfa7eec9

PR #632, opened 2024-06-18
Change tx mining server screen

r4mmer:
### Acceptance Criteria
- Add a setting option to change the tx-mining server
- Allow resetting to default value
- Persist change on wallet restart
- Do NOT persist on wallet reset

image

image

mining_server

### Security Checklist

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -53,6 +53,7 @@ export const types = {
   SET_UNLEASH_CLIENT: 'SET_UNLEASH_CLIENT',
   UPDATE_TX_HISTORY: 'UPDATE_TX_HISTORY',
   UPDATE_MINING_SERVER: 'UPDATE_MINING_SERVER',
+  SET_MINING_SERVER: 'SET_MINING_SERVER',
 };
 
 /**
@@ -482,7 +483,22 @@ export const changeWalletState = (newState) => ({
   payload: newState,
 });
 
+/**
+ * Update the tx mining service config on the lib and redux.
+ *
+ * @param {string|undefined} url
+ * @param {boolean} reset
+ */
 export const updateMiningServer = (url, reset) => ({
   type: types.UPDATE_MINING_SERVER,
   payload: { url, reset },
+});
+
+/**
+ * Set the mining server url
+ * @param {string|undefined} url
+ */
+export const setMiningServer = (url) => ({
+  type: types.SET_MINING_SERVER,
+  payload: url,
 });
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -52,6 +52,7 @@ export const types = {
   SET_FEATURE_TOGGLES: 'SET_FEATURE_TOGGLES',
   SET_UNLEASH_CLIENT: 'SET_UNLEASH_CLIENT',
   UPDATE_TX_HISTORY: 'UPDATE_TX_HISTORY',
+  UPDATE_MINING_SERVER: 'UPDATE_MINING_SERVER',
 };
 
 /**
@@ -479,4 +480,9 @@ export const updateTxHistory = (tx, tokenId, balance) => ({
 export const changeWalletState = (newState) => ({
   type: types.WALLET_CHANGE_STATE,
   payload: newState,
+});
+
+export const updateMiningServer = (url, reset) => ({
+  type: types.UPDATE_MINING_SERVER,
+  payload: { url, reset },
 });
--- a/src/constants.js
+++ b/src/constants.js
@@ -87,11 +87,6 @@ export const DEFAULT_SERVERS = [
   'https://node2.mainnet.hathor.network/v1a/',
 ];
 
-export const DEFAULT_MINING_SERVERS = {
-  'mainnet': 'https://txmining.mainnet.hathor.network/',
-  'testnet': 'https://txmining.testnet.hathor.network/',
-}
-
 /**
  * Default server user will connect when none has been chosen
  */
--- a/src/constants.js
+++ b/src/constants.js
@@ -87,6 +87,11 @@ export const DEFAULT_SERVERS = [
   'https://node2.mainnet.hathor.network/v1a/',
 ];
 
+export const DEFAULT_MINING_SERVERS = {
+  'mainnet': 'https://txmining.mainnet.hathor.network/',
+  'testnet': 'https://txmining.testnet.hathor.network/',
+}
+
 /**
  * Default server user will connect when none has been chosen
  */
--- a/src/reducers/index.js
+++ b/src/reducers/index.js
@@ -15,6 +15,7 @@ import { PROPOSAL_DOWNLOAD_STATUS } from '../utils/atomicSwap';
 import { HATHOR_TOKEN_CONFIG } from "@hathor/wallet-lib/lib/constants";
 import helpersUtils from '../utils/helpers';
 import LOCAL_STORE from '../storage';
+import { getGlobalWallet } from "../modules/wallet";
 
 /**
  * @typedef TokenHistory
@@ -130,6 +131,7 @@ const initialState = {
   featureToggles: {
     ...FEATURE_TOGGLE_DEFAULTS,
   },
+  miningServer: null,
 };
 
 const rootReducer = (state = initialState, action) => {
@@ -256,6 +258,8 @@ const rootReducer = (state = initialState, action) => {
       return onUpdateTxHistory(state, action);
     case types.WALLET_CHANGE_STATE:
       return onWalletStateChanged(state, action);
+    case types.UPDATE_MINING_SERVER:
+      return onUpdateMiningServer(state, action);
     default:
       return state;
   }
@@ -1003,5 +1007,24 @@ export const onWalletStateChanged = (state, { payload }) => ({
   ...state,
   walletState: payload,
 });
+
+export const onUpdateMiningServer = (state, { payload }) => {
+  const wallet = getGlobalWallet();
+  if (payload.reset) {
+    wallet.storage.config.setTxMiningUrl();
+    LOCAL_STORE.resetMiningServer();
+    return {
+      ...state,
+      miningServer: null,
+    }
+  } else {
+    wallet.storage.config.setTxMiningUrl(payload.url);
+    LOCAL_STORE.setMiningServer(payload.url);
+    return {
+      ...state,
+      miningServer: payload.url,
+    }
+  }
+}
 
 export default rootReducer;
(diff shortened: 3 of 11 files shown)


src/screens/MiningServer.js
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    setErrorMessage('');
+    newServerRef.current.value = '';
+    pinRef.current.value = '';
+
+    // reset the mining server the wallet config
+    wallet.storage.config.setTxMiningUrl(newServer);
+    setCurrentServer(newServer);
+    LOCAL_STORE.setMiningServer(newServer);
```
andreabadesso, line 71: Side-effects should be in sagas to keep components as pure as possible

andreabadesso (changes requested):

src/constants.js
```diff
@@ -87,6 +87,11 @@ export const DEFAULT_SERVERS = [
   'https://node2.mainnet.hathor.network/v1a/',
 ];
 
+export const DEFAULT_MINING_SERVERS = {
+  'mainnet': 'https://txmining.mainnet.hathor.network/',
+  'testnet': 'https://txmining.testnet.hathor.network/',
+}
+
```
pedroferreira1, line 94: The lib has this information in the `src/config.ts` file. Should we export it from there, so we don't duplicate this information?

src/storage.js
```diff
@@ -569,6 +571,18 @@ export class LocalStorageStore {
   getLedgerAppVersion() {
     return this.getItem(LEDGER_APP_VERSION_KEY);
   }
+
+  setMiningServer(server) {
+    this.setItem(MINING_SERVER_KEY, server);
+  }
+
+  resetMiningServer(server) {
```
pedroferreira1, line 579:
```suggestion
  resetMiningServer() {
```

src/screens/MiningServer.js
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * @memberof Screens
+ */
+function MiningServer() {
+  const wallet = getGlobalWallet();
+
+  /* errorMessage {string} Message to be shown in case of error in form */
+  const [errorMessage, setErrorMessage] = useState('');
+  /* isHardwareWallet {boolean} If the application is using a hardware wallet */
+  const [isHardwareWallet] = useState(LOCAL_STORE.isHardwareWallet());
```
pedroferreira1, line 29: Why does it need to be a state if it can't be changed?

src/screens/MiningServer.js
```diff
@@ -0,0 +1,120 @@
...
+      setErrorMessage(t`New server is not valid`);
+      return;
+    }
+
+    const newServer = newServerRef.current.value;
+
+    // we don't ask for the pin on the hardware wallet
+    if (!isHardwareWallet) {
+      if (!await wallet.checkPin(pinRef.current.value)) {
+        setErrorMessage(`Invalid PIN`);
```
pedroferreira1, line 59:
```suggestion
        setErrorMessage(t`Invalid PIN`);
```

src/screens/MiningServer.js
```diff
@@ -0,0 +1,120 @@
...
+    setCurrentServer(newServer);
+    LOCAL_STORE.setMiningServer(newServer);
+    alertSuccessRef.current.show(3000);
+  }
+
+  const resetSelected = async () => {
+    // we don't ask for the pin on the hardware wallet
+    if (!isHardwareWallet) {
+      if (!await wallet.checkPin(pinRef.current.value)) {
+        setErrorMessage(`Invalid PIN`);
```
pedroferreira1, line 79:
```suggestion
        setErrorMessage(t`Invalid PIN`);
```

src/screens/MiningServer.js
```diff
@@ -0,0 +1,120 @@
...
+
+  /* errorMessage {string} Message to be shown in case of error in form */
+  const [errorMessage, setErrorMessage] = useState('');
+  /* isHardwareWallet {boolean} If the application is using a hardware wallet */
+  const [isHardwareWallet] = useState(LOCAL_STORE.isHardwareWallet());
+  /* currentServer {string} currently connected tx-mining server url */
+  const [currentServer, setCurrentServer] = useState('');
+
+  const currentNetwork = wallet.getNetwork();
+  const defaultServer = DEFAULT_MINING_SERVERS[currentNetwork]
```
pedroferreira1, line 34: You are not using them

src/screens/MiningServer.js
```diff
@@ -0,0 +1,120 @@
...
+
+  useEffect(() => {
+    setCurrentServer(wallet.storage.config.getTxMiningUrl());
```
pedroferreira1, line 37: Why do you need this in the `useEffect`? You can just use `wallet.storage.config.getTxMiningUrl()` as the default state

src/constants.js
r4mmer: I think exporting from the lib is the best approach, but until then I don't see why we can't leave this here.

src/screens/MiningServer.js
r4mmer: This effect does not exist anymore

src/storage.js
r4mmer: Method no longer exists

src/screens/MiningServer.js
r4mmer: This was copied from the `Server.js` screen, you're right, but we may want to change this there as well.

src/screens/MiningServer.js
r4mmer: Done

src/screens/MiningServer.js
r4mmer: Done

src/screens/MiningServer.js
r4mmer: You are correct, removing them

src/screens/MiningServer.js
r4mmer: Done

src/storage.js
pedroferreira1: It's still there

src/screens/MiningServer.js
```diff
@@ -0,0 +1,121 @@
...
+        return;
+      }
+    }
+
+    setErrorMessage('');
+    newServerRef.current.value = '';
+    pinRef.current.value = '';
+
+    setLoading(true);
+    dispatch(updateMiningServer(newServer, reset));
```
pedroferreira1, line 88: This operation doesn't have anything async. Why do you need a loading?

src/screens/MiningServer.js
```diff
@@ -0,0 +1,121 @@
...
+        setErrorMessage(t`New server is not valid`);
+        return;
+      }
+
+      newServer = newServerRef.current.value;
+
+      if (newServer === currentServer) {
+        setErrorMessage(t`New server is the same as the current server`);
+        return;
+      }
```
pedroferreira1, line 72: I think we should at least validate that the string is a valid URL. What do you think?

src/actions/index.js
```diff
@@ -480,3 +481,8 @@ export const changeWalletState = (newState) => ({
   type: types.WALLET_CHANGE_STATE,
   payload: newState,
 });
+
+export const updateMiningServer = (url, reset) => ({
```
andreabadesso, line 485: Missing docstring

src/reducers/index.js
```diff
@@ -15,6 +15,7 @@ import { PROPOSAL_DOWNLOAD_STATUS } from '../utils/atomicSwap';
 import { HATHOR_TOKEN_CONFIG } from "@hathor/wallet-lib/lib/constants";
 import helpersUtils from '../utils/helpers';
 import LOCAL_STORE from '../storage';
+import { getGlobalWallet } from "../modules/wallet";
```
andreabadesso, line 18:
```suggestion
import { getGlobalWallet } from '../modules/wallet';
```

src/reducers/index.js
```diff
@@ -1004,4 +1008,23 @@ export const onWalletStateChanged = (state, { payload }) => ({
   walletState: payload,
 });
 
+export const onUpdateMiningServer = (state, { payload }) => {
+  const wallet = getGlobalWallet();
+  if (payload.reset) {
+    wallet.storage.config.setTxMiningUrl();
+    LOCAL_STORE.resetMiningServer();
```
andreabadesso, line 1015: Why is this on the reducer? I think this is even worse than being on the component... Can you use a saga for side effects?

r4mmer: Created the logic on the saga and a separate event on the reducer

src/reducers/index.js
r4mmer: Removed this line

src/actions/index.js
r4mmer: Added docstring

src/screens/MiningServer.js
r4mmer: While there is no network or io call, the saga is an async call.
Alsoo, in case some error on the saga happens that does not update the current mining server the user will be left with an indicator that the change has not happened.

What do you think?

src/utils/helpers.js
```diff
@@ -69,6 +69,11 @@ const helpers = {
         storage.store.setItem('wallet:wallet_service:ws_server', wsServer);
       }
     }
+
+    let miningServer = LOCAL_STORE.getMiningServer();
+    if (miningServer) {
+      store.dispatch(updateMiningServer(miningSever, false));
```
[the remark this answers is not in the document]
r4mmer: Done

src/screens/MiningServer.js
r4mmer: Done

src/storage.js
r4mmer: Sorry, was looking in the wrong file, fixed

src/screens/MiningServer.js
andreabadesso: If we're adding a loading, I think it should be in the saga (and remove it from the useEffect)

The component should just reflect what is in the redux store

In the case you're describing (an error in the saga), the loading will be there forever, because the `currentServer` will never change thus never causing the `useEffect` to run, right?

src/screens/MiningServer.js
```diff
@@ -0,0 +1,128 @@
...
+  const pinRef = useRef(null);
+  const alertSuccessRef = useRef();
+  const firstRef = useRef(true);
+
+  useEffect(() => {
+    if (firstRef.current) {
+      firstRef.current = false;
+      return;
+    }
+    alertSuccessRef.current.show(3000);
```
pedroferreira1, line 46: Why don't you add this in the `changeServer` method after the `dispatch`? I'm okay continuing with this but I think it would be better.

pedroferreira1 (approved):

src/screens/MiningServer.js
r4mmer: Done

pedroferreira1 (approved):

andreabadesso (approved):

PR #640, opened 2024-07-01
feat: add data output to token transaction

alexruzenhack:
### Acceptance Criteria
- It should add "Add data output" button on Token Transaction screen
- It should add a new tile in the list of tokens representing a data output when "Add data output" is clicked 
- It should have a "Remove" button on each data output tile
- It should remove the data output tile itself when "Remove" is clicked
- It should add each data output to transaction's outputs as a `data` output, converting any input value to a string

##### Initial Screen
Initial Screen

##### Add a Data Output
Transaction with 1 Data Output

##### Add another Data Output

Transaction with 2 Data Output

##### Remove the first Data Output

Transaction without the first Data Output added

##### Example

The following transaction is an example of a token transaction sent with data output.
https://explorer.testnet.hathor.network/transaction/00002a0dabb26966a74f67522db275d7174dfdfc3c151c4d1d788f4daa0b16c9

### Security Checklist

### DIFF
--- a/src/components/OutputData.js
+++ b/src/components/OutputData.js
@@ -1,0 +1,41 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import { t } from 'ttag';
+
+/**
+ * It renders a card containing a input field for the data cotent as string,
+ * and an action to remove the card.
+ *
+ * @param {Object} props
+ * @property {MutableRefObject} dataInputRef Reference for input field value
+ * @property {number} index Data output position in the list
+ * @property {(index: number) => void} remove Callback to remove the data output from the list
+ */
+export const OutputData = ({ dataInputRef, index, remove }) => (
+  <div className='send-tokens-wrapper card'>
+    <div className="outputs-wrapper">
+      <label>{t`Data Output`}</label>
+      <button
+        type="button"
+        className="text-danger remove-token-btn ml-3"
+        onClick={() => remove(index)}
+      >
+        {t`Remove`}
+      </button>
+      <div className="input-group mb-3">
+        <input
+          className="form-control output-address col-5"
+          type="text"
+          ref={dataInputRef}
+          placeholder={t`Data content`}
+        />
+      </div>
+    </div>
+  </div>
+);
--- a/src/components/SendDataOutputOne.js
+++ b/src/components/SendDataOutputOne.js
@@ -9,7 +9,7 @@ import React from 'react';
 import { t } from 'ttag';
 
 /**
- * It renders a card containing a input field for the data cotent as string,
+ * It renders a card containing an input field for the data content as string,
  * and an action to remove the card.
  *
  * @param {Object} props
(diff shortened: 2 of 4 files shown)


src/screens/SendTokens.js
```diff
@@ -182,6 +192,15 @@ function SendTokens() {
...
     }
+    // if there is any data output, add it to transaction's outputs
+    if (dataOutputs.length) {
+      dataOutputs.forEach((uId) => {
+        data.outputs.push({
+          type: OutputType.DATA,
+          data: String(dataOutputRefs.current[uId].current.value),
+        });
+      });
+    }
```
r4mmer, line 203: Just to be clear on how this will work, the outputs array will have `[...tokenOutputs, ...dataOutputs]` meaning that conventionally all data outputs will be at the end of the outputs array.
If there is a change output added later the outputs will be shuffled.

This way works for me but if the intention is to have the outputs always at the end we should revisit how we organize the outputs.

src/screens/SendTokens.js
```diff
@@ -556,9 +603,29 @@ function SendTokens() {
...
+            >
+              {t`Add another token`}
+            </button>
+            <button
+              type="button"
+              className="btn btn-hathor"
+              onClick={onSendTokensClicked}
+            >
+              {t`Send Tokens`}
+            </button>
```
r4mmer, line 628: Since adding data outputs is an advanced feature should we have a checkbox for "Show advanced features" and only show the "Add data output" button when it is clicked?

If not, I think at least that the first option should be "Add another token" since it is the most commonly used, what do you think?

src/components/OutputData.js
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * It renders a card containing a input field for the data cotent as string,
+ * and an action to remove the card.
+ *
+ * @param {Object} props
+ * @property {MutableRefObject} dataInputRef Reference for input field value
+ * @property {number} index Data output position in the list
+ * @property {(index: number) => void} remove Callback to remove the data output from the list
+ */
+export const OutputData = ({ dataInputRef, index, remove }) => (
```
r4mmer, line 20: Maybe we should name the same way we name the component to add a token output.
Something like `SendDataOne` or `SendDataOutputOne` what do you think?

src/screens/SendTokens.js
pedroferreira1: It's ok to shuffle the outputs in this case, no problem.

src/screens/SendTokens.js
pedroferreira1: I think we can have the button visible but with the add another token as the first option.

alexruzenhack: I've changed the order.

src/components/OutputData.js
alexruzenhack: Renamed to `SendDataOutputOne`...

r4mmer (approved):

src/components/SendDataOutputOne.js
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import { t } from 'ttag';
+
+/**
+ * It renders a card containing a input field for the data cotent as string,
```
pedroferreira1, line 12:
```suggestion
 * It renders a card containing an input field for the data content as string,
```

pedroferreira1 (approved):

src/components/SendDataOutputOne.js
alexruzenhack: Fixed at https://github.com/HathorNetwork/hathor-wallet/pull/640/commits/f3dda2a86dfef0a8a9f8aece7e5db57982ae9bf8

PR #672, opened 2024-10-16
fix: wallet for Ubuntu 24.04 with backward compatibility

alexruzenhack:
### Acceptance Criteria
- Add appropriate AppArmor profile
- Add script to install the profile on app installation
- Add script to remove the profile on app removal
- It should be backward compatible with previous ubuntu versions

Closes https://github.com/HathorNetwork/internal-issues/issues/392

>[!WARNING]
>The AppImage target is **not** addressed in this fix because it doesn't support after install script execution, which is required for the fix.

### Security Checklist


build_resources/linux/after-install.tpl
```diff
@@ -0,0 +1,66 @@
...
+    cp -f "$APPARMOR_PROFILE_SOURCE" "$APPARMOR_PROFILE_TARGET"
+
+    if hash apparmor_parser 2>/dev/null; then
+      # Extra flags taken from dh_apparmor:
+      # > By using '-W -T' we ensure that any abstraction updates are also pulled in.
+      # https://wiki.debian.org/AppArmor/Contribute/FirstTimeProfileImport
+      apparmor_parser --replace --write-cache --skip-read-cache "$APPARMOR_PROFILE_TARGET"
+    fi
+  else
+    echo "Skipping the installation of the AppArmor profile as this version of AppArmor does not seem to support the profile bundled with Hathor Wallet."
```
luislhl, line 64: I believe there is a risk that future versions of Ubuntu and/or AppArmor could introduce changes that will make the current profile fail this test.

In this case, the application would probably break again.

I don't know how likely this is. Maybe it's unlikely in the short term, but could become more likely as time goes by.

But I think we should probably add in some release or QA step that we want to test the Ubuntu packages before every release to avoid this scenario.

What do you think?

build_resources/linux/after-install.tpl
```diff
@@ -0,0 +1,66 @@
...
+if hash update-mime-database 2>/dev/null; then
+  update-mime-database /usr/share/mime || true
+fi
+if hash update-desktop-database 2>/dev/null; then
+  update-desktop-database /usr/share/applications || true
+fi
+
+
+###
+# Custom after-install.tpl script.
```
luislhl, line 23: Isn't there a way to avoid overwriting the Electron's project `after-install` script, just add an additional script to be run after the default?

If they make any change there in the next releases, we would miss it.

build_resources/linux/after-install.tpl
alexruzenhack: Sure, I believe this is the best approach to decouple from memory.

build_resources/linux/after-install.tpl
alexruzenhack: I don't agree with it because this custom script wraps up default one doing some polish work over it. Maybe we should have a review process? Idk, maybe in the release process? It could be also a QA step? What do you think?

luislhl: Ok, I hadn't noticed you made changes over the default script from Electron as well.

I guess when we next upgrade Electron would be the time to check this, but I'm not sure where we should document this.

luislhl (approved):

package.json
```diff
@@ -125,7 +125,20 @@
...
+          "to": "./apparmor-profile"
+        }
       ]
+    },
+    "deb": {
+      "afterInstall": "build_resources/linux/after-install.tpl",
+      "afterRemove": "build_resources/linux/after-remove.tpl"
+    },
+    "directories": {
+      "buildResources": "build_resources"
```
pedroferreira1, line 141: Why do you need this?

build_resources/linux/apparmor-profile
```diff
@@ -0,0 +1,9 @@
+abi <abi/4.0>,
+include <tunables/global>
+
+profile hathor-wallet /opt/Hathor\ Wallet/hathor-wallet flags=(unconfined) {
```
pedroferreira1, line 4: Why are you using `unconfined` flag? Did you try other flags?

As far as I understood, this flag disables apparmor protection when using our application, if I'm not wrong. If there are other flags that don't disable completely but limits the application in a way that the wallet works fine, it would be better.

Did you see any other possibility?

alexruzenhack: I didn't apply any other flag for test. I don't know any other possibility, to be honest. This whole file is a template provided by the Ubuntu team to work around the apparmor limitation. In general lines this profile is only identifying the app as "hathor-wallet" and because it is not an anonymous app it can make use of `userns`.

package.json
alexruzenhack: This is relevante to the build process, but I can't tell why with precision.

build_resources/linux/apparmor-profile
alexruzenhack: Changed to `default_allow`, it is the must correct flag value. For more details refer to: https://manpages.ubuntu.com/manpages/noble/man5/apparmor.d.5.html

build_resources/linux/apparmor-profile
```diff
@@ -0,0 +1,9 @@
+abi <abi/4.0>,
+include <tunables/global>
+
+profile hathor-wallet /opt/Hathor\ Wallet/hathor-wallet flags=(default_allow) {
```
alexruzenhack, line 4: Changed to default_allow, it is the must correct flag value. For more details refer to: https://manpages.ubuntu.com/manpages/noble/man5/apparmor.d.5.html

luislhl (approved):

pedroferreira1 (approved):

PR #673, opened 2024-10-17
chore: add build for linux arm64

alexruzenhack:
### Acceptance Criteria

- Add linux arm64 build

### Security Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,6 +75,7 @@
     "electron-pack": "electron-builder --mac --win --linux -c.extraMetadata.main=build/electron.js",
     "electron-pack-mac": "electron-builder --mac -c.extraMetadata.main=build/electron.js",
     "electron-pack-linux": "electron-builder --linux -c.extraMetadata.main=build/electron.js",
+    "electron-pack-linux-arm64": "electron-builder --linux --arm64 -c.extraMetadata.main=build/electron.js",
     "electron-pack-win": "electron-builder --win -c.extraMetadata.main=build/electron.js",
     "watch-electron": "ELECTRON_START_URL=http://localhost:3000 NODE_ENV=dev nodemon --watch ./public/**/* --watch . --exec 'npm run electron'",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -75,6 +75,7 @@
     "electron-pack": "electron-builder --mac --win --linux -c.extraMetadata.main=build/electron.js",
     "electron-pack-mac": "electron-builder --mac -c.extraMetadata.main=build/electron.js",
     "electron-pack-linux": "electron-builder --linux -c.extraMetadata.main=build/electron.js",
+    "electron-pack-linux-arm64": "electron-builder --linux --arm64 -c.extraMetadata.main=build/electron.js",
```
luislhl, line 78: I think you should add this to `release.sh` as well, if this is to be run in all our builds

alexruzenhack: Added

luislhl: Sorry, going back to this now, I'm wondering whether we should really add this in the script that builds the others.

I wasn't sure at the time.

And I think I'm not the best person to decide this, actually. Maybe the people more involved in the hathor-wallet build and release process are better than me to give an opinion here.

I think that if we add it to the script, it means we expect it to be past of the release process, be included in the release assets, etc. And I don't know whether we will have to adjust our QA to test in ARM Linux, and whether we want to offer support for users trying to run there (although I think there won't be many)

In short, I think I'm not a suitable reviewer for this one.

alexruzenhack: @pedroferreira1, what do you think about it? I'm adding the `arm64` pack in the build but I don't think we should give support to it, just have it in the build. Are you ok with that? If so we should make clear in the release note that we don't provide support for `arm64` and that the package is present to make testers life easier.

pedroferreira1: I agree to add the command in the `package.json` file but we can remove the code from the `release.sh`. If we feel the need for that, we can add it there, but for now we will use the command only for testing.

pedroferreira1 (approved):

luislhl (approved):

PR #684, opened 2024-11-07
ci: add support for node v22

glevco:
### Acceptance Criteria
- Include here all things that this PR should solve

### Security Checklist


pedroferreira1 (approved):

tuliomir (approved):

tuliomir (commented):
Since we are requiring use of Node 22 on the Markdown documents, we should also update the `package.json` with this information:

https://github.com/HathorNetwork/hathor-wallet/blob/813d6fa4f00e5d851bf3b734467dd4415bb0e8c0/package.json#L26-L29

glevco:
> Since we are requiring use of Node 22 on the Markdown documents, we should also update the package.json

As agreed via Slack, we'll keep v20 in the package.json until we perform QA on v22

PR #690, opened 2024-11-12
fix: allTokens variable is a list, not an object

pedroferreira1:
### Acceptance Criteria
- Correctly load custom token balances when starting the wallet

This bug was introduced in this refactor: https://github.com/HathorNetwork/hathor-wallet/pull/587

### Security Checklist

### DIFF
--- a/src/sagas/wallet.js
+++ b/src/sagas/wallet.js
@@ -708,7 +708,7 @@ export function* walletReloading() {
 
     // We might have lost transactions during the reload, so we must invalidate the
     // token histories:
-    for (const tokenUid of Object.keys(allTokens)) {
+    for (const tokenUid of allTokens) {
       if (tokenUid === hathorLibConstants.NATIVE_TOKEN_UID) {
         continue;
       }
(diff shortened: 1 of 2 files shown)


andreabadesso (approved):

tuliomir (approved):

src/sagas/wallet.js
```diff
@@ -708,7 +708,7 @@ export function* walletReloading() {
 
     // We might have lost transactions during the reload, so we must invalidate the
     // token histories:
-    for (const tokenUid of Object.keys(allTokens)) {
+    for (const tokenUid of allTokens) {
```
pedroferreira1, line 711: This `allTokens` comes from `loadTokens` method, so it's also a list.

src/utils/wallet.js
```diff
@@ -215,10 +215,10 @@ const wallet = {
   },
 
   /**
-   * Filters only the non-registered tokens from the allTokens list.
+   * Filters only the non-registered tokens from the allTokens object.
    * Optionally filters only those with non-zero balance.
    *
-   * @param {Object[]} allTokens list of all available tokens
+   * @param {Record<string, string>} allTokens object with tokenUid as key and value
```
pedroferreira1, line 221: This value comes from the redux state, so it's an object

tuliomir (approved):

andreabadesso (approved):

Issue #696, opened 2024-11-27
Restore automated tests


tuliomir:
Automated tests were implemented through #357 and #365 , but disabled on #416 .

We should reactivate those tests to improve our QA process, both unit and e2e. It would be nice to have visual regression tests and simulated ledger interactions as well.

# Found issues
The unit tests are not able to run because Jest cannot process the modern code we're using. This happens because we're using an older version of jest that is defined by CRA, and we have to decide an upgrade path ( mapped in #441 ) before we can upgrade jest.

tuliomir:
This task was partially implemented by #698, #736, and will be handled from now on by #704

PR #698, opened 2024-11-27
fix: Restore e2e tests

tuliomir:
### Acceptance Criteria
- Restores e2e tests
- Adds coverage to:
  - Starting a new wallet with no backup
  - Backing up words
  - Successfully fetching tx history from fullnode
  - Handling fullnode failure on fetch tx history
- Partially solves #696 

### Notes
These initial tests serve as a proof of concept for end to end testing on the Desktop Wallet, allowing us to aim at having the full QA executed automatically. It will be open to analysis if we should run the full QA at each CI iteration, or what part of the coverage should be run at developer machines.

The next PRs should include relevant plugins like the ones that execute visual inspection on the screens, but are outside the scope of this basic test restoration.

### Security Checklist


src/components/ModalBackupWords.js
```diff
@@ -244,6 +244,9 @@ class ModalBackupWords extends React.Component {
               {renderWords(6)}
             </tbody>
           </table>
+          <div id="hiddenWordsForTest" style={{ display: 'none' }} >
+            {this.state.words}
+          </div>
```
tuliomir, line 249: This is a helper for the end-to-end tests, which allows for easy fetching of all the words. Since this is being displayed only when the words are also being displayed visually in plain text to the user, I believe there is no security issue here.

andreabadesso (approved):

cypress/e2e/00-welcome.cy.js
```diff
@@ -0,0 +1,263 @@
...
+        },
+      });
+    });
+
+    cy.findByText('Next').click();
+
+    // PIN was successful
+    cy.contains('Loading transactions'); // For a few seconds this screen will be shown
+
+    // There is a timeout in place that needs to be waited. The error should be handled gracefully
```
pedroferreira1, line 260: Will the wallet load after this?

If yes, maybe we should have a verification that the screen does not contain the "You should do a backup" message as before in here

```
    cy.contains(`You haven't done the backup`);
```

pedroferreira1 (approved):

cypress/e2e/00-welcome.cy.js
tuliomir: The wallet won't load, as the history fetch from the fullnode has failed by network intercept. This test only validates the failure modal.

PR #699, opened 2024-11-28
refactor: improve amount handling

glevco:
### Motivation

In our network, the amount used by output values is always an integer number, that is, it cannot have a fractional part. However, we do have decimal places for _cosmetic only_ reasons: we show 2 decimal places for tokens and 0 decimal places for NFTs.

This means internally amount values should always be handled (stored, sent, processed) as integer values, and we should only consider decimal places when interfacing with the user, that is, displaying values or receiving input. This is straightforward in Python by using the `int` type. In JS, the `Number` type is used, which is a `double`.

This resulted in a bit of a mess in our wallet code. Currently, we have different functions/react components storing and expecting amounts in different formats. Sometimes, strings are used and they have to be processed as they contain thousand and decimal separators (`,` and `.`). Sometimes, `Numbers` are used, but depending on the function/component, it can expect it as a whole number or a fractional number. That is, the amount `12.34` may be represented as the fractional `12.34` JS `Number`, or the whole `1234.0` JS `Number`. On top of that, NFTs introduce different handling.

This means in multiple parts of the code we have to consider decimal places and convert back and forth between formats. This is not only redundant but error prone. What we should do instead, is only ever store and process amounts as whole `Numbers`, that is, the amount `12.34` should always be handled as the `1234.0` JS `Number`. We should never use fractional parts for amounts internally. This PR implements this refactor, removing the `decimalToInteger` conversion function that was used to handle this, and now amounts in React props and states are always whole.

This is done as part of the conversion to `BigInt` project. After this refactor, changing the amount type from `Number` to `BigInt` will be straight forward, as we won't have to deal with any conversion logic — the amount values will already be integers, but typed as `Numbers`, so the conversion to `BigInts` is simply a type conversion.

The refactor is implemented by changing the way the `InputNumber` component is used. Currently, its value is available as a `string` through `ref` or `onValueChange`. Then, it's the component's user responsibility to handle format conversion. The component now handles everything as whole `Numbers`, so users can work with them directly, and only consider decimals when pretty printing. The `ref` is removed as it provided access to the underlying `<input>` element, which is of `text` type. Instead, users should use the updated `onValueChange` with `Numbers`.

**ATTENTION**: The changes in this PR are extremely sensitive and important as they deal with conversion between user input and internal representation. It should be reviewed with extra caution.

### Acceptance Criteria

- Update `InputNumber` component
  - Change `defaultValue` prop and `value` state from `string` to `number`.
  - Remove `precision` prop and introduce `isNFT` instead. This allows for removing some duplication.
  - Remove support for `ref`.
  - For further simplification, remove unused props `separator`, `locale`, `onChange`.
- Remove `decimalToInteger` function.

### Security Checklist

### DIFF
--- a/src/components/InputNumber.js
+++ b/src/components/InputNumber.js
@@ -8,7 +8,7 @@
 import React, { useState, useLayoutEffect, useRef, useEffect } from "react";
 import PropTypes from "prop-types";
 import { useSelector } from 'react-redux';
-import hathorLib from "@hathor/wallet-lib";
+import { numberUtils } from "@hathor/wallet-lib";
 
 /**
  * Component that enhances typing numbers
@@ -28,43 +28,38 @@ const InputNumber = React.forwardRef(
   (
     {
       defaultValue,
-      precision,
-      separator,
-      locale,
+      isNFT,
       onValueChange,
       requirePositive,
       ...otherProps
     },
     ref
   ) => {
-
-    const decimalPlaces = useSelector((state) => state.serverInfo.decimalPlaces);
-    const decimalPrecision = precision ?? decimalPlaces;
+    if (ref !== null) {
+        // TODO: We should just convert InputNumber from a React.forwardRef to a normal React.Component,
+        //  but we may do this in a separate PR
+        const msg = 'do not use ref in InputNumber, instead use onValueChange'
+        console.error(msg)
+        throw Error(msg)
+    }
+    const decimalPlaces = isNFT ? 0 : useSelector((state) => state.serverInfo.decimalPlaces);
 
     /**
-     * Formats a string following the pattern 9,999.99. It decomposes rawValue into decimal and fractional parts, mainly to add the thousands separator.
+     * Formats a number following the pattern 9,999.99
      *
-     * @param {string} rawValue String to be formatted
+     * @param {number} rawValue Number to be formatted
      *
      * @return {string} Formatted string
      */
-    const format = (rawValue = "") => {
-      const value = String(rawValue)
-        .replace(/[^\d]/g, "")
-        .padStart(decimalPrecision + 1, "0");
-      const decimalPart = Intl.NumberFormat(locale).format(
-        value.substr(0, value.length - decimalPrecision)
-      );
-      const fractionalPart = value.substr(value.length - decimalPrecision);
-      if (fractionalPart.length === 0) {
-        return decimalPart;
-      } else {
-        return `${decimalPart}${separator}${fractionalPart}`;
+    const format = (rawValue) => {
+      if (typeof rawValue !== 'number') {
+          throw Error(`value must be a number: ${rawValue}`)
       }
+      return numberUtils.prettyValue(rawValue, decimalPlaces)
     };
 
-    const innerRef = ref || useRef();
-    const [value, setValue] = useState(format(defaultValue));
+    const innerRef = useRef();
+    const [value, setValue] = useState(defaultValue);
 
     /**
      * Listen keydown events while this component is focused overriding the default native input behavior.
@@ -78,7 +73,7 @@ const InputNumber = React.forwardRef(
         const isBackspace = evt.key === "Backspace" || evt.key === "Delete";
         const isDeleteAll =
           isBackspace &&
-          evt.target.selectionEnd - evt.target.selectionStart >= value.length;
+          evt.target.selectionEnd - evt.target.selectionStart >= format(value).length;
         const isCtrlOrMeta = evt.ctrlKey || evt.metaKey;
 
         // Do not handle keyboard events when ctrlKey or metaKey are present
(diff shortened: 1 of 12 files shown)


src/screens/CreateNFT.js
```diff
@@ -314,7 +314,7 @@ function CreateNFT() {
         <div className="row">
           <div className="form-group col-4">
             <label>{t`Amount`}</label>
-            <InputNumber required className="form-control" precision={0} placeholder="How many NFT units to create" onValueChange={onAmountChange} />
```
pedroferreira1, line 317: Why did you remove the placeholder?

src/components/OutputsWrapper.js
```diff
@@ -55,11 +54,7 @@ class OutputsWrapper extends React.Component {
   render = () => {
     const renderInputNumber = () => {
       const classNames = "form-control output-value col-2";
-      if (this.props.isNFT) {
-        return <InputNumber key="nft-value" ref={this.value} className={classNames} placeholder="0" precision={0} />;
-      } else {
-        return <InputNumber key="value" ref={this.value} placeholder={hathorLib.numberUtils.prettyValue(0, this.props.decimalPlaces)} className={classNames} />;
-      }
+      return <InputNumber key={this.props.isNFT ? "nft-value" : "value"} onValueChange={(value) => this.value = value} className={classNames} isNFT={this.props.isNFT} />;
```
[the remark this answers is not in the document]
glevco: I purposefully removed all placeholder values as they were not used. The `InputNumber` component always provides the `defaultValue` which guarantees the input will always display `0.00`. If you compare inputs in `master` and this branch, you should see that they're exactly the same.

src/components/atomic-swap/ModalAtomicReceive.js
```diff
@@ -116,9 +115,7 @@ export function ModalAtomicReceive ({ sendClickHandler, receivableTokens, manage
                                 <label htmlFor="amount" className="col-3 my-auto">{t`Amount`}: </label>
                                 <InputNumber key="value"
                                              name="amount"
-                                             ref={amountRef}
-                                             defaultValue={hathorLib.numberUtils.prettyValue(amount, decimalPlaces)}
-                                             placeholder={hathorLib.numberUtils.prettyValue(0, decimalPlaces)}
+                                             defaultValue={amount}
```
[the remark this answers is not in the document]
glevco: The HTML `<input>` element's `value` is a `string` — as it was in `master`, so it supports separators (e.g. `1,000.00`).

The `InputNumber` component's `defaultValue` prop and `value` state are `Number`, so all handling is done with (whole) numbers. Then, the `value` state is converted to a `string` with `format` so it can be used in the HTML element's `value`.

src/screens/CreateNFT.js
glevco: Please see https://github.com/HathorNetwork/hathor-wallet/pull/699#discussion_r1869887972

src/screens/nano-contract/NanoContractExecuteMethod.js
```diff
@@ -264,8 +263,7 @@ function NanoContractExecuteMethod() {
       }
 
       if (typeToCheck === 'Amount') {
-        const amountValue = walletUtils.decimalToInteger(value, decimalPlaces);
-        argValues.push(amountValue);
+        argValues.push(value);
```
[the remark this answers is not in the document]
glevco: It's a `Number`. The only place it's a `string` is in the HTML element (see https://github.com/HathorNetwork/hathor-wallet/pull/699#discussion_r1869901737)

src/components/InputNumber.js
```diff
@@ -28,43 +28,36 @@ const InputNumber = React.forwardRef(
...
     },
     ref
   ) => {
-
-    const decimalPlaces = useSelector((state) => state.serverInfo.decimalPlaces);
-    const decimalPrecision = precision ?? decimalPlaces;
+    if (ref !== null) {
+        // TODO: We should just convert InputNumber from a React.forwardRef to a normal React.Component,
+        //  but we may do this in a separate PR
+        throw Error('do not use ref in InputNumber, instead use onValueChange')
```
r4mmer, line 41: Can you also `console.error` this to make this error easier to catch during development.

r4mmer (dismissed):

src/components/InputNumber.js
glevco: Done in https://github.com/HathorNetwork/hathor-wallet/pull/699/commits/07afc271fef65073accfb308cfb99d9901b7a469

src/screens/nano-contract/NanoContractExecuteMethod.js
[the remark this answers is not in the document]
glevco: I fixed the nano execution page in a new commit: https://github.com/HathorNetwork/hathor-wallet/pull/699/commits/efa58fe608c5c7f71f6e2f70e90f6cb39326f392

src/screens/nano-contract/NanoContractExecuteMethod.js
```diff
@@ -140,8 +144,8 @@ function NanoContractExecuteMethod() {
       if (typeToCheck !== 'Address') {
         continue;
       }
-      const addressInputValue = formRefs[i].current.value;
-      validateAddress(addressInputValue, formRefs[i].current);
+      const addressInputValue = formRefs[i].getValue();
+      validateAddress(addressInputValue, formRefs[i].ref);
```
pedroferreira1, line 148: Shouldn't this be `formRefs[i].ref.current`? Is it still working when validating the address?

glevco: You're right, that was my mistake. Fixed in https://github.com/HathorNetwork/hathor-wallet/pull/699/commits/37e16014eefa804dc9a627366f7ee2d7dca3b186. Note: this is only for custom address inputs, so it didn't fail for neither the sign address nor withdraw address inputs, only for Dozer's custom address arg. We should probably include those in the QA.

pedroferreira1 (approved):

r4mmer (approved):

PR #712, opened 2024-12-23
feat: lavamoat

andreabadesso:
### Motivation

Add LavaMoat to protect against supply chain attacks and upgrade react-scripts to version 5.

### Description

Added LavaMoat webpack plugin to production builds only. This required upgrading react-scripts from v4 to v5, which in turn required several changes:

1. Added react-app-rewired to customize the webpack config (and add the lavamoat plugin)
2. Added polyfills that were removed in react-scripts v5
3. Created buffer-shim.js to handle Buffer compatibility issues in the elliptic package
4. Updated imports in our code to use browserify versions (e.g. path -> path-browserify)

### Acceptance Criteria
- LavaMoat runs only in production builds as `@lavamoat/webpack` is missing a hook
- Development builds work without LavaMoat
- All functionality remains unchanged with lavamoat and react-scripts v5 activated


lavamoat/webpack/policy.json
```diff
@@ -0,0 +1,1383 @@
+{
```
andreabadesso, line 1: This file is generated by lavamoat automatically (because we have `generatePolicy: true` in our plugin config). All manual changes are done in the `policy-override.json`

src/components/ModalPin.js
```diff
@@ -79,9 +77,18 @@ export class ModalPin extends React.Component {
 
     // Set the PIN on the instance variable and close the modal.
     this.pin = pin;
+    this.onSuccess();
```
tuliomir, line 80: suggestion: Replace this with `this.onModalHidden()`. The logic will be the same, as I see it.

src/components/ModalPin.js
```diff
@@ -79,9 +77,18 @@ export class ModalPin extends React.Component {
...
+    this.props.onSuccess({ pin: this.pin })
+    this.props.onClose();
+  }
+
+  onCancel(e) {
+    e.preventDefault();
 
-    // Necessary callbacks will be executed at the `onHidden` modal event
-    $('#modalPin').modal('hide');
+    this.props.onClose();
```
tuliomir, line 91: suggestion: Replace this too with `this.onModalHidden()`. The logic will be the same, as I see it.

In short, centralize all modal closing logic in this function, to simplify it.

src/components/ModalPin.js
```diff
@@ -101,14 +108,15 @@ export class ModalPin extends React.Component {
...
+    return (
+      <div>
         <div className="modal fade" id="modalPin" tabIndex="-1" role="dialog" aria-labelledby="modalPin"
              aria-hidden="true">
           <div className="modal-dialog" role="document">
             <div className="modal-content">
               <div className="modal-header">
-                <h5 className="modal-title" id="exampleModalLabel">{t`Write your PIN`}</h5>
-                <button type="button" className="close" data-dismiss="modal" aria-label="Close">
+                <h5 className="modal-title" id="pinModal">{t`Write your PIN`}</h5>
```
tuliomir, line 118: suggestion: This identifier may cause confusion with the modal div itself. How about `pinModalInputTitle`, or even remove the unique identifier altogether?

src/store/index.js
```diff
@@ -11,15 +11,14 @@ import createSagaMiddleware from 'redux-saga';
 import rootReducer from '../reducers/index';
 import rootSagas from '../sagas';
 
-
-const saga = createSagaMiddleware();
-const middlewares = [saga, thunk];
+export const sagaMiddleware = createSagaMiddleware();
```
tuliomir, line 14: question: Why do we need to export the saga middleware?

src/utils/helpers.js
```diff
@@ -194,7 +194,7 @@ const helpers = {
    * @inner
    */
   getFullExplorerURL(urlPath) {
-    return path.join(this.getExplorerURL(), urlPath);
+    path.join(this.getExplorerURL(), urlPath);
```
tuliomir, line 197: chore: It's necessary to return this string, right?

tuliomir (changes requested):

config-overrides.js
```diff
@@ -0,0 +1,107 @@
...
+  const basePlugins = [
+    ...config.plugins.filter(p => !(p instanceof webpack.ProvidePlugin)),
+    new webpack.ProvidePlugin({
+      process: 'process/browser',
+      Buffer: ['buffer-shim', 'default']
+    }),
+  ];
+
+  // Only add LavaMoat in production
+  if (env === 'production') {
```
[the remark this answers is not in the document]
andreabadesso: I improved the comment to explain that lavamoat does not work in the dev environment because of the hot reload feature, lavamoat-webpack does not rebuild when the code changes

lavamoat/webpack/policy-override.json
```diff
@@ -0,0 +1,332 @@
...
+        "window.setInterval": true,
+        "window.clearInterval": true,
+        "window.requestAnimationFrame": true,
+        "window.cancelAnimationFrame": true,
+        "window.getComputedStyle": true,
+        "window.Event": true,
+        "window.CustomEvent": true,
+        "window.MutationObserver": true,
+        "MutationObserver": true,
+        "performance": true
```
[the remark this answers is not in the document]
andreabadesso: We don't

Removed it, thanks!

package.json
```diff
@@ -63,14 +71,14 @@
...
     "build-css": "sass --no-source-map src/index.module.scss src/index.css",
     "watch-css": "npm run build-css && sass --no-source-map -w src/index.module.scss src/index.css",
-    "start-js": "react-scripts --openssl-legacy-provider start",
+    "start-js": "react-app-rewired --openssl-legacy-provider start",
     "start": "npm-run-all -p watch-css start-js",
-    "build-js": "react-scripts --openssl-legacy-provider build",
-    "build": "npm-run-all build-css build-js",
-    "test": "react-scripts test --env=./tests/env.js",
+    "build-js": "react-app-rewired --openssl-legacy-provider build",
+    "build": "react-app-rewired build",
```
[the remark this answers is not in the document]
andreabadesso: Yes, the CSS is working correctly! react-app-rewired handles CSS processing automatically as part of its build process, so we don't need the separate build-css step anymore. The CSS files are being properly compiled and included in the build output when running react-app-rewired build.

package.json
```diff
@@ -63,14 +71,14 @@
...
+    "test": "react-app-rewired test",
     "e2e": "cypress run",
     "eject": "react-scripts eject",
-    "electron": "electron --inspect=5858 .",
+    "electron": "electron --inspect=5858 --trace-uncaught .",
```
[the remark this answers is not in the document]
andreabadesso: Yes!

Done, thanks

src/App.js
```diff
@@ -109,13 +108,12 @@ function Root() {
     if (IPC_RENDERER) {
       // Event called when the user wants to reset all data
       IPC_RENDERER.on('app:clear_storage', async () => {
-        context.showModal(MODAL_TYPES.CONFIRM_CLEAR_STORAGE, {
+        modalContext.showModal(MODAL_TYPES.CONFIRM_CLEAR_STORAGE, {
```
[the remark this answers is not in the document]
andreabadesso: Wow, this really was broken

thanks

src/store/index.js
andreabadesso: We don't need to since we're not testing the sagas, but it will be useful when we start doing so

src/utils/helpers.js
andreabadesso: Definitely, fixed!

Thanks

patches/elliptic++bn.js+4.12.0.patch
```diff
@@ -11,19 +10,21 @@ index 3a4371e..629ed9c 100644
...
-+      }
-     }
+-    }
++    var customBuffer = require('../../../../../src/buffer-shim');
++    BNBuffer = customBuffer.Buffer || customBuffer.default;
    } catch (e) {
++    try {
++      BNBuffer = require('buffer/').Buffer;
++    } catch (e2) {
++      BNBuffer = null;
```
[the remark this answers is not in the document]
andreabadesso: Added the console.error, thanks

src/components/ModalPin.js
```diff
@@ -27,28 +27,26 @@ export class ModalPin extends React.Component {
...
-      if (this.pin) {
-        this.props.onSuccess({ pin: this.pin })
-      }
-    });
+    $('#modalPin').on('hidden.bs.modal', this.onModalHidden);
 
     // Focus the PIN field on modal load
-    $('#modalPin').on('shown.bs.modal', (e) => {
-      this.refs.pinInput.refs.pin.focus();
-    });
```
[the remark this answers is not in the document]
andreabadesso: It does, but no need to remove this. 

Thanks

src/components/ModalPin.js
andreabadesso: I agree, thanks!

src/components/ModalPin.js
andreabadesso: Done, thanks!

src/components/ModalPin.js
andreabadesso: Done! Thanks

tuliomir (dismissed):

config-overrides.js
```diff
@@ -0,0 +1,108 @@
...
+  config.module.rules.push({
+    test: /[\\/]axios[\\/]/,
+    resolve: {
+      alias: {
+        'axios': path.resolve(__dirname, 'node_modules/axios')
+      }
+    }
+  });
+
+  config.module.rules.push({
```
pedroferreira1, line 63: Add comment for this rule also

lavamoat/webpack/policy-override.json
```diff
@@ -0,0 +1,328 @@
...
+        "getComputedStyle": true,
+        "window.getComputedStyle": true,
+        "requestAnimationFrame": true,
+        "cancelAnimationFrame": true
+      }
+    },
+    "jquery": {
+      "globals": {
+        "document": true,
+        "window": true,
```
pedroferreira1, line 54: When you allow `window`, does it allow all `window.*`?

lavamoat/webpack/policy-override.json
```diff
@@ -0,0 +1,328 @@
...
+        "navigator": true,
+        "location": true
+      }
+    },
+    "viz.js": {
+      "globals": {
+        "document.title": true
+      }
+    },
+    "unleash-proxy-client": {
```
pedroferreira1, line 97: This is our package, right?

lavamoat/webpack/policy-override.json
```diff
@@ -0,0 +1,328 @@
...
+        "util": true
+      }
+    },
+    "@hathor/wallet-lib": {
+      "globals": {
+        "process": true,
+        "console": true,
+        "Error": true,
+        "Object": true,
+        "XMLHttpRequest": true,
```
pedroferreira1, line 257: We send requests only through axios. Does it make sense to remove this and allow only in axios?

pedroferreira1 (approved):

pedroferreira1 (commented):
Waiting dependencies review

lavamoat/webpack/policy-override.json
andreabadesso: Not really, lavamoat follows the principle of least privilege, so setting `window: true` only gives basic access to the window object itself, so you do these:

```
if (window) {
...
}
```

`typeof window`

But you need specific permissions for window's methods, e.g.:

// These need explicit permissions like "window.localStorage": true
window.localStorage.getItem('key')
window.sessionStorage.setItem('key', 'value')

// These need "window.location": true
window.location.href
window.location.reload()

// These need "window.history": true
window.history.pushState()
window.history.back()

// These need "window.alert": true
window.alert('message')

lavamoat/webpack/policy-override.json
andreabadesso: It does make sense, removed XMLHttpRequest and some others that are not being used

package.json
```diff
@@ -45,61 +51,70 @@
...
     "react-router-dom": "6.22.3",
-    "react-scripts": "4.0.3",
+    "react-scripts": "5.0.1",
     "redux-saga": "1.2.1",
     "redux-thunk": "2.4.1",
     "ttag": "1.8.6",
     "unleash-proxy-client": "3.2.0",
-    "viz.js": "2.1.2"
+    "viz.js": "2.1.2",
+    "node-stdlib-browser": "1.2.0"
```
pedroferreira1, line 67: Do we need it here? Or just in the `devDependencies`? I'm asking just because you also added it there.

pedroferreira1 (approved):

config-overrides.js
```diff
@@ -0,0 +1,105 @@
...
+    ...config.resolve,
+    fallback: {
+      buffer: require.resolve('buffer/'),
+      assert: stdLibBrowser.assert,
+      crypto: stdLibBrowser.crypto,
+      path: stdLibBrowser.path,
+      process: stdLibBrowser.process,
+      stream: stdLibBrowser.stream,
+      os: stdLibBrowser.os,
+      vm: false,
```
pedroferreira1, line 28: What's this?

pedroferreira1 (approved):

config-overrides.js
andreabadesso: It means we're not polyfilling it, `false` is the default, but react-scripts gives a warning if we don't do it:

```
BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
        - add a fallback 'resolve.fallback: { "vm": require.resolve("vm-browserify") }'
        - install 'vm-browserify'
If you don't want to include a polyfill, you can use an empty module like this:
        resolve.fallback: { "vm": false }
```

.github/workflows/main.yml
```diff
@@ -27,8 +27,14 @@ jobs:
...
       with:
         node-version: ${{ matrix.node-version }}
+    - name: Install system dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y libusb-1.0-0-dev libudev-dev
     - name: Install dependencies
       run: npm install
+    - name: Install ttag-cli globally
+      run: npm install -g ttag-cli
```
tuliomir, line 37: question(non-blocking): Why do we need to install it globally?

tuliomir (approved):

.github/workflows/main.yml
andreabadesso: For some reason, it's failing to find `ttag` in the node_modules

I think it's not being added to the runner's PATH

PR #735, opened 2025-02-12
feat: send tokens rpc

andreabadesso:
https://github.com/user-attachments/assets/d7027146-f70b-4f2f-b1c4-981f0d3c3de4

### Acceptance Criteria
- We should be able to handle `sendTransaction` RPC requests
- We should be able to handle data outputs
- We should register tokens used in outputs or inputs automatically
- We should handle insufficient balance errors properly 

### Security Checklist


package.json
```diff
@@ -31,19 +31,28 @@
   "resolutions": {
     "axios": "1.7.7",
     "@hathor/wallet-lib/axios": "1.7.7",
-    "@hathor/hathor-rpc-handler/@hathor/wallet-lib/axios": "1.7.7"
+    "@hathor/hathor-rpc-handler/@hathor/wallet-lib/axios": "1.7.7",
+    "bitcore-lib": "8.25.10",
+    "@hathor/wallet-lib/bitcore-lib": "8.25.10",
+    "@hathor/hathor-rpc-handler/@hathor/wallet-lib/bitcore-lib": "8.25.10",
+    "../hathor-rpc-lib/node_modules/bitcore-lib": "8.25.10",
```
raul-oliveira, line 38: Is this relative path right?

andreabadesso: It's not, I will change this as soon as the rpc-lib is published

raul-oliveira (approved):
Approved. Don't forget to change the dependencies to the published version

src/components/Reown/modals/SendTransactionModal.js
```diff
@@ -0,0 +1,198 @@
...
+ */
+
+import React, { useEffect } from 'react';
+import { t } from 'ttag';
+import { useSelector, useDispatch } from 'react-redux';
+import { constants, numberUtils } from '@hathor/wallet-lib';
+import { unregisteredTokensDownloadRequested } from '../../../actions';
+import { CopyButton } from '../../CopyButton';
+
+export function SendTransactionModal({ data, firstAddress, onAccept, onReject }) {
```
pedroferreira1, line 15: Why do you need `firstAddress` in this component?

src/components/Reown/TransactionFeedbackModal.js
```diff
@@ -0,0 +1,113 @@
...
+        <>
+          <div className="modal-body text-center">
+            <ReactLoading type="spin" color={colors.purpleHathor} height={32} width={32} className="d-inline-block" />
+            <p className="mt-3">{t`Processing transaction...`}</p>
+          </div>
+          <div className="modal-footer">
+            <button type="button" className="btn btn-secondary" onClick={handleDismiss}>
+              {t`Cancel`}
+            </button>
+          </div>
```
pedroferreira1, line 53: Does it make sense to allow cancelling the request while it's processing?

src/actions/index.js
```diff
@@ -102,6 +105,14 @@ export const types = {
   SHOW_GLOBAL_MODAL: 'SHOW_GLOBAL_MODAL',
   HIDE_GLOBAL_MODAL: 'HIDE_GLOBAL_MODAL',
   SERVER_INFO_UPDATED: 'SERVER_INFO_UPDATED',
+  REOWN_SEND_TX_STATUS_LOADING: 'REOWN_SEND_TX_STATUS_LOADING',
+  REOWN_SEND_TX_STATUS_READY: 'REOWN_SEND_TX_STATUS_READY',
+  REOWN_SEND_TX_STATUS_SUCCESS: 'REOWN_SEND_TX_STATUS_SUCCESS',
+  REOWN_SEND_TX_STATUS_FAILURE: 'REOWN_SEND_TX_STATUS_FAILURE',
```
pedroferreira1, line 111: In all other RPC requests you use `FAILED` instead of `FAILURE`, e.g., `REOWN_NEW_NANOCONTRACT_STATUS_FAILED`. I think we should keep consistency

src/actions/index.js
```diff
@@ -871,6 +882,11 @@ export const showNanoContractSendTxModal = (onAccept, onReject, data, metadata)
   payload: { accept: onAccept, deny: onReject, data, dapp: metadata },
 });
 
+export const showSendTransactionModal = (onAccept, onReject, data, metadata) => ({
```
pedroferreira1, line 885: Missing docstring

src/actions/index.js
```diff
@@ -886,3 +902,65 @@ export const showGlobalModal = (modalType, modalProps = {}) => ({
...
+ */
+export const unregisteredTokensDownloadFailure = (error) => ({
+  type: types.UNREGISTERED_TOKENS_DOWNLOAD_FAILURE,
+  payload: { error },
+});
+
+/**
+ * End of unregistered tokens download process
+ */
+export const unregisteredTokensDownloadEnd = () => ({
```
pedroferreira1, line 964: What's the difference between the success and end events?

src/components/Reown/modals/SendTransactionModal.js
```diff
@@ -0,0 +1,198 @@
...
+    tokenMetadata: state.tokenMetadata,
+    tokens: state.tokens,
+  }));
+
+  useEffect(() => {
+    // Collect all unregistered token UIDs
+    const unregisteredTokens = new Set();
+    
+    data?.data?.inputs?.forEach(input => {
+      const token = registeredTokens.find(t => t.uid === input.token);
```
pedroferreira1, line 27: What happens if the token is HTR? Will it already be in the redux `tokens` array?

src/components/Reown/modals/SendTransactionModal.js
```diff
@@ -0,0 +1,198 @@
...
+            </div>
+            <div className="text-monospace">
+              {truncateTxId(input?.txId)} ({input?.index})
+              {input?.txId && (
+                <button 
+                  className="btn btn-link btn-sm p-0 ml-2" 
+                  onClick={() => navigator.clipboard.writeText(input.txId)}
+                >
+                  <i className="fa fa-copy"></i>
+                </button>
```
pedroferreira1, line 97: Why don't you use the `CopyButton` component?

src/components/Reown/modals/SendTransactionModal.js
```diff
@@ -0,0 +1,198 @@
...
+              <div className="text-monospace">
+                {data.data.changeAddress}
+                {data.data.changeAddress && (
+                  <button 
+                    className="btn btn-link btn-sm p-0 ml-2" 
+                    onClick={() => navigator.clipboard.writeText(data.data.changeAddress)}
+                  >
+                    <i className="fa fa-copy"></i>
+                  </button>
+                )}
```
pedroferreira1, line 186: Why don't you use the `CopyButton` component?

src/reducers/index.js
```diff
@@ -1500,4 +1502,28 @@ export const onUpdateNetworkSettings = (state, { payload }) => {
...
+    symbol: token.symbol
+  }));
+
+  return {
+    ...state,
+    tokens: [...state.tokens, ...newTokens],
+    tokenMetadata: {
+      ...state.tokenMetadata,
+      ...tokens
+    }
```
pedroferreira1, line 1525: I didn't understand this. To update the token metadata you should call the metadata API.

src/components/Reown/modals/SendTransactionModal.js
```diff
@@ -0,0 +1,198 @@
...
+
+    return '?';
+  };
+
+  const formatValue = (value) => {
+    if (value == null) {
+      return '-';
+    }
+
+    return numberUtils.prettyValue(value);
```
pedroferreira1, line 63: We should consider if the token is an NFT here as well.

src/components/Reown/modals/SendTransactionModal.js
andreabadesso: It was unused, removed it, thanks

src/components/Reown/TransactionFeedbackModal.js
andreabadesso: Added a notice making it clear that cancelling doesn't actually cancel the transaction and only closes the modal.

What do you think? 

I can also remove the button

src/actions/index.js
andreabadesso: I agree, refactored, thanks

src/sagas/reown.js
```diff
@@ -446,6 +456,43 @@ export function* processRequest(action) {
...
+          retryHandler,
+          types.REOWN_SEND_TX_RETRY,
+          types.REOWN_SEND_TX_RETRY_DISMISS,
+        );
+
+        if (retry) {
+          shouldAnswer = false;
+          yield* processRequest(action);
+        }
+      } break;
```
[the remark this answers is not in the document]
andreabadesso: Yes, refactored, thanks

src/components/Reown/modals/SendTransactionModal.js
```diff
@@ -0,0 +1,198 @@
...
+            </div>
+            <div className="text-monospace">
+              {output?.address}
+              <CopyButton text={output?.address} />
+            </div>
+            {output?.data && (
+              <>
+                <div className="mt-2 mb-1"><small>{t`Data field:`}</small></div>
+                <div className="text-monospace">
+                  {output.data.join(',')}
```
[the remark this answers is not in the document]
andreabadesso: Yes

src/components/Reown/modals/SendTransactionModal.js
andreabadesso: Refactored! Thanks

src/components/Reown/modals/SendTransactionModal.js
andreabadesso: Done, thanks

src/reducers/index.js
andreabadesso: This was indeed wrong and not fetching the actual token metadata.

Refactored the `requestUnregisteredTokensDownload` to actually dispatch the action to fetch tokens metadata

Since this code was copied from the wallet-mobile, I'll check if we're doing the same there

src/components/Reown/modals/SendTransactionModal.js
andreabadesso: Refactored this to skip HTR

src/components/Reown/modals/SendTransactionModal.js
```diff
@@ -0,0 +1,198 @@
...
+      }
+    });
+
+    if (unregisteredTokens.size > 0) {
+      dispatch(unregisteredTokensDownloadRequested(Array.from(unregisteredTokens)));
+    }
+  }, [data, registeredTokens, dispatch]);
+
+  const getTokenSymbol = (tokenId) => {
+    if (!tokenId) {
```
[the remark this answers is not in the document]
andreabadesso: We were defaulting to `HTR`, which doesn't make much sense indeed

I refactored this code, can you check again?

src/components/Reown/modals/SendTransactionModal.js
```diff
@@ -0,0 +1,198 @@
...
+      return constants.DEFAULT_NATIVE_TOKEN_CONFIG.symbol;
+    }
+
+    const token = registeredTokens.find(t => t.uid === tokenId);
+    if (token) {
+      return token.symbol;
+    }
+
+    return '?';
```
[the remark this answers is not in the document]
andreabadesso: It might happen for a brief moment for unregistered tokens while it's being downloaded

src/components/Reown/modals/SendTransactionModal.js
andreabadesso: Refactored to use it, thanks

src/components/Reown/TransactionFeedbackModal.js
pedroferreira1: I think it's ok with the note but I'd rather keep the modal opened so we guarantee to the user that the tx was processed. What happens if the processing fails? Will the user be notified after the modal is closed?

src/components/Reown/modals/SendTransactionModal.js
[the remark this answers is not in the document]
pedroferreira1: Maybe we could leave as empty string, since it will be updated as soon as the download arrives. I believe the '?' blinking might be confusing to the user. What do you think?

src/components/Reown/modals/SendTransactionModal.js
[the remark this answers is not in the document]
pedroferreira1: Where does it come from? And why can we have an array of strings in the output data?

src/components/Reown/modals/SendTransactionModal.js
```diff
@@ -0,0 +1,209 @@
...
+
+  const formatValue = (value, tokenId) => {
+    if (value == null) {
+      return '-';
+    }
+
+    // Check if the token is an NFT using the helpers utility
+    const isNFT = tokenId && helpers.isTokenNFT(tokenId, tokenMetadata);
+    
+    return numberUtils.prettyValue(value, isNFT ? 0 : undefined);
```
pedroferreira1, line 84: Why `undefined`? Should we use the `decimalPlaces` from redux?

src/components/Reown/modals/SendTransactionModal.js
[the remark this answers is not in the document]
pedroferreira1: As we discussed on Slack we could receive a list of data strings but each string will create a different output.

src/components/Reown/modals/SendTransactionModal.js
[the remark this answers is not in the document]
andreabadesso: I agree, done

src/components/Reown/modals/SendTransactionModal.js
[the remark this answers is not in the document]
andreabadesso: Refactored! Can you please check again?

src/components/Reown/TransactionFeedbackModal.js
```diff
@@ -0,0 +1,114 @@
...
+import ReactLoading from 'react-loading';
+import { colors } from '../../constants';
+
+export const MODAL_ID = 'transactionFeedbackModal';
+
+/**
+ * Component that shows a modal with feedback for transactions
+ * Shows loading, success or error message and provides retry option on failure
+ */
+export function TransactionFeedbackModal({ isError, isLoading = true, errorMessage, onClose, manageDomLifecycle }) {
```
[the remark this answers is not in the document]
andreabadesso: Done! Thanks

src/reducers/index.js
```diff
@@ -455,6 +455,9 @@ const rootReducer = (state = initialState, action) => {
         ...state,
         reown: reownReducer(state.reown, action),
       };
+    case types.UNREGISTERED_TOKENS_DOWNLOAD_SUCCESS:
+      return onUnregisteredTokensDownloadSuccess(state, action);
+    case types.UNREGISTERED_TOKENS_DOWNLOAD_FAILED:
```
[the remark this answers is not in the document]
andreabadesso: It is not doing anything, but I don't see any downsides on keeping it there, makes it explicit that it's not doing anything

src/components/Reown/TransactionFeedbackModal.js
andreabadesso: Refactored this feedback modal to use the FeedbackModal component, which will not show the cancel button here

PR #756, opened 2025-03-24
fix: invalid args

andreabadesso:
### Acceptance Criteria
- We should fix a bug where sending extra arguments to a nano contract crashes the wallet

### Security Checklist

### DIFF
--- a/src/components/Reown/modals/SendNanoContractTxModal.js
+++ b/src/components/Reown/modals/SendNanoContractTxModal.js
@@ -7,6 +7,7 @@
 
 import React, { useState, useEffect } from 'react';
 import { t } from 'ttag';
+import { get } from 'lodash';
 import { useDispatch, useSelector } from 'react-redux';
 import { types, setNewNanoContractStatusReady } from '../../../actions';
 import helpers from '../../../utils/helpers';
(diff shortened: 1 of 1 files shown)


src/components/Reown/modals/SendNanoContractTxModal.js
```diff
@@ -47,13 +48,13 @@ export function SendNanoContractTxModal({ data, onAccept, onReject }) {
...
     const methodInfo = blueprintInfo.public_methods?.[data.data.method];
-    const argEntries = methodInfo
-      ? data.data.args.map((arg, idx) => [
-          methodInfo.args[idx].name,
-          arg,
-          methodInfo.args[idx].type
-        ])
-      : data.data.args.map((arg, idx) => [t`Position ${idx}`, arg]);
+    const methodInfoArgs = get(methodInfo, 'args', []);
+    const dataArgs = get(data.data, 'args', []);
```
pedroferreira1, line 52: Can we show an error if `methodInfoArgs.length !== dataArgs.length`? I think it doesn't make sense to continue with the request if the arguments count is different than the expected count.

pedroferreira1 (approved):

src/components/Reown/modals/SendNanoContractTxModal.js
andreabadesso: We'll handle this in the rpc-lib when we start validating blueprint arguments

I don't think we should validate it in the wallets

r4mmer (approved):

PR #773, opened 2025-04-17
fix: network error redirects to invalid screen

andreabadesso:
### Acceptance Criteria
- Include here all things that this PR should solve

### Security Checklist


src/api/axiosInstance.js
```diff
@@ -58,7 +59,16 @@ const createRequestInstance = (resolve, timeout) => {
...
-    $('#requestErrorModal').modal('show');
+    // Use global modal system
+    store.dispatch(showGlobalModal(MODAL_TYPES.REQUEST_ERROR, {
+      lastFailedRequest: error.config,
+      requestErrorStatusCode: statusCode,
+      onClose: () => {
+        store.dispatch(lastFailedRequest(undefined));
+        store.dispatch(updateRequestErrorStatusCode(undefined));
+        store.dispatch(hideGlobalModal());
+      }
```
tuliomir, line 70: ⭐ Nice use of the `onClose()` feature!

src/components/RequestError.js
```diff
@@ -25,31 +24,35 @@ const MODAL_DOM_ID = '#requestErrorModal';
...
    * User clicked to change server, then push to choose server screen
    */
   const handleChangeServer = () => {
     $(MODAL_DOM_ID).modal('hide');
-    navigate('/server/');
+    if (onClose) onClose();
+    if (onChangeServer) {
+      onChangeServer();
+    } else {
+      window.location.hash = '#/network_settings_recovery';
```
tuliomir, line 46: question: Why not use `navigate()` here?

src/sagas/networkSettings.js
```diff
@@ -41,7 +40,17 @@ export function* changeNetworkSettings({ data, pin }) {
...
   let versionData;
   try {
-    versionData = yield call([wallet, wallet.getVersionData]);
+    const versionUrl = new URL('version', data.node).toString();
+    /*
+     * Using fetch instead of the wallet-lib axios instance to bypass
+     * the axios interceptor that is used to check for network errors.
+     * If this request fails, we don't want the network error modal to be shown.
+     */
+    const response = yield call(fetch, versionUrl);
```
tuliomir, line 49: question(non-blocking): Given that we've found the problem with Axios ( the same solved by https://github.com/HathorNetwork/hathor-explorer/pull/404 ), would it be a good idea to use a "blank" `axios` instance here instead of using `fetch`?

That way, we would bypass the lib interceptor, but keep the syntax that is already used throughout the app already.

andreabadesso: We don't use axios in the wallet-desktop except for the `axiosInstance` created from the wallet-lib (which is exactly what we don't want to do here)

So I prefer just using `fetch`, which is native since this is a simple http request

pedroferreira1 (approved):

tuliomir (approved):

src/screens/NetworkSettings.js
```diff
@@ -27,172 +27,23 @@ import { getGlobalWallet } from "../modules/wallet";
 import { useNavigate } from 'react-router-dom';
 import { networkSettingsRequestUpdate, setNetworkSettingsStatus } from "../actions";
 import BackButton from '../components/BackButton';
+import NetworkSettingsForm from '../components/NetworkSettingsForm';
```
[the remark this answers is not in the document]
andreabadesso: Done, thanks

pedroferreira1 (approved):

PR #809, opened 2025-09-24
fix: allow wallet to register nano create by another nano

pedroferreira1:
### Acceptance Criteria
- Allow wallet to register and get details of a nano that was created by another nano
- Remove address from nano screens

https://github.com/user-attachments/assets/d85b5604-0d96-4e12-bdd3-01accf946b5c

### Security Checklist

### DIFF
--- a/locale/texts.pot
+++ b/locale/texts.pot
@@ -34,7 +34,7 @@ msgid "Invalid PIN"
 msgstr ""
 
 #: src/screens/ChoosePassphrase.js:83
-#: src/screens/Settings.js:336
+#: src/screens/Settings.js:335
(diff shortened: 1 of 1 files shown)


src/index.module.scss
```diff
@@ -898,7 +898,7 @@ span.software {
 .nc-detail-wrapper {
   padding: 1rem;
   border: 1px solid;
-  max-width: 100%;
+  width: 100%;
```
pedroferreira1, line 901: The screen was really bad with data that had small width. I could've added a `min-width: 80%` but I preferred keeping it fixed as 100%

PR #821, opened 2025-11-12
feat: call reown init after network change

pedroferreira1:
### Motivation

We usually start the wallet connected to mainnet, then the `init` method for reown saga returns early, because we don't have nanos enabled. After changing to testnet, the `init` was not being called, not even the clear sessions, because we were depending on genesis hash update, which we don't have in the serverInfo, so it's always `undefined`.

Apart from that, I changed the `init` to be called in the `listenForNetworkChange` method, so we can store it in a variable and cancel it every time we change networks, in order to prevent leaks.

### Acceptance Criteria
- Call reown init and clear sessions after network change

### Security Checklist

### DIFF
--- a/src/sagas/reown.js
+++ b/src/sagas/reown.js
@@ -17,6 +17,7 @@ import {
   select,
   race,
   actionChannel,
+  cancel,
 } from 'redux-saga/effects';
 import { eventChannel } from 'redux-saga';
 import { get, values } from 'lodash';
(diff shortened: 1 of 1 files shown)


src/sagas/reown.js
```diff
@@ -166,23 +167,27 @@ function* init() {
...
+ * Cancels and restarts init when the network changes
  */
 export function* listenForNetworkChange() {
-  let previousGenesisHash = yield select((state) => get(state.serverInfo, 'genesisHash'));
+  // Start the initial init task
+  let initTask = yield fork(init);
 
   while (true) {
     // Wait for the server info to be updated with the new network data
     yield take(types.SERVER_INFO_UPDATED);
```
r4mmer: Could `SERVER_INFO_UPDATED` be emitted even if the network did not change?

PR #823, opened 2025-11-13
fix: some arg types must be rendered in a different way

pedroferreira1:
### Acceptance Criteria
- SignedData, Timestamp, Amount, TxOutputScript and TokenUid must be rendered in a different way.

### Security Checklist

### DIFF
--- a/src/components/Reown/modals/BaseNanoContractModal.js
+++ b/src/components/Reown/modals/BaseNanoContractModal.js
@@ -113,8 +113,7 @@ const ArgumentsTable = ({ args, decimalPlaces, tokens, network }) => {
         displayValue = `${parsedScript.address.base58} (${value})`;
       }
     } else if (type === 'TokenUid') {
-      const HTR_UID = constants.NATIVE_TOKEN_UID;
-      if (value === HTR_UID) {
+      if (value === constants.NATIVE_TOKEN_UID) {
         displayValue = `${constants.DEFAULT_NATIVE_TOKEN_CONFIG.symbol} (${value})`;
       } else if (tokens && value in tokens) {
         displayValue = `${tokens[value].symbol} (${value})`;


r4mmer (approved):

src/components/Reown/modals/BaseNanoContractModal.js
```diff
@@ -13,8 +13,22 @@ import helpers from '../../../utils/helpers';
...
 import { DAppInfo } from '../DAppInfo';
+import { SignedDataDisplay } from '../SignedDataDisplay';
+
+/**
+ * Parse script data from hex string
+ */
+const parseScriptData = (scriptData, network) => {
+  try {
+    const script = bufferUtils.hexToBuffer(scriptData);
+    return scriptsUtils.parseScript(script, network);
```
tuliomir: suggestion(non-blocking): We could reproduce the behavior from the `TxData` file, which adds a bit of error context.

https://github.com/HathorNetwork/hathor-wallet/blob/65a57c1ae0f680decc54c15418385b27c43da7a7/src/components/TxData.js#L555

tuliomir (dismissed):

src/components/Reown/modals/BaseNanoContractModal.js
pedroferreira1: This script doesn't need to be decoded. We try to decode but it can be literally anything. Having a message like this seems an error, which is not.

src/components/Reown/modals/BaseNanoContractModal.js
```diff
@@ -70,9 +84,50 @@ const BlueprintInfoCard = ({ nanoContract, blueprintInfo }) => (
...
+    } else if (type === 'TxOutputScript') {
+      const parsedScript = parseScriptData(value, network);
+      if (parsedScript && parsedScript.getType() === 'data') {
+        displayValue = `${parsedScript.data} (${value})`;
+      } else if (parsedScript) {
+        displayValue = `${parsedScript.address.base58} (${value})`;
+      }
+    } else if (type === 'TokenUid') {
+      const HTR_UID = constants.NATIVE_TOKEN_UID;
+      if (value === HTR_UID) {
```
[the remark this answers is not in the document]
pedroferreira1: done https://github.com/HathorNetwork/hathor-wallet/pull/823/commits/302b4c5a5521c0443f8d4476f062365c675aaf82

PR #859, opened 2026-03-26
feat: easy token import banner and modal

raul-oliveira:
Requires #856

### Acceptance Criteria
- When unregistered tokens exist, a banner appears on the Wallet screen prompting the user to import them
- Clicking "Add tokens" opens a modal where the user can select which tokens to register
- Tokens are registered via the existing `tokenRegisterRequested` saga with loading, success, and error states
- Banner can be dismissed for the session; reappears on app restart
- `UnknownTokens` screen removed, replaced by this new flow
- TokenBar "unknown tokens" icon now opens the import modal

### Design References
- [Banner](https://www.figma.com/design/F2SIdlG7tLHxGcfAfH2wqd/Easy-token-import-in-the-wallets?node-id=46-790) | [Modal](https://www.figma.com/design/F2SIdlG7tLHxGcfAfH2wqd/Easy-token-import-in-the-wallets?node-id=46-874)

### Screenshots
image
image
image
image
image
image

### Security Checklist

## Summary by CodeRabbit

* **New Features**
  * Token import banner alerts users to newly detected unknown tokens; “Add tokens” opens a modal to review, select, register, retry failures, and auto-close on success.
  * Incoming transactions now surface newly detected unknown tokens to the UI.

* **Removed**
  * Dedicated Unknown Tokens screen removed; workflow consolidated into banner + modal.

* **Localization & Styling**
  * Added UI strings and responsive styles for banner, modal, and token list.

* **Chores**
  * Banner dismissal persisted so closed banners remain hidden.


src/sagas/wallet.js
```diff
@@ -535,6 +536,23 @@ export function* handleNewTx(action) {
   const affectedTokens = yield call(handleTx, wallet, tx);
   const stateTokens = yield select((state) => state.tokens);
   const registeredTokens = stateTokens.map((token) => token.uid);
+  const stateAllTokens = yield select((state) => state.allTokens);
+
+  // Detect tokens in this tx that belong to this wallet and are not yet tracked in allTokens
+  const txBalance = yield call([wallet, wallet.getTxBalance], tx);
```
raul-oliveira, line 542: @pedroferreira1 the wallet.getTxBalance already return a filtered list of tokens that belongs to the wallet.

src/index.module.scss
```diff
@@ -481,6 +485,38 @@ svg.svg-wrapper {
   }
 }
 
+.token-import-banner {
```
raul-oliveira, line 488: @pedroferreira1 should this styles be placed here or should we wrap this inside the component?

pedroferreira1: It's fine to keep here

src/components/ModalTokenImport.js
```diff
@@ -0,0 +1,437 @@
...
+
+    async function fetchDetails() {
+      const wallet = getGlobalWallet();
+      const details = {};
+
+      await Promise.all(
+        unknownTokens.map(async (token) => {
+          try {
+            // Try MemoryStore first (populated during tx processing, no API call)
+            const tokenData = await wallet.storage.store.getToken(token.uid);
```
pedroferreira1: We should avoid accessing storage methods and store methods even more. I know that we don't have a method for it in the wallet yet but we could use the storage method, instead of the store one. 

`wallet.storage.getToken`

src/components/ModalTokenImport.js
```diff
@@ -0,0 +1,437 @@
...
+  const decimalPlaces = useSelector((state) => state.serverInfo.decimalPlaces);
+
+  // Current modal state
+  const [modalState, setModalState] = useState(MODAL_STATE.LOADING);
+
+  // Map of uid -> { uid, name, symbol, balance } for all resolved tokens
+  const [tokenDetails, setTokenDetails] = useState({});
+
+  // Set of selected token uids
+  const [selected, setSelected] = useState(new Set());
```
pedroferreira1: I'm not sure I like this data structure for storing the selected tokens. To know ir a token is selected, it must check in a set, to toggle, it must remove or add from the set. I know that this Will not grow in a point where this is not performatic but we could just add a new key to the `tokenDetails` map and it's indexed by uid, so every get is O(1). I think this would also make the other methods (e.g. toggle) easier

What do you think?

src/components/ModalTokenImport.js
```diff
@@ -0,0 +1,437 @@
...
+    };
+  }, [unknownTokens]);
+
+  /**
+   * Toggle an individual token's selection.
+   * @param {string} uid
+   */
+  const toggleToken = (uid) => {
+    setSelected((prev) => {
+      const next = new Set(prev);
```
tuliomir: issue: This is a weird way of interacting with sets. We rebuild it from scratch just to change a property?

If we'll keep using a `Set`, we should have it as a `useRef()` instead, so that we can have a single instance of the set and use it as it was meant to be used.

src/components/ModalTokenImport.js
```diff
@@ -0,0 +1,437 @@
...
+   * for every unknown token.
+   */
+  useEffect(() => {
+    manageDomLifecycle('#tokenImportModal');
+  }, []);
+
+  useEffect(() => {
+    let cancelled = false;
+
+    async function fetchDetails() {
```
tuliomir: suggestion(non-blocking): Declare the `fetchDetails` function outside the `useEffect()`. This both makes it: 

- Easier to read since all data it needs will be offered by parameters
- More performant, as it will only be declared once, instead of re-declared on every effect run

src/components/ModalTokenImport.js
```diff
@@ -0,0 +1,437 @@
...
+  const dispatchRegistrations = (uids) => {
+    setModalState(MODAL_STATE.REGISTERING);
+    let completed = 0;
+    const errors = {};
+
+    uids.forEach((uid) => {
+      dispatch(
+        tokenRegisterRequested(uid, {
+          // Keep imported tokens visible even with zero balance,
+          // so they don't vanish when "Hide zero-balance tokens" is enabled.
```
tuliomir: suggestion(non-blocking): Make the "Unknown tokens tracker" also take into consideration if `hide zero balance tokens` is active.

If it is active, do not offer to register zero-balance tokens.

Imagine the situation where the user has 10 unregistered tokens with zero balance AND "Hide Zero Balance Tokens" active. This workflow here would add all 10 tokens as "always show", and essentially break the user experience by having the user navigate to each token and disable "Always Show" individually.

src/screens/Wallet.js
```diff
@@ -82,6 +83,7 @@ function Wallet() {
   // Refs
   const alertSuccessRef = useRef(null);
   const unregisterModalRef = useRef(null);
+  const tokenInfoInFlight = useRef(null);
```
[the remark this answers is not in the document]
raul-oliveira: Done

src/components/ModalTokenImport.js
raul-oliveira: Nice catch. I wasn't aware of this feature. 
Done in https://github.com/HathorNetwork/hathor-wallet/pull/859/commits/9434826beccd518ced64595f44a4170fcf6f5fb0

src/components/ModalTokenImport.js
raul-oliveira: Done in https://github.com/HathorNetwork/hathor-wallet/pull/859/commits/f6e5ab3f92d602c2c567efa048d7c3ec5b5b6475

src/components/ModalTokenImport.js
raul-oliveira: Done

src/components/ModalTokenImport.js
raul-oliveira: Both of the implementations, your suggestion and the set approach are O(1), the set also is better when getting the size of the list in the toggle all function with the `.size()` which is O(1) and is O(n) in the object approach. 

I got your point of using only one state variable and use the other auxiliary as computed and I like it. So it's done in https://github.com/HathorNetwork/hathor-wallet/pull/859/commits/effa37537f77879d1bdccb288f0fc12f3d518aa2

src/screens/Wallet.js
```diff
@@ -97,11 +105,17 @@ function Wallet() {
     initializeWalletScreen();
   }, [selectedToken]);
 
-  // When the tokens history changes, update the token info
+  // Debounce: the reducer creates a new reference for the selected
+  // token's history on every registration, so rapid successive
+  // changes are coalesced into a single API call.
+  const selectedTokenHistory = tokensHistory[selectedToken];
```
raul-oliveira, line 111: Bringing some context to this change:

When we call register token in the import tokens modal, the tokenHistory was being updated and calling `wallet.getTokenDetails` again to fetch the updated balance. Since we don't it to be updated in all token registration process, I've changed this to listen only to the selected token.

The debounce is here for cases that the selected token is affected by the other tokens that is being registered, like HTR or DBT used to pay fees.

src/screens/Wallet.js
```diff
@@ -156,19 +170,32 @@ function Wallet() {
     if (tokenUid === hathorLib.constants.NATIVE_TOKEN_UID) {
       return;
     }
-    const tokenDetails = await wallet.getTokenDetails(tokenUid);
 
-    // If the user has changed the selectedToken while we were fetching the data, discard it
-    if (selectedToken !== tokenUid) {
+    // Skip if a request for this token is already in flight
+    if (tokenInfoInFlight.current.has(tokenUid)) {
```
raul-oliveira, line 175: This piece was a suggestion of code rabbit, in a real life case, the user may not be clicking between two tabs in the token bar. In any case, it's a tiny improvement that will prevent this kind of duplicated requests.

src/components/ModalTokenImport.js
```diff
@@ -0,0 +1,442 @@
...
+    };
+  }, [unknownTokens]);
+
+  async function fetchDetails(tokens, isCancelled) {
+    const wallet = getGlobalWallet();
+    const details = {};
+    let errorCount = 0;
+
+    const makeEntry = (uid, name, symbol, balance) => ({
+      uid, name, symbol, balance, selected: true,
```
pedroferreira1: Why default selected?

raul-oliveira: Nice catch. The default behavior is unselected. It's fixed. tks!

pedroferreira1 (dismissed):

src/utils/wallet.js
```diff
@@ -249,10 +249,11 @@ const wallet = {
 
       // If the "show only non-zero balance tokens" flag is active, filter here.
       if (hideZeroBalance) {
-        const totalBalance = balance.available + balance.locked;
+        const balanceData = balance.data || balance;
```
pedroferreira1: Why `balance.data`? Was it a bug before?

raul-oliveira: Yes. It was. The unregistered tokens was showing even with the flag "hide zero balance tokens" enabled. We should always access via `.data` since the saga wraps the balance inside it.

pedroferreira1 (dismissed):

pedroferreira1 (dismissed):

src/components/ModalTokenImport.js
```diff
@@ -0,0 +1,442 @@
...
+          token.uid, tokenInfo?.name || token.uid, tokenInfo?.symbol || '???', token.balance
+        );
+
+        // Throttle between API calls
+        await delay(200);
+      } catch (_err) {
+        // Fallback: use uid as name and ??? as symbol
+        details[token.uid] = makeEntry(token.uid, token.uid, '???', token.balance);
+        errorCount += 1;
+        console.error(_err);
```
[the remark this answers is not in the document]
tuliomir: @raul-oliveira was this solved?

src/utils/wallet.js
```diff
@@ -252,7 +252,7 @@ const wallet = {
         const totalBalance = balance.available + balance.locked;
 
         // This token has zero balance: skip it.
-        if (hideZeroBalance && totalBalance === 0n) {
+        if (totalBalance === 0n) {
```
tuliomir, line 255: issue: Why was this removed?

If the user has activated the "hide zero balance tokens" feature, won't this be necessary?

raul-oliveira:
https://github.com/HathorNetwork/hathor-wallet/pull/859#discussion_r3074597688 the following comment was addressed by removing the redudant call to fullnode since we already fetched all the tokens in the wallet startup. This decision was made after some conversions between us and Pedro.

src/utils/wallet.js
raul-oliveira: It's not removed, the totalBalance === 0n is inside the hizeZeroBalance condition

src/components/ModalTokenImport.js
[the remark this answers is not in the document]
raul-oliveira: Yes. The code was removed. https://github.com/HathorNetwork/hathor-wallet/pull/859#issuecomment-4252734646

src/screens/Settings.js
```diff
@@ -107,6 +109,22 @@ function Settings() {
...
+   */
+  const importTokens = () => {
+    const hideZeroBalance = wallet.areZeroBalanceTokensHidden();
+    const unknownTokens = wallet.fetchUnknownTokens(allTokens, registeredTokens, tokensBalance, hideZeroBalance);
+
+    let hasHiddenZeroBalanceTokens = false;
+    if (hideZeroBalance) {
+      const allUnknown = wallet.fetchUnknownTokens(allTokens, registeredTokens, tokensBalance, false);
+      hasHiddenZeroBalanceTokens = allUnknown.length > unknownTokens.length;
+    }
```
pedroferreira1: Why don't you move this code to the token import modal component? Then you don't need to duplicate it here and in the `TokenImportBanner` component.

raul-oliveira: Nice catch. It's moved. Tks

pedroferreira1 (approved):

src/components/ModalTokenImport.js
```diff
@@ -53,13 +55,28 @@ function truncateUid(uid) {
...
+  const allTokens = useSelector((state) => state.allTokens);
+  const registeredTokens = useSelector((state) => state.tokens);
+  const tokensBalance = useSelector((state) => state.tokensBalance);
   const explorerUrl = useSelector((state) => state.networkSettings.data.explorer);
   const decimalPlaces = useSelector((state) => state.serverInfo.decimalPlaces);
 
+  const hideZeroBalance = walletUtils.areZeroBalanceTokensHidden();
+  const unknownTokens = useMemo(
+    () => walletUtils.fetchUnknownTokens(allTokens, registeredTokens, tokensBalance, hideZeroBalance),
+    [allTokens, registeredTokens, tokensBalance, hideZeroBalance]
```
tuliomir, line 71: thought(non-blocking): There is a improvement possibility here, since we don't really need `tokensBalance` to identify an unknown token. Not an issue for the scope of this PR, however.

src/components/ModalTokenImport.js
```diff
@@ -53,13 +55,28 @@ function truncateUid(uid) {
...
+  );
+
+  const hasHiddenZeroBalanceTokens = useMemo(() => {
+    if (!hideZeroBalance) return false;
+    const allUnknown = walletUtils.fetchUnknownTokens(allTokens, registeredTokens, tokensBalance, false);
```
tuliomir, line 76: chore(non-blocking): Another improvement possibility. Fetch all unknown tokens and then process them for zero balance or not, avoiding the two calls made here between those two memoized functions.

tuliomir (approved):

src/components/ModalTokenImport.js
raul-oliveira: It will be addressed in https://github.com/HathorNetwork/hathor-wallet/issues/862

src/components/ModalTokenImport.js
raul-oliveira: It will be addressed in https://github.com/HathorNetwork/hathor-wallet/issues/862

PR #861, opened 2026-04-01
feat: single address mode

raul-oliveira:
## Summary

This pull request implements a **single address mode feature** for the Hathor Wallet, allowing users to choose between single and multi-address wallet modes through a settings interface.

## Acceptance Criteria

- Adds single address feature flag
- Shows change address mode in settings page
- Settings page facelift
- Address mode modal with two options: single and multi address modes
- Disables single address option when wallet already has transactions

## Address Mode Determination on Wallet Start

The decision logic is centralized in the pure helper `decideAddressMode({ singleAddressFeatureEnabled, storedAddressMode })`:

1. **Flag off** → `MULTI` (feature unavailable on this network)
2. **Flag on, stored preference = `MULTI`** → `MULTI` (user explicitly chose multi via Settings; respect)
3. **Flag on, no stored preference or stored `SINGLE`** → attempt `SINGLE`. The wallet-lib downgrades to `GAP_LIMIT` automatically if it detects transactions on addresses with index > 0 during the first connection. The saga syncs Redux + localStorage with the real mode after `wallet.start`.

The same rules apply on **network change** and **Reset All Data** — both flows re-run `startWallet`, which re-evaluates the flag against the target network.

The address mode is persisted **per network** (`wallet:address_mode:<network>`), so a preference in one network does not leak into another where the rollout flag has a different value. Custom networks are covered too. Persistence to `localStorage` happens at a single point, after the connection is established, with the value the wallet-lib actually accepted.

`Reset All Data` clears every address mode preference for every network via `walletUtils.clearAllAddressModes()`, called before `LOCAL_STORE.resetStorage()` (which only removes static keys).

## Pre-existing bug fixed in this PR

While testing the network-change flow, I found a **feature flag context leak** that already existed in master and was exposed by `single-address-desktop.rollout`: `state.featureToggles` could still reflect the previous network after switching networks or Reset All Data. The reducers were preserving stale toggles, and `updateUnleashClientContext` was not synchronizing Redux right after refreshing the Unleash client.

The leak affected **every flag** read via `checkForFeatureFlag` — not only the new single-address one. Fixed in this PR:

- `updateUnleashClientContext` now dispatches `setFeatureToggles` synchronously once `unleashClient.updateContext(...)` resolves.
- `onWalletReset` calls `updateUnleashClientContext` after `loadStorageState()` so the Unleash client and Redux reflect the post-reset network before the next `startWallet`.
- Companion reducer change (`dbcd26e`) preserves `state.featureToggles` across `START_WALLET_RESET` and `CLEAN_DATA`.

### Security Checklist

#### Settings screen
image

#### Existing wallet using multi address 
image

#### New wallet
image

#### Wallet screen (single address mode)
image

#### Wallet screen (multi address mode - as is)
image

####  Reown without caller section
image

## Summary by CodeRabbit

* **New Features**
  * Address mode control in Settings with a modal to choose Single vs Multi address; UI adapts to the chosen mode (hides generate/see-all links when Single).
* **Style**
  * Settings screen facelift with improved layout, spacing, and address-mode styling.
* **Documentation**
  * Updated translatable strings and source references for Settings and address-mode UI.
* **Tests**
  * Increased e2e timeout and adjusted assertions for wallet error-flow reliability.
* **Chores**
  * Ignored yalc artifacts and bumped wallet-lib dependency.


src/components/ModalAddressMode.js
```diff
@@ -0,0 +1,146 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React, { useState, useEffect } from 'react';
+import $ from 'jquery';
+import { t } from 'ttag';
+import { getGlobalWallet } from '../modules/wallet';
+import helpers from '../utils/helpers';
+
+const LEARN_MORE_URL = 'https://www.google.com';
```
raul-oliveira: I'm waiting for the latest version of hathor docs with the link

src/components/ModalAddressMode.js
```diff
@@ -0,0 +1,146 @@
...
+        const result = await wallet.hasTxOutsideFirstAddress();
+        setHasTxOutside(result);
+      } catch (e) {
+        console.error('Error checking hasTxOutsideFirstAddress:', e);
+      } finally {
+        setLoading(false);
+      }
+    };
+
+    checkTxOutside();
```
raul-oliveira, line 26: declare and use this function outside the useEffectHook

cypress/e2e/00-welcome.cy.js
```diff
@@ -243,17 +243,18 @@ describe('create a new wallet and back it up', () => {
       req.reply({
         statusCode: 500,
         body: {
-          history: [],
+          success: false,
```
raul-oliveira, line 246: The success field is expected by zod in wallet-lib lastest version

cypress/e2e/00-welcome.cy.js
```diff
@@ -243,17 +243,18 @@ describe('create a new wallet and back it up', () => {
...
+          message: 'Internal Server Error',
         },
       });
     });
 
     cy.findByText('Next').click();
 
-    // PIN was successful
-    cy.contains('Loading transactions'); // For a few seconds this screen will be shown
```
raul-oliveira, line 254: I'm talking to @tuliomir to see why this isn't passing in the ci. Locally it's passing either in headless or interactive modes

src/components/ModalAddressMode.js
```diff
@@ -0,0 +1,147 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React, { useState, useEffect } from 'react';
+import $ from 'jquery';
+import { t } from 'ttag';
+import { getGlobalWallet } from '../modules/wallet';
+import helpers from '../utils/helpers';
+
+const LEARN_MORE_URL = 'https://docs.hathor.network/explanations/features/wallet-service/address-mode';
```
tuliomir, line 15: issue: This URL does not exist yet 👀

src/components/ModalAddressMode.js
```diff
@@ -0,0 +1,147 @@
...
+  const handleSave = () => {
+    onSave(selectedMode);
+  };
+
+  const openLearnMore = (e) => {
+    e.preventDefault();
+    helpers.openExternalURL(LEARN_MORE_URL);
+  };
+
+  const isSingleDisabled = hasTxOutside && currentMode !== 'single';
```
tuliomir: question: If there are no txs on many addresses but it's on "multi address" mode, should the single mode be disabled?

src/components/ModalAddressMode.js
```diff
@@ -0,0 +1,147 @@
...
+            <div>
+              <div
+                className={`address-mode-option ${selectedMode === 'single' ? 'address-mode-option--selected' : ''} ${isSingleDisabled ? 'address-mode-option--disabled' : ''}`}
+                onClick={() => !isSingleDisabled && setSelectedMode('single')}
+              >
+                <div className="d-flex align-items-center mb-2" style={{ gap: '6px' }}>
+                  <input
+                    type="radio"
+                    name="addressMode"
+                    checked={selectedMode === 'single'}
```
tuliomir: chore: Create a constant for those new modes, avoiding hardcoded strings and making it easier to refactor in the future.

This is especially important in the `sagas` and `utils`, and should be refactored on all files.

```suggestion
                    checked={selectedMode === WALLET_MODE_SINGLE_ADDRESS}
```

src/sagas/wallet.js
```diff
@@ -139,6 +145,24 @@ export function* startWallet(action) {
...
+  const singleAddressFeatureEnabled = yield call(checkForFeatureFlag, SINGLE_ADDRESS_FEATURE_TOGGLE);
+  let addressMode = walletUtils.getAddressMode();
+
+  if (!isExistingWallet) {
+    // New wallet (import or create): always determine fresh
+    addressMode = singleAddressFeatureEnabled ? 'single' : 'multi';
+    walletUtils.setAddressMode(addressMode);
+  } else if (addressMode === null) {
+    // Existing wallet with no stored preference: default to multi
+    addressMode = 'multi';
```
tuliomir: suggestion(non-blocking): Add these rules to the PR description for quick reference by future maintainers

tuliomir (dismissed):

src/utils/wallet.js
```diff
@@ -576,6 +578,37 @@ const wallet = {
...
+   * Gets the address mode preference
+   *
+   * @return {string|null} 'single', 'multi', or null if not set
+   *
+   * @memberof Wallet
+   * @inner
+   */
+  getAddressMode() {
+    const mode = LOCAL_STORE.getItem(storageKeys.addressMode);
+    if (mode === 'single' || mode === 'multi') {
```
pedroferreira1: I think we could have these strings in the constants file as an ENUM or a simple object. What do you think?

If we do this, we should update all places with this string

src/sagas/wallet.js
```diff
@@ -318,6 +346,21 @@ export function* startWallet(action) {
...
+  if (singleAddressFeatureEnabled && addressMode === 'single') {
+    const hasTxOutside = yield call([wallet, wallet.hasTxOutsideFirstAddress]);
+
+    if (hasTxOutside) {
+      walletUtils.setAddressMode('multi');
+      // Re-dispatch start so the wallet loads with the correct scanning policy
+      yield put(action);
+      return;
+    }
+  }
```
pedroferreira1: Why do you do this here? Isn't it better to check this in the beginning of the method, so you execute the whole start with the correct address mode, instead of re-dispatching it?

raul-oliveira: Nice catch it's simplified now. Done in https://github.com/HathorNetwork/hathor-wallet/pull/861/commits/5024a1b1a064890c5e1fb8bc68356b18b49d9765

src/utils/wallet.js
raul-oliveira: Done in https://github.com/HathorNetwork/hathor-wallet/pull/861/commits/d7e568c61ed8b2d58d80bf4bb9c73c01afd4fe55

src/components/ModalAddressMode.js
raul-oliveira: Done in https://github.com/HathorNetwork/hathor-wallet/pull/861/commits/d7e568c61ed8b2d58d80bf4bb9c73c01afd4fe55

src/components/ModalAddressMode.js
raul-oliveira: it there's not tx on many address the "single address" option will be enabled

src/components/ModalAddressMode.js
raul-oliveira: Jorge deployed it while we are reviewing this pr. :)

src/sagas/wallet.js
raul-oliveira: done

raul-oliveira: Done

src/screens/Settings.js
```diff
@@ -278,72 +281,124 @@ function Settings() {
...
+        </p>
+        {singleAddressEnabled && (
+          <p>
+            <strong>{t`Address Mode:`}</strong>{' '}
+            {addressMode === 'single' ? t`Single address` : t`Multi address`}
+            <a className="settings-change-link" href="true" onClick={(e) => { e.preventDefault(); openAddressModeModal(); }}>
+              {t`Change`}
+            </a>
+          </p>
+        )}
```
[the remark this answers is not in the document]
raul-oliveira: it's fixed in the last version of the code.

src/components/ModalAddressMode.js
```diff
@@ -0,0 +1,148 @@
...
+import { getGlobalWallet } from '../modules/wallet';
+import helpers from '../utils/helpers';
+import { ADDRESS_MODE } from '../constants';
+
+const LEARN_MORE_URL = 'https://docs.hathor.network/explanations/features/wallet-service/address-mode';
+
+function ModalAddressMode({ currentMode, onSave, onClose }) {
+  const [selectedMode, setSelectedMode] = useState(currentMode);
+  const [hasTxOutside, setHasTxOutside] = useState(false);
+  const [loading, setLoading] = useState(true);
```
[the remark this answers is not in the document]
raul-oliveira: done in https://github.com/HathorNetwork/hathor-wallet/pull/861/commits/8583017b946322c2c7040564417a93a389c7933f

src/utils/wallet.js
```diff
@@ -576,6 +581,64 @@ const wallet = {
...
+   * The per-network keys are dynamic (one entry per network the user has connected to,
+   * including custom networks), so they cannot live in LOCAL_STORE's static
+   * storageKeys list. Wallet reset must call this to avoid leaking a stored
+   * preference into a freshly-created wallet.
+   *
+   * @memberof Wallet
+   * @inner
+   */
+  clearAllAddressModes() {
+    const prefix = 'wallet:address_mode:';
```
tuliomir: suggestion(non-blocking): Move this to a constant

tuliomir (dismissed):

src/screens/Settings.js
```diff
@@ -285,73 +289,126 @@ function Settings() {
...
+      <div className="settings-info">
+        <CopyToClipboard text={uniqueIdentifier} onCopy={copyClicked}>
+          <span>
+            <p><strong>{t`Unique identifier:`}</strong> {uniqueIdentifier} <i className="fa fa-clone pointer ml-1" title={t`Copy to clipboard`}></i></p>
+          </span>
+        </CopyToClipboard>
+        <p onDoubleClick={() => setShowTimestamp(!showTimestamp)}><strong>{t`Date and time:`}</strong> {showTimestamp ? hathorLib.dateFormatter.dateToTimestamp(now) : now.toString()}</p>
       </div>
+
+      <button type="button" className="settings-reset-button" onClick={resetClicked}>{t`Reset wallet`}</button>
```
pedroferreira1: We had a conditional before in this button.

` { !isHardwareWallet && <button className="btn btn-hathor mt-4" onClick={resetClicked}>{t`Reset all data`}</button> }`

Does it make sense to keep it?

src/utils/wallet.js
```diff
@@ -576,6 +581,64 @@ const wallet = {
...
+    const keysToRemove = [];
+    for (let i = 0; i < localStorage.length; i++) {
```
pedroferreira1, line 631: In the wallet we've been using `LOCAL_STORE` instead of `localStorage`. Should we change?

src/reducers/index.js
```diff
@@ -596,6 +600,8 @@ const onCleanData = (state) => {
     loadingAddresses: state.loadingAddresses,
     ledgerWasClosed: state.ledgerWasClosed,
     featureTogglesInitialized: state.featureTogglesInitialized,
+    // Must be preserved together with featureTogglesInitialized — see onStartWalletReset.
+    featureToggles: state.featureToggles,
```
pedroferreira1: This is from another PR, we should remove it here (the change below as well)

src/sagas/featureToggle.js
```diff
@@ -110,7 +110,15 @@ export function* updateUnleashClientContext(networkSettings = null) {
...
-  yield call(() => unleashClient.updateContext(options));
+  // updateContext refetches toggles for the new context. Once the promise
+  // resolves, unleashClient.toggles already reflects the new network.
+  // Sync redux synchronously here so consumers reading state.featureToggles
+  // immediately after (e.g. startWallet -> checkForFeatureFlag) see the
+  // post-context values, not the previous network's stale values.
+  yield call([unleashClient, unleashClient.updateContext], options);
+
+  const featureToggles = mapFeatureToggles(unleashClient.toggles);
+  yield put(setFeatureToggles(featureToggles));
```
pedroferreira1, line 121: Same comment as in the reducer

src/sagas/wallet.js
```diff
@@ -267,6 +316,8 @@ export function* startWallet(action) {
       nanoContractsEnabled,
       genesisHash,
     }));
+
+    yield put(setAddressMode(attemptedAddressMode));
```
pedroferreira1: After the wallet starts, we might be in a different address mode. The wallet can redirect to multi address automatically, for example, if the wallet has more txs outside first address.

Should we set the address mode from the wallet lib, instead of the attempted one? Or they will always be the same?

pedroferreira1: I just saw you do this check below. Can't you remove this from here and just leave the code below?

src/sagas/wallet.js
```diff
@@ -318,6 +369,21 @@ export function* startWallet(action) {
...
+  let finalAddressMode = attemptedAddressMode;
+  if (attemptedAddressMode === ADDRESS_MODE.SINGLE) {
+    const actualScanPolicy = yield call([wallet.storage, wallet.storage.getScanningPolicy]);
+    if (actualScanPolicy !== SCANNING_POLICY.SINGLE_ADDRESS) {
+      finalAddressMode = ADDRESS_MODE.MULTI;
+      yield put(setAddressMode(ADDRESS_MODE.MULTI));
+    }
+  }
+
+  walletUtils.setAddressMode(network, finalAddressMode);
```
pedroferreira1:
```suggestion
  const actualScanPolicy = yield call([wallet.storage, wallet.storage.getScanningPolicy]);
  const finalAddressMode = actualScanPolicy === SCANNING_POLICY.SINGLE_ADDRESS ? ADDRESS_MODE.SINGLE : ADDRESS_MODE.MULTI;
  yield put(setAddressMode(finalAddressMode));
  walletUtils.setAddressMode(network, finalAddressMode);
```

It seems cleaner and you can remove the `setAddressMode` above. What do you think?

src/sagas/wallet.js
```diff
@@ -777,12 +843,33 @@ export function* onWalletReset() {
   const wallet = getGlobalWallet();
 
   localStorage.removeItem(IGNORE_WS_TOGGLE_FLAG);
+  // Clear per-network address mode preferences. LOCAL_STORE.resetStorage()
+  // only removes keys in its static storageKeys list, but addressMode keys
+  // are namespaced per network (wallet:address_mode:<network>) and so must
+  // be cleared explicitly here.
```
pedroferreira1, line 844: Hummmm does it make sense to change the `resetStorage` to remove also namespaced keys? It was good that you noticed it. In the future we might have new namespaced keys and this could become a bug

src/sagas/wallet.js
```diff
@@ -777,12 +843,33 @@ export function* onWalletReset() {
...
+  // from the user's perspective); on failure we explicitly write
+  // FEATURE_TOGGLE_DEFAULTS so checkForFeatureFlag (which reads via lodash.get
+  // with a per-key default) doesn't return the previous network's stale values
+  // — the defaults only kick in for missing keys, not pre-existing stale ones.
+  try {
+    yield call(updateUnleashClientContext);
+  } catch (e) {
+    console.error('Failed to refresh unleash context after wallet reset', e);
+    yield put(setFeatureToggles(FEATURE_TOGGLE_DEFAULTS));
+  }
```
pedroferreira1: The same as in the reducer comment. This is from another PR

src/utils/wallet.js
raul-oliveira: `LOCAL_STORE` doesn't expose key iteration (`length` / `key(i)`), so the loop itself has to hit `localStorage` directly — same pattern used by the migration code in `src/storage.js` (`for (const key of Object.keys(localStorage))`). Switched the `removeItem` call to `LOCAL_STORE.removeItem(key)` for consistency, since that part has a wrapper available.

src/screens/Settings.js
raul-oliveira: Done

src/reducers/index.js
raul-oliveira: It'll be removed when we merge the #865

src/sagas/wallet.js
raul-oliveira: It'll be removed when we merge the https://github.com/HathorNetwork/hathor-wallet/pull/865

src/reducers/index.js
```diff
@@ -1151,6 +1157,13 @@ export const onStartWalletFailed = (state) => ({
  * - isVersionAllowed: API version check is independent of wallet data
  * - ledgerWasClosed: Ledger device state persists across wallet instances
  * - featureTogglesInitialized: Unleash client runs independently
+ * - featureToggles: must be preserved together with the initialized flag,
```
raul-oliveira: Also this comments will be udpated with #865 version

src/sagas/wallet.js
raul-oliveira: Done

src/sagas/featureToggle.js
raul-oliveira: Done

src/reducers/index.js
raul-oliveira: Done

src/utils/wallet.js
raul-oliveira: Done

src/sagas/wallet.js
raul-oliveira: Good catch — you're right on both points.

The block further down already computes the effective mode from `wallet.storage.getScanningPolicy()`, which is the real source of truth. The dispatch here was redundant and also created a brief window where Redux said "SINGLE" while the lib was about to flip it to MULTI.

Removed this line and made the post-ready block always dispatch `finalAddressMode` (not only on the override path), so Redux and localStorage are written from the same source.

src/sagas/wallet.js
raul-oliveira: Done

src/sagas/wallet.js
raul-oliveira: Hmm, good point — but isn't that risk kind of already present in the current `resetStorage`?

Today it iterates over a manually maintained list (`storageKeys`), so if someone adds a new top-level key tomorrow and forgets to register it there, we'd hit the exact same orphan-key bug at reset.

Moving the namespaced case into `resetStorage` — say, by introducing a parallel `storageKeyPrefixes` list — wouldn't really eliminate that risk; it would mirror the same "register-or-leak" pattern, just for prefixed keys. The developer still has to remember to register the new prefix somewhere.

So I think the trade-off is more about *where* the registration lives: next to the call site that creates the namespaced keys (current approach in the saga), or consolidated in `storage.js` alongside `storageKeys`. Both are equally vulnerable to a missed registration.

That said, I do see the value in centralizing it for discoverability. If you think it's worth it, happy to move the prefix handling into `storage.js` as a follow-up so everything reset-related lives in one place. WDYT?

pedroferreira1 (dismissed):

pedroferreira1 (dismissed):

src/sagas/wallet.js
```diff
@@ -318,6 +367,18 @@ export function* startWallet(action) {
...
+  // After connection, confirm the actual scan policy. The wallet-lib changes
+  // SINGLE_ADDRESS → GAP_LIMIT during the first connection if it finds tx on
+  // addresses with index > 0. We persist the *real* mode here — this is the
+  // single point of write for addressMode (both Redux and localStorage).
+  const actualScanPolicy = yield call([wallet.storage, wallet.storage.getScanningPolicy]);
+  const finalAddressMode = actualScanPolicy === SCANNING_POLICY.SINGLE_ADDRESS
+    ? ADDRESS_MODE.SINGLE
+    : ADDRESS_MODE.MULTI;
+
+  yield put(setAddressMode(finalAddressMode));
```
tuliomir, line 379: suggestion(non-blocking): Knowing that we only have "Multi" and "Single", it would make sense to simplify by removing `finalAddressMode`, wouldn't it?
```suggestion
  yield put(setAddressMode(actualScanPolicy));
```

tuliomir (dismissed):

src/sagas/wallet.js
raul-oliveira: Thanks Tulio! Unfortunately we can't simplify it that way — the two enums use different string values:

- `SCANNING_POLICY.SINGLE_ADDRESS` = `'single-address'` / `GAP_LIMIT` = `'gap-limit'` / `INDEX_LIMIT` = `'index-limit'` (defined in `wallet-lib` `types.ts`)
- `ADDRESS_MODE.SINGLE` = `'single'` / `MULTI` = `'multi'` (defined in `src/constants.js`)

Dispatching `actualScanPolicy` directly would persist `'single-address'`/`'gap-limit'` into Redux + localStorage, which would break every downstream consumer that compares against `ADDRESS_MODE.SINGLE`/`MULTI` (e.g. `utils/wallet.js`, `screens/Settings.js`, the rest of this saga). It'd also misclassify `INDEX_LIMIT`, which today is intentionally folded into `MULTI`.

The mapping is deliberate: `ADDRESS_MODE` is the app's UI/persistence enum, `SCANNING_POLICY` is wallet-lib's internal enum — keeping them decoupled means a future scan policy added in wallet-lib doesn't silently leak into our stored state. Happy to extract a tiny `scanPolicyToAddressMode()` helper if it'd read better, but I'd keep the translation step.

package.json
```diff
@@ -38,7 +38,7 @@
   },
   "dependencies": {
     "@hathor/hathor-rpc-handler": "4.4.0",
-    "@hathor/wallet-lib": "2.17.0",
+    "@hathor/wallet-lib": "3.0.1",
```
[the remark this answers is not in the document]
raul-oliveira: ## Audit: wallet-lib v2.17.0 → v3.0.1 against current code

I diffed wallet-lib between v2.17.0 and v3.0.1, listed the breaking surface, and grepped this repo for each affected API. **The codebase is not affected by any v3 breaking change.**

### What changed in wallet-lib v3.x (breaking surface)

| # | Change | Source |
|---|---|---|
| 1 | `wallet.signTx(tx)` now requires PIN; uses `transactionUtils.signTransaction` + `prepareToSend` instead of the old per-input `getSignatures` + `setData` loop | `src/new/wallet.ts` |
| 2 | `wallet.changeServer(...)` was sync `void`, now `async Promise<void>` | `src/new/wallet.ts` |
| 3 | Type renames: `Balance` → `WalletServiceBalance`, `Authority` → `WalletServiceAuthority`, `DelegateAuthorityOptions` / `DestroyAuthorityOptions` → `WalletService...` variants | `src/wallet/types.ts` |
| 4 | `CreateNanoTxData` → `FullnodeCreateNanoTxData` | `src/new/wallet.ts` |
| 5 | New required APIs for single-address: `enableSingleAddressMode`, `enableMultiAddressMode`, `getAddressMode`, `WalletAddressMode`, `SCANNING_POLICY`, `HasTxOutsideFirstAddressError`, `singleAddressMode` start option | `src/new/wallet.ts`, `src/wallet/wallet.ts` |
| 6 | `createNanoContractTransaction` options now include `signTx?: boolean` (defaults `true`); when `false`, returns an unsigned `SendTransaction` | `src/new/wallet.ts` |
| 7 | Public exports broadened via `export *` | `src/lib.ts` |

`partial_tx.ts`, `partialTxProposal.ts`, and `swapService.ts` had **zero changes** between v2.17.0 and v3.0.1 — atomic-swap code is untouched.

### Audit of call sites in this repo

**`wallet.signTx()` callers** — ✅ Safe. No direct callers in `src/`. The local `getSignatures` in `SendTokens.js:291` calls `ledger.getSignatures`, not `wallet.signTx`.

**`wallet.changeServer()` callers** — ✅ Safe. All four callers use `yield call([wallet, wallet.changeServer], ...)` (`sagas/networkSettings.js:43,65,93,165`). redux-saga's `call` awaits returned Promises automatically, so sync→async is transparent.

**`Balance` / `Authority` / `DelegateAuthorityOptions` / `DestroyAuthorityOptions` type imports** — ✅ Safe. None imported by name.

**`CreateNanoTxData` import** — ✅ Safe. Not imported.

**`createNanoContractTransaction` at `NanoContractExecuteMethod.js:285`** — ✅ Safe. The call passes `{ pinCode: pin }`. v3 merges `{ pinCode: null, signTx: true, ...options }` → `signTx: true`, identical to pre-v3 behavior.

**Single-address APIs** (`enableSingleAddressMode`, `SCANNING_POLICY`, `singleAddressMode`) — ✅ Already adopted. Used in `Settings.js:306-308`, `sagas/wallet.js:10,186-238`.

**`HathorWallet`, `HathorWalletServiceWallet`, `transactionUtils`, `walletUtils`, `tokensUtils`, `helpersUtils`, `numberUtils`, `bigIntUtils`, `dateUtils`, `scriptsUtils`, `bufferUtils`, `cryptoUtils`, `config`, `errors`** — ✅ Safe. All still exported by v3 (the public surface widened, didn't shrink).

**Disabled UI buttons (`createNC` / `executeMethod`)** — Cosmetic TODO. The underlying handler at `NanoContractExecuteMethod.js:285` is itself API-compatible with v3. The buttons were defensively commented out, but re-enabling is a behavior decision, not a v3 fix.

### Validation

`npm test` results: 8 suite failures, all the same root cause — Jest can't transform axios's ESM import (`Cannot use import statement outside a module`). That's a `transformIgnorePatterns` problem in the Jest config, **not** a wallet-lib v3 incompatibility.

### Conclusion

Current code is not affected by wallet-lib v3.0.1 breaking changes. The broad "may be incompatible with `HathorWallet`, `walletUtils`, `PartialTx`, `swapService`, `config`, `transactionUtils`" claim is unfounded — those modules either were not changed (`PartialTx`, `swapService`) or only had additive/renamed-type changes that this repo doesn't touch.

The only real follow-up item is the policy decision on whether to re-enable the `createNC` and `executeMethod` UI buttons — and that's orthogonal to the v3 bump.

tuliomir (approved):

pedroferreira1 (approved):

PR #873, opened 2026-06-23
feat: consent to dApp balance access at connection time

raul-oliveira:
### Acceptance Criteria
- Disclose at connection time that the dApp will be able to view the balance of the user's registered tokens
- Remove the per-request Get Balance confirmation modal (`GetBalanceModal`), its modal/action types, and the `onGetBalanceRequest` saga handler
- Bump `@hathor/hathor-rpc-handler` to 5.0.0

### Why
`hathor-rpc-handler` 5.0.0 drops the `GetBalanceConfirmationPrompt` trigger — balance reads are no longer confirmed per request. Consent now happens once, during connect.

`@hathor/hathor-rpc-handler` 5.0.0 is published on npm (`latest`), so the dependency pin resolves cleanly.

### Security Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed the balance-approval modal from the wallet connection flow.
  * Updated the connection prompt to clearly show that a connected dApp can view token balances.
  * Refreshed translation references for several settings, wallet, and Reown-related screens.


package.json
```diff
@@ -37,7 +37,7 @@
     "@hathor/hathor-rpc-handler/@hathor/wallet-lib/bitcore-lib": "8.25.10"
   },
   "dependencies": {
-    "@hathor/hathor-rpc-handler": "4.4.0",
+    "@hathor/hathor-rpc-handler": "5.0.0",
```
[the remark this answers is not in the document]
raul-oliveira: \`@hathor/hathor-rpc-handler@5.0.0\` is published on npm — \`npm view @hathor/hathor-rpc-handler dist-tags\` returns \`latest: 5.0.0\` (published 2026-07-01T01:22Z) and it is listed in \`npm view @hathor/hathor-rpc-handler versions\`. The earlier web query was stale because it hit the third-party npmx.dev mirror rather than the npm registry. The \`5.0.0\` pin in package.json resolves cleanly (package-lock.json is regenerated against it), so installs will not fail. Keeping the pin.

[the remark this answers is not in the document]
raul-oliveira: You're wrong. It's published.

lavamoat/webpack/policy.json
```diff
@@ -110,37 +110,6 @@
         "@hathor/wallet-lib>zod": true
       }
     },
-    "@hathor/hathor-rpc-handler>@hathor/wallet-lib": {
```
pedroferreira1, line 113: Is this correct? You removed it but you actually still have it in package-lock. Just not sure if we should regenerate this file

lavamoat/webpack/policy.json
```diff
@@ -953,7 +922,7 @@
       "packages": {
         "node-stdlib-browser>crypto-browserify>public-encrypt>browserify-rsa>bn.js": true,
         "buffer": true,
-        "node-stdlib-browser>crypto-browserify>randombytes": true
+        "ttag-cli>serialize-javascript>randombytes": true
```
tuliomir, line 956: issue: I think it's suspicious to have so many changes on the lavamoat policy file because of such a small change.

This deserves a double-check.

raul-oliveira: Fixed. It was something wrong with my local environement.
The tag-cli was previously added as a dev dependency but I didn't regenerate the policy file. That's why even after the fix in the file, tag-cli still referenced as a change.

lavamoat/webpack/policy.json
raul-oliveira: Fixed. It was something wrong with my local environement.
The tag-cli was previously added as a dev dependency but I didn't regenerate the policy file. That's why even after the fix in the file, tag-cli still referenced as a change.

pedroferreira1 (approved):

tuliomir (approved):

PR #874, opened 2026-06-24
fix: remove insecure Electron flags (nodeIntegration, contextIsolation)

raul-oliveira:
### Acceptance Criteria
- Remove the insecure Electron flags from the renderer `BrowserWindow`: `nodeIntegration: true` → `false` and `contextIsolation: false` → `true` (the warnings Electron itself raises). `sandbox` is kept `false` so the preload can still use Node to set up the bridge — enabling full sandboxing is a follow-up.
- Replace the renderer's `window.require('electron')` access with a `contextBridge` API (`window.electronAPI`) exposed from `preload.js`, with a **whitelisted** IPC surface (`send` / `on` / `removeAllListeners`) plus `openExternal`.
- Keep the Ledger hardware-wallet integration working **unchanged**: the USB/HID transport stays entirely in the main process (`public/ledger.js`) and is reached over the IPC bridge; binary payloads are restored to `Buffer` when crossing the bridge so `public/ledger.js` does not change.
- Move the renderer's Sentry off `window.require('@sentry/electron')` to the preload (reached via the bridge, outside the LavaMoat-governed bundle); `@sentry/browser` is kept only as the browser/dev fallback.
- Fix the Send Tokens screen crash (`TypeError: ft.setTimeout is not a function`): `lodash` reads `setTimeout` off the captured global object and relied on Node's `global` (which bypassed LavaMoat) — disabling `nodeIntegration` exposed the missing grant. Fixed with a `lavamoat/webpack/policy-override.json` grant.

### Security Checklist
  - No new dependencies added. `@sentry/browser` was already a production dependency (previously the browser/dev fallback).

---

### Testing
Validated on a physical **Ledger Nano S** (Hathor app):
- ✅ Connect + open app — `getVersion`, `getPublicKeyData` (xpub import)
- ✅ Show/verify address — `checkAddress`
- ✅ **Send transaction** — `sendTx` + `getSignatures` (the binary path through the bridge); transaction broadcast successfully
- ✅ Send Tokens screen renders correctly (after the lodash policy fix)

The Electron `nodeIntegration` / `contextIsolation` security warnings are gone from the console; only the unrelated CSP warning remains.

## Summary by CodeRabbit

* **Security**
  * Hardened the desktop app’s renderer environment with safer Electron settings and a controlled bridge for app access.
  * Restricted external link opening to trusted web URLs only.

* **New Features**
  * Added support for improved error reporting in the desktop app.
  * Enabled the app to communicate with the main process through a more limited, secure API.

* **Bug Fixes**
  * Improved compatibility for data passed between app processes.
  * Updated desktop external-link handling to work reliably through the new bridge.


pedroferreira1 (commented):
Review by claudin:

```
P1 — send() throws for every non-Buffer payload; Ledger send hangs forever

public/preload.js:81 — ipcRenderer.send(channel, ...args.map(restoreBuffers)). Array.prototype.map passes (element, index, array), so the array index lands in restoreBuffers's seen parameter, replacing the WeakSet default. Any top-level argument that's a plain object or array hits seen.has(value) → TypeError: seen.has is not a function. I ran the exact function from the diff: ledger:getSignatures (number array, src/utils/ledger.js:184), ledger:sendTokens and ledger:verifyManyTokenSignatures (token-map objects) all throw; only Buffer/primitive/no-arg channels survive.

User-visible impact, traced through the screens:
- HTR send: sendTx (a single Buffer) crosses fine, the Ledger prompts, the user approves on device — then handleTxSent calls getSignatures (src/screens/SendTokens.js:170), which is async and called without await/.catch (SendTokens.js:293), so the TypeError becomes a swallowed rejection and the app hangs forever on the footer-less, data-backdrop="static" "Validate outputs on Ledger" modal. Restart required; transaction never broadcast.
- Custom-token send: ledger.sendTokens throws synchronously before the flow starts → generic "Unhandled error" modal.

This contradicts the PR's "getSignatures ✅ transaction broadcast successfully" testing claim — no code path in commit ec185a0 can complete that flow, so the device testing must predate the seen parameter. Ask for a re-test on hardware after the fix.

The right fix is to delete restoreBuffers entirely, not repair it. The function can't achieve its stated purpose: ipcRenderer.send re-serializes with the Structured Clone Algorithm (Electron 27 docs), so Buffers created in the preload arrive in the main process as Uint8Array anyway — exactly what main received before this PR. Verifiers confirmed public/ledger.js is already Uint8Array-safe (data.length/data.slice at ledger.js:339-340; @ledgerhq/hw-transport's Buffer.concat accepts Uint8Array). ipcRenderer.send(channel, ...args) is the whole fix. While in there, make the SendTokens.js:293 wrapper .catch → handleSendError so a future rejection surfaces instead of hanging that modal.

P1 — Sentry consent is broken both ways in Electron

Two sides of the same migration, both verified against the pinned @sentry/electron 3.0.7 sources:

1. Opt-out is a silent no-op. sentrySetEnabled: (dsn) => Sentry.init({ dsn, ... }) (preload.js:109) relies on re-init, but the renderer SDK has a one-shot guard (window.__SENTRY__RENDERER_INIT__ in renderer/sdk.js) — after the unconditional startup init with the real DSN (preload.js:13), every later Sentry.init returns early. So updateSentryState → initSentry('') never disables anything; users who decline error reporting keep an active SDK whose events forward via IPC to the main process (also unconditionally init'd, electron.js:17). The "Empty dsn disables it" comment codifies a contract the SDK doesn't honor. (Partially pre-existing at base, but this PR makes sentrySetEnabled the only Electron path, so this is the place to fix it.)
2. Opt-in users lose all reporting. With contextIsolation: true, the preload SDK's global handlers live in the isolated world and never see main-world uncaught errors (at base, the shared window meant auto-capture worked). The browser SDK is never initialized in Electron (wallet.js:461 returns early via the bridge), and ModalUnhandledError.js:87 shows the manual "Send error report" button only when !isSentryAllowed() — so consented users have no path at all for app errors to reach Sentry. This is a real regression introduced by the contextIsolation flip.

Suggested shape: make the bridge API a boolean (sentrySetEnabled(enabled)) with the DSN kept inside the preload (also removes the renderer-supplied-DSN surface); gate captures by flipping the live client (getClient().getOptions().enabled = enabled — checked before the EventToMain processor runs; beforeSend would be too late), ideally starting disabled until consent is read; and capture main-world errors explicitly (e.g. ErrorWrapper's handlers → sentryCapture when consented). Note the related pre-existing bug that becomes load-bearing: ErrorWrapper.js:46 does setError(error) instead of setError(newError) and never passes the error to the modal, so the one surviving report path sends new Error(undefined) with no stack — worth fixing in passing.

P2 â LavaMoat grant keyed to cypress>lodash is a production time bomb

policy-override.json:324 keys the Send-Tokens crash fix to "cypress>lodash". lodash is a phantom dependency â imported by ~36 src files but declared nowhere in package.json; @lavamoat/aa canonicalizes it through cypress (a devDependency) only because it prefers the shortest direct-dep name. The policy regenerates on every production build, so if cypress is removed/upgraded or hoisting shifts, the key silently stops matching (no build error, only a verbosity-gated debug log), the setTimeout grant vanishes â the generated policy can only detect define for lodash â and the ft.setTimeout is not a function crash returns in packaged builds only, invisible to dev/e2e. Fix: add lodash@4.17.21 to dependencies, re-key the override as "lodash", regenerate policy.json.

Minor

- Release tag: both Sentry.init calls in preload use process.env.npm_package_version, which is undefined in packaged builds. Use constants.versionNumber (already required; matches electron.js). Note for the CodeRabbit thread: its suggested constants.VERSION does not exist in public/constants.js, and its "init with dsn: '' then enable later" patch would break enabling due to the same one-shot init guard â so don't apply it as written.
```

